PR #667, opened 2015-09-03
Only issue warning when provisioning.

drj11:
There is a warning about `GALAXY_DATABASE_DIR` not being set, but it is commented out. I think it's useful. The naive way of including it means that you get the warning for innocuous commands like `vagrant --help`, but adding the warning as a provision script means that you only get it when you provision.

It looks like this for a `vagrant up`

screenshot from 2015-09-03 13 48 48

But as you see, `vagrant --help` is not affected:

screenshot from 2015-09-03 13 53 19


drj11:
thanks. I was intending to chat to you about it, but now you've merged, it's good.

PR #798, opened 2015-12-04
Deploy on Amazon AWS

drj11:
documentation in `deployment/AWS.md`

At the moment, this uses separate copies of the Puppet manifest (`aws.pp` instead of `default.pp`), `apache.conf` (templated via `apache.conf.erb` instead of `apache.conf`) and `config.json` (edited on startup via a hacky `sed` script).

Next step (after this PR is merged) will be fix those issues.

### DIFF
--- a/deployment/AWS.md
+++ b/deployment/AWS.md
@@ -1,5 +1,7 @@
 # Deployment on Amazon AWS
 
+## Summary
+
 Generate AWS Cloud Formation template file using `cfn-pyplates`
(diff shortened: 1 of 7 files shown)


hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;OejRcN0q&#x2F;15-implement-single-instance-aws-deployment"> implement single instance AWS deployment</a>

PR #828, opened 2015-12-15
All the puppet code in one place

drj11:
The only way I could find to share puppet code was to make a module. That module is in `deployment/refinery-modules/refinery` (it gets copied to `modules` via `Puppetfile` and `librarian-puppet` in the usual way). There is just one big module right now, but in the future, perhaps more.

So now, `manifests/aws.pp` and `manifests/default.pp` have greatly shrunk.

`config.json` and `apache.conf` are now both templated using `.erb` template syntax (only a tiny bit in each case).

The future of `config.json` is uncertain: that's fine.

I've only done minimal testing of this on Vagrant.

Probably best to chat about this first, before merging.


hackdna:
Can we simply move `deployment/refinery-modules/refinery/manifests/init.pp` into `deployment/manifests`?

drj11:
I'll try it (I don't think so, but I don't think I've tried it).

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;OejRcN0q&#x2F;15-implement-single-instance-aws-deployment"> implement single instance AWS deployment</a>

PR #846, opened 2016-01-12
Document how to delete a stack

drj11:
@ngehlenborg pointed out that we need to know how to delete a stack as well!

As far as I'm concerned, this is good to merge.


ngehlenborg:
I will merge this once the current release has been finalized.

PR #870, opened 2016-02-08
Move AWS config into a directory

drj11:


drj11:
Probably best to see this section of the AWS doc: https://github.com/parklab/refinery-platform/wiki/AWS-installation#generate-json-file-and-start-stack

Issue #839, opened 2016-01-06
Enable tagging of all AWS resources during deployment


hackdna:
Tags should be added automatically to all resources in the stack.
http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html
http://cloud.huit.harvard.edu/financial-management-amazon-web-services-accounts#appendix

Proposed scheme:
owner=drj11
department=dbmi
environment=dev
project=scc
product=refinery
subproduct=galaxy|platform

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;sBJxolm4&#x2F;195-implement-deployment-to-aws-using-ebs-and-rds"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Implement deployment to AWS using EBS and RDS</a>

drj11:
This is now working on a branch (soon to be a PR!). It looks like this:

![tags](https://cloud.githubusercontent.com/assets/1215412/12918494/6145ce42-cf37-11e5-92ea-fc9269de949a.png)

hackdna:
Great. We probably don't need the 'subproduct' tag.

drj11:
can argue about exactly which tags we do/don't need in another issue. Easy to change, just add stuff in `aws-tags` directory. The set of tags is likely to change according to AWS account holder anyway.

PR #874, opened 2016-02-09
tag AWS resources

drj11:
The code is okay, but it needs a bit of doc. The short version is that the tags are in YAML files in the `aws-tags` directory.

### DIFF
--- a/deployment/Makefile
+++ b/deployment/Makefile
@@ -7,7 +7,7 @@ web-stack: web.json .PHONY
 	aws cloudformation create-stack --stack-name refinery-web-$$(date +%Y%m%dT%H%M) --template-body file://web.json
 
 volume.json: .PHONY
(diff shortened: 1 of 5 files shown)


drj11:
implements #839

drj11:
something to note here: the EC2 instances start with a small (8GB) EBS as their root device. This device isn't yet tagged.

drj11:
tagging the root device may be a bit tricky. http://stackoverflow.com/questions/24026425/is-there-a-way-to-tag-a-root-volume-when-initializing-from-the-cloudformation-te applies. The solution there is for the EC2 instance to discover its own root volume ID and tag it using the AWS API.

drj11:
I've opened another issue for tagging the root device: https://github.com/parklab/refinery-platform/issues/875

PR #881, opened 2016-02-11
Set the Availability Zone for the new EC2 instance

drj11:
Fixing a bug where `make web-stack` would fail when it created an instance in a different Availability Zone to the existing Volume that it was trying to attach.

Took a surprisingly long time to find because the preferred Availability Zone seems to be a bit "sticky".

### DIFF
--- a/deployment/requirements.txt
+++ b/deployment/requirements.txt
@@ -5,3 +5,4 @@ paramiko==1.14.0
 pycrypto==2.6.1
 cfn-pyplates==0.5.0
 awscli==1.8.13
+boto3==1.2.3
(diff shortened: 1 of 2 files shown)


hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;hilnqIeP&#x2F;216-merge-david-s-pull-request"> merge David&#39;s pull request</a>

drj11:
This now works (no change, I just had to wait), but fails to mount the `/data/` directory; which is a separate issue.

Merge at will.

PR #899, opened 2016-02-23
Integrate with AWS SES

drj11:
Enables email on Amazon.

This now works (phew!) at least for me when I start a fresh stack and I register a new refinery account.


drj11:
Note: The URL in the email that you get is wrong:

registration

drj11:
Fix #836

PR #913, opened 2016-02-26
Create fresh role for EC2 instance

drj11:
Creating the role is now pushed into the Cloud Formation template, which eliminates the need to create it "by hand" first.

- [x] update the documentation in the wiki


drj11:
Fixes https://github.com/parklab/refinery-platform/issues/912

PR #921, opened 2016-03-02
automatically create IAM user and get SMTP credentials

drj11:
This means the admin no longer needs to create SMTP credentials using the Web UI.


drj11:
Merge at will

PR #924, opened 2016-03-02
Scottx611x/file extension rework

### LINKED ISSUE #305: explore advantages (and disadvantages) of using a model to manage file types
Right now files with an extension that is not registered as a (hard-corded) file type will have an UNKNOWN file type, which is ok in principle but since file types depend on the tools used in workflows it is probably not the best solution to hard-code them. A model would allow workflow engine managers and admins to create file types for their workflows. Since file types eventually will also be used to check whether a particular file can be fed into a given workflow or viewer, knowing the file type might become crucial and file type UNKNOWN would be a serious problem.

Pivotal Tracker story 47211551 (Nils Gehlenborg - Apr 1, 2013)

scottx611x:
This PR separates the `extension` field that existed in the `FileType` model into its own model (`FileExtension`). 

Doing so addresses a larger potential issue that when users will be adding the file types, they might not know one exists already and/or they might misspell the name and then would have have two distinct entries for the same `FileType`

Fixes #305


hackdna:
Still in progress

Issue #927, opened 2016-03-04
URLs for downloading analysis results are incorrect on AWS


hackdna:
Commit 0b896f6.

## Steps to reproduce
Run an analysis

## Observed results
URLs for analysis results are of the form: `http://192.168.50.50:8000/media/file_store/38/6e/1_MACS2_plot.pdf`

## Expected results
Analysis result URLs should contain the actual IP address or host name instead of `192.168.50.50:8000`

drj11:
While this isn't the same bug as #900, I very much hope that it has the same underlying cause as #900

ngehlenborg:
Make download URLs relative.

drj11:
So I think the discussion in the meeting was that the front-end was constructing full absolute URLs and it should not be, as @ngehlenborg says.

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;0dqHQzfq&#x2F;34-prepare-for-scc-demo-on-aws"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> prepare for SCC demo on AWS</a>

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;0dqHQzfq&#x2F;34-prepare-scc-demo-on-aws"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> prepare SCC demo on AWS</a>

PR #993, opened 2016-04-07
Return only the relative url, but still use the full url in vizualiza…

scottx611x:
…tion_manager/views.py

Fixes: #927

>Note: This needs to be tested on AWS before merging!

**Functionality:**
apr 07 2016 09 14

### DIFF
--- a/refinery/file_store/models.py
+++ b/refinery/file_store/models.py
@@ -26,7 +26,6 @@ from django.dispatch import receiver
 from django.db import models
 from django.db.models.signals import pre_delete
 from django_extensions.db.fields import UUIDField
-from django.contrib.sites.models import Site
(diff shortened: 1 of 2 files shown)


scottx611x:
@hackdna or @drj11  Could you please test these changes on an AWS instance, and merge if the issue has been addressed?

hackdna:
@ngehlenborg: could you provide AWS account access to Scott?

scottx611x:
This is not working properly on AWS. I've tried editing the source of some `FileStoreItems` to point to an actual file on disk and still get pointed to `192.168.50.50:8000<file_path>` 

It seems that the front end is constructing this url based on the value of the Domain Name field of the Site Model. 

Can we simply change that value to the EC2 instance's IP?

PR #1002, opened 2016-04-12
Scottx611x/django 1.7 upgrade

### LINKED ISSUE #614: Upgrade to Django 1.7
https://docs.djangoproject.com/en/1.7/releases/
- https://docs.djangoproject.com/en/1.7/releases/1.7/
  - initial_data fixtures are no longer loaded for apps with migrations; if you want to load initial data for an app, we suggest you create a migration for your application and define a RunPython or RunSQL operation in the operations section of the migration (https://docs.djangoproject.com/en/1.7/releases/1.7/#initial-data)
- [Upgrading from South](https://docs.djangoproject.com/en/1.7/topics/migrations/#upgrading-from-south):
  - [x] Remove 'south' from INSTALLED_APPS
  - [x] Delete all your (numbered) migration files, but not the directory or **init**.py - make sure you remove the .pyc files too
  - [x] `python manage.py makemigrations`
  - [x] `python manage.py migrate`

scottx611x:
Fixes #614


drj11:
Because the `syncdb` puppet rule has been renamed in this branch, it will break this code when merging: https://github.com/parklab/refinery-platform/pull/1004/files#diff-05de70d9570a93320c3895d841911716R58

The `require` will need changing, shouldn't be too much trouble.

Issue #980, opened 2016-03-22
Solr index is not stored in EBS storage


ngehlenborg:
Solr index is currently stored in ephemeral storage, which means we can't bring down the stack and re-launch with the same data.

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;P37o9MWh&#x2F;226-data-set-contents-not-displayed-when-re-launching-refinery-on-aws"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Data set contents not displayed when re-launching Refinery on AWS</a>

ngehlenborg:
It looks like the index directory is configured in the `solrconfig.xml` for each core (note that we have two, `core` and `data_set_manager`), e.g.:

https://github.com/parklab/refinery-platform/blob/ab73cac5ee25882040e57e0a09b3e52249f62a65/refinery/solr/data_set_manager/conf/solrconfig.xml#L114

drj11:
So, as per https://wiki.apache.org/solr/SolrConfigXml#System_property_substitution I think we can create a `<solr.home>/conf/solrcore.properties` file and add `solr.data.dir` to it.

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;0dqHQzfq&#x2F;34-prepare-for-scc-demo-on-aws"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> prepare for SCC demo on AWS</a>

hackdna:
Sounds good though I think it would be [solrconfig.xml](https://cwiki.apache.org/confluence/display/solr/Configuring+solrconfig.xml) for each of the cores/collections for our version of Solr. Also, I don't think we need to use system property substitutions.

drj11:
closing now #1027 is merged.

PR #1027, opened 2016-04-19
Move solr data to persistent volume

drj11:
This PR moves the solr data to the `/data` volume.

I have tested this by adding a dataset and then deleting the stack and recreating it. The dataset still appears to be intact, but I'm not sure I would be able to tell if it was wrong.

dataset


drj11:
Fixes #980

PR #1037, opened 2016-04-21
Tag the root EBS

drj11:
MERGE AT WILL!


drj11:
- [ ] rebase onto `origin/develop`

drj11:
As part of this change, the EC2 instances get a new power to set EC2 tags (an additional policy is attached to the role that cloudformation creates).

PR #1080, opened 2016-05-04
Single CloudFormation stack

drj11:
- [x] update AWS doc


drj11:
I don't understand codecov's problem, but apart from that, it's Good To Merge.

PR #1142, opened 2016-05-17
Add configuration for REFINERY_DATA_IMPORT_DIR

### LINKED ISSUE #1139: REFINERY_DATA_IMPORT_DIR is not set correctly on AWS
Current value in `config.json.erb` is hardcoded to `/vagrant/import`. It should something like `/data/import` (similar to `MEDIA_ROOT`).

drj11:
- [x] Testing on an AWS stack now...

Good from my end.

Fixes #1139

Merge at will.


drj11:
On a fresh instance:

```
$ grep IMPORT refinery/config/config.json
  "REFINERY_DATA_IMPORT_DIR": "/data/import",
```

PR #1156, opened 2016-05-20
AWS Django settings

drj11:
Now requires `DJANGO_SETTINGS_MODULE` to be set in `aws-config/config.yaml`


drj11:
Tested, seems to work. I also tested I could break it by specifying a Django module that wasn't there.

PR #1189, opened 2016-06-07
Get RDS endpoint directly from CloudFormation

drj11:
The AWS deployment now gets the RDS credentials directly from the CloudFormation template.

Meaning that the script that runs on the EC2 instance to fetch the RDS details has been removed.


drj11:
And I now realise that this probably mean we can remove from the EC2 instance role the need to access the RDS api.

PR #1194, opened 2016-06-10
Scottx611x/gzipped files fix

### LINKED ISSUE #1083: Downloading input file after its been analyzed yields weird results.
Commit: 43089c9de423f230b9b874b0e7c2fa2c5f8d6217 or newer

### Steps to reproduce
1. Run an analysis using the `Test workflow: 5 steps without branching` workflow and the`Request for Comments (RFC) Test` dataset

### Observed behavior
1. Before an Analysis has been run on one of the dataset's files, the arrow download button next to it can be clicked and the text file can be viewed in the browser. 

2. After an Analysis has been run on a given file,  the arrow download button next to it can be clicked, but an actual download occurs. When the downloaded text file is viewed it is comprised of a bunch of odd characters.

See: 
screen shot 2016-05-04 at 1 12 33 pm

### Expected behavior
1. To see the actual text file upon clicking the download arrow like we see before an analysis is run on the file

scottx611x:
This PR fixes #1083

### DIFF
--- a/refinery/analysis_manager/tasks.py
+++ b/refinery/analysis_manager/tasks.py
@@ -397,7 +397,7 @@ def get_galaxy_download_tasks(analysis):
                 # only download files if size is greater than 1
                 if file_size > 0:
                     task_id = import_file.subtask(
-                            (filestore_uuid,))
+                            (filestore_uuid, False, file_size))
                     task_list.append(task_id)
 
     return task_list
--- a/refinery/analysis_manager/tasks.py
+++ b/refinery/analysis_manager/tasks.py
@@ -79,7 +79,7 @@ def run_analysis(analysis_uuid):
         refinery_import_tasks = []
         for input_file_uuid in analysis.get_input_file_uuid_list():
             refinery_import_task = import_file.subtask(
-                    (input_file_uuid, False, ))
+                    (input_file_uuid,))
             refinery_import_tasks.append(refinery_import_task)
         refinery_import = TaskSet(tasks=refinery_import_tasks).apply_async()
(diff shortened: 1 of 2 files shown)


scottx611x:
> **NOTE:** This PR shouldn't be merged until the new functionality is tested upon remote files encoded with the `deflate` format

scottx611x:
I've tested this with a webserver configured to encode content using each of 'gzip', zlib', and 'deflate' formats.

Everything seems to be working properly as I can import and run an analysis with each of the respective encoding formats. 

When I emulate a misconfigured webserver, i.e. (`content-encoding` header is set, but content isn't encoded) bad things happen.

So far, with the broken configuration, I see this behavior:
 -  Trying to download an imported file: screen shot 2016-06-13 at 2 03 53 pm

- Trying to run an analysis:
`ERROR/MainProcess] Task file_store.tasks.import_file[0a7dcc7e-cc09-4b7c-a781-6142c264f1c4] raised unexpected: AttributeError("'ConnectionError' object has no attribute 'reason'",)`

hackdna:
Thanks. It's likely that zlib raises an exception when trying to decode the content that's not compressed (there should be a traceback in the log). We need to handle this condition accordingly.

scottx611x:
@hackdna Good call, `zlib` throws: `Error -3 while decompressing data: incorrect header check` when this issue occurs. I'll handle accordingly and push.

hackdna:
Great but we probably don't need to iterate through the rest of the file when this occurs and we should fail the import task which should result in an error displayed in the browser to inform the user of the problem.

PR #1249, opened 2016-07-15
Add Private IP to ALLOWED_HOSTS

drj11:
on AWS.

Notes

- Only works on AWS. That's okay, because it's in a settings file that is only used on AWS;
- Only needed when `Debug=False`, so only needed in `aws.py` not `awsdev.py`

### DIFF
--- a/refinery/config/settings/aws.py
+++ b/refinery/config/settings/aws.py
@@ -6,6 +6,8 @@
 # Like prod, but overriding some things.
 from .prod import *  # NOQA
 
+import requests
+
 # Ensure that the AWS private IP is an Allowed Host.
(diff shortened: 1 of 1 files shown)


drj11:
Have tested:

- Normal access is not broken;
- accessed via curl using Private IP. For example: `curl http://172.31.42.141/` (compare with `curl http://127.0.0.2/` which does not work).

Have not tested with real ELB health check.

hackdna:
Thanks. We should add this to `awsdev.py` too. Also, we should probably move `import requests` above `from .prod import *`.

drj11:
We don't need it in `awsdev.py` because `Debug=True`

PR #1256, opened 2016-07-19
Jkmarx/data set name bug

### LINKED ISSUE #1242: Data set name is not displayed correctly in the launch pad
Commit: ec43d5ba6d2f6cd5efcd583c9f420c11a4d5f548

### Steps to reproduce
1. Import http://data.cloud.refinery-platform.org.s3.amazonaws.com/data/sample-data/E-GEOD-23694/e-geod-23694.txt
1. Navigate to the launch pad

### Observed behavior
Data set is listed as "Untitled data set" in the launch pad (data set name is displayed correctly in the data set browser).

### Expected behavior
Data set name should be displayed correctly in the launch pad.

### LINKED ISSUE #1255: Change layout of the metadata upload page
Move "Select file" button below the list of column separators.

jkmarx:
Resolves #1255 and resolves #1242 
Also added conditional to model method to avoid leading/trailing white spaces in slug.

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -215,11 +215,11 @@ class BaseResource(models.Model):
         # Check if model being saved/altered in Django Admin has a slug
         # duplicated elsewhere.
 
-        # Strip whitespace from slugs
+        # Removes trailing/leading white spaces
         if self.slug:
             self.slug = self.slug.strip()
 
-        if not self.slug == '':
+        if self.slug:
             if self.duplicate_slug_check() == 0:
                 pass
             else:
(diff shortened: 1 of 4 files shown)


refinery/core/models.py
```diff
@@ -214,6 +214,11 @@ def duplicate_slug_check(self):
     def clean(self):
         # Check if model being saved/altered in Django Admin has a slug
         # duplicated elsewhere.
+
+        # Removes trailing/leading white spaces
```
drj11: i would start the method with

    if not self.slug:
        return

then avoid the `if self.slug` later

jkmarx: @drj11 My comment here was not clear. I will update it. The other reason for this method is to catch blank spaces passed as a slug (tested through the admin interface). This reason is why they are separated.

scottx611x: The duplicated logic here is necessary for the special case where a slug is set to just whitespace or None (which we want to allow duplicates of) There may be a better way to do this, but we need the assignment in the first `if self.slug`  so that the second if block isn't reached if a slug is `None` or `""`

drj11: subtle.

PR #1259, opened 2016-07-20
On metadata file upload, remove leading/trailing white spaces on all …

### LINKED ISSUE #1252: Problem loading tab delimited metadata file generated as output from R
Commit:

### Steps to reproduce
1. Tried to load the attached file. The first column heading is missing (this is the default behaviour for writing an object out to file in R). 
[Mov10_full_meta.txt](https://github.com/parklab/refinery-platform/files/370073/Mov10_full_meta.txt)

### Observed behavior

Doesn't display table but looks like it is loading it - headings are shown in the "Source Index Column" input box.

screen shot 2016-07-18 at 4 30 19 pm

### Expected behavior
Should load the preview correctly.

jkmarx:
Resolves #1252 

@hackdna Trimming each row may be overkill. If so, the code can be modified for just trimming the header.

### DIFF
--- a/refinery/ui/source/js/data-set-import/controllers/metadata-table.js
+++ b/refinery/ui/source/js/data-set-import/controllers/metadata-table.js
@@ -104,7 +104,14 @@ MetadataTableImportCtrl.prototype.renderTable = function () {
   var reader = new FileReader();
   reader.onload = function (event) {
     self.$rootScope.$apply(function () {
-      self.metadata = self.parser(event.target.result);
+      // trim whitespaces before and after the entire str then split on newline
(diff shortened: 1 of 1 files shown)


hackdna:
Are we sending the original uploaded metadata file or its parsed contents to the server? If the former then the backend code needs to be updated as well.

jkmarx:
@hackdna Separate issue. https://github.com/parklab/refinery-platform/issues/1260

hackdna:
OK. Is it better to duplicate this functionality in the backend or just send the parsed contents to the server instead?

PR #1263, opened 2016-07-22
Jkmarx/node group api reformat

### LINKED ISSUE #1244: Node Group API - Current Selection First
The default current selection node group should be the first item in array.
The api should sort the current selection before sending response.

Ref #1038

jkmarx:
Resolves #1244

### DIFF
--- a/refinery/core/utils.py
+++ b/refinery/core/utils.py
@@ -829,7 +829,7 @@ def admin_ui_deletion(request, objects_to_delete, single_model=None):
             create_delete_response_message(delete_response)
 
 
-def move_obj_to_front(obj_arr, prop_key, prop_value):
+def move_obj_to_front(obj_arr, match_key, match_value):
     """
         Helper method move the first obj matching to the front of the arr
         based on key and value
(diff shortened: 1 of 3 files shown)


refinery/core/utils.py
```diff
@@ -827,3 +827,24 @@ def create_delete_response_message(del_response):
...
+def move_obj_to_front(obj_arr, prop_key, prop_value):
+    """
+        Helper method move the first obj matching to the front of the arr
+        based on key and value
+        :param objects_to_delete: iterable of Objects selected in admin UI,
+        or a single object instance if `delete_model:admin_ui_deletion` is
+        called with `single_model` having a truthy value
+        :param obj_arr: An array of objects
+        :param prop_key: Key to match
+        :param prop_key: Value to match
```
scottx611x: checkout typo `:param prop_key:` x 2

refinery/core/utils.py
```diff
@@ -827,3 +827,24 @@ def create_delete_response_message(del_response):
             # Fix for multiple messages displaying
             messages.set_level(request, messages.ERROR)
             create_delete_response_message(delete_response)
+
+
+def move_obj_to_front(obj_arr, prop_key, prop_value):
```
scottx611x: Can we rename `prop_key` and `prop_value` to something like `key_to_match`  & `value_to_match`??

jkmarx: @scottx611x How about match_key and match_value?

scottx611x: @jkmarx sounds good to me 👍

PR #1266, opened 2016-07-26
Override django config

drj11:
Single `aws-config/config.yaml` file. Includes automatically generated stuff.

Any setting made here will be visible to Django (via this commit included in this PR: https://github.com/parklab/refinery-platform/pull/1266/commits/77c3b7f9cdb9e529aa5be0bcd944ebf3d3fa7ace).
This means you can add or change any `config.json` setting without writing new deployment code.

The entire `config.yaml` file is copied to the target EC2 instance as both YAML, `/home/ubuntu/config.yaml`, and JSON, `/home/ubuntu/config.json`.
This is how configuration that happens on the target should find its configured values (the historical stuff doesn't do this, but it's simpler, so we should).

Tested and ready to merge.


drj11:
so this works, but I'm waiting for comments on the approach.

drj11:
So right now this involves create an extra file, called `override-django-config.yaml`.
It gets copied to the config directory of the refinery Django app on the target instance, and any setting in that file overrides whatever is says in `config.json`.

In terms of puppet it is simpler to do this if all you want to do is set a load of Django settings.
But it is an extra file for the Site Manager to worry about.

Issue #1170, opened 2016-05-25
Integrate ELB with the CloudFormation stack


hackdna:
The current beta site is deployed behind an ELB which terminates HTTPS. We should be able to create a new ELB when setting up a brand new stack or re-use an existing one if available.
There should be settings added to the config.yml.template for load balancer name, all necessary HTTPS certificate files and other items necessary for provisioning an ELB.
Apache config template should be changed to configure redirect from HTTP to HTTPS.
Set ServerName in Apache config from SITE_URL

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;yK35fIKV&#x2F;260-add-elb-to-the-stack"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Add ELB to the stack</a>

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;WKWVJxrD&#x2F;261-configure-https-using-puppet"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Configure HTTPS using Puppet</a>

hackdna:
An example of Apache config file updated for HTTPS:
```
# Generated from apache.erb.conf
# by refinery-modules/refinery/manifests/init.pp

WSGIPythonHome /home/ubuntu/.virtualenvs/refinery-platform

SetEnv DJANGO_SETTINGS_MODULE config.settings.aws

<VirtualHost *:80>
    ServerName dev.stemcellcommons.org

    RewriteEngine On
    RewriteCond %{HTTP:X-Forwarded-Proto} !https
    RewriteRule ^.*$ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]

    <Directory /srv/refinery-platform/refinery/config>
        <Files wsgi.py>
            Order deny,allow
            Require all granted
        </Files>
    </Directory>
    WSGIScriptAlias / /srv/refinery-platform/refinery/config/wsgi.py

    WSGIDaemonProcess refinery user=ubuntu group=ubuntu \
        python-path=/srv/refinery-platform/refinery:/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages
    WSGIProcessGroup refinery

    #Alias /robots.txt /vagrant/refinery/static/robots.txt
[...]

PR #1319, opened 2016-08-09
Elastic Load Balancer

drj11:
GOOD TO MERGE

This works, but you only after manually adding a DNS entry that is aliased to the ELB DNS name.

As an aid to debugging the instance, you can use the instance's public IP address directly and avoid using the ELB.

First part of #1170 

- [x] document recommended dev and prod setups in wiki
- [x] Split #1170

### DIFF
--- a/deployment/apache.conf.erb
+++ b/deployment/apache.conf.erb
@@ -6,7 +6,7 @@ WSGIPythonHome <%= @virtualenv %>
 SetEnv DJANGO_SETTINGS_MODULE <%= @django_settings_module %>
 
 <VirtualHost *:80>
-    ServerName refinery.example.org
+    ServerName <%= @site_url %>
 
     RewriteEngine On
     RewriteCond %{HTTP:X-Forwarded-Proto} !https
--- a/deployment/apache.conf.erb
+++ b/deployment/apache.conf.erb
@@ -8,9 +8,11 @@ SetEnv DJANGO_SETTINGS_MODULE <%= @django_settings_module %>
 <VirtualHost *:80>
     ServerName <%= @site_url %>
 
+    <% if @tls_rewrite %>
     RewriteEngine On
     RewriteCond %{HTTP:X-Forwarded-Proto} !https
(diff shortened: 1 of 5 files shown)


drj11:
Adding HTTP -> HTTPS redirect.

I've investigated the following (by editing the configuration by hand):

- Configure the ELB to have a listener on port 80 that forwards to port 8302 on the instance.
- Configure apache on the instance to listen on port 8302 and always redirect to `https://site.example.com/`.
- requires another VirtualHost directive; and,
- another Listen directive (in `ports.conf`); and,
- an additional rule in the instance's security group (this is currently the default security group).

This works.
And I like the fact I can still, for debugging, contact the site directly on `instance-ip:80`.

It might be a bit complicated though. In particular to do this now would involve avoiding the default security group and configuring a new on in CloudFormation. Which we should do, but we should not add it to this PR.

drj11:
Oh. Now my current apache config is issuing a redirect to the ELB Health Check, which the ELB doesn't like, so the ELB is marking the instance as "OutOfService".

drj11:
So we can either:

1. detect the health check and avoid the redirect; or,
2. wire the health check to a different port and add a VirtualHost (and a security group rule); or,
3. reinstate the default VirtualHost and configure the refinery one to use the correct hostname; or,
4. something else I haven't thought of.

We should configure the refinery VirtualHost to use the correct hostname anyway, which leans me towards option 3.
This should work because then the ELB healthcheck will hit the default VirtualHost,
and therefore get a valid response.

drj11:
so this looks good now, but, is miserable for developers (a bit).

The apache config now has two VirtualHost entries, `000-default.conf`, and `001-refinery.conf`,
which means that the refinery VirtualHost only gets used if the correct hostname appears in the web request.
Meaning you have to have DNS set up.

I plan to remove the `000-default.conf` VirtualHost (again), so that all web requests go to refinery.
That creates the problem that the ELB health check needs to avoid the TLS redirect.
Investigating how to make that work.

PR #1345, opened 2016-08-16
Jkmarx/data import reorganize tabs

### LINKED ISSUE #1336: Data Import - Reorganize Tab and Content
- [x] Move optional raw data upload beneath options but before ISA-tab upload
- [x] Move the optional raw data upload into the tabular metadata form.
- ~~[ ] Upload raw data form first -> the tabular metadata form -> redirect to file browser~~
- [x] Move ISA-tab upload to a new tab

- [x] Rename tabs, Tabular Metadata and ISA-Tab Metadata
- [x] Change download text to tabular metadata file
- [x] Update required pre-select to [1] datafile and [2] for species
- [x] Update section names to Provide Metadata file, Configure Metadata Import, and Optional: Upload Data Files
- [x] Make clear button more pronounced
- [x] Remove (e.g .csv, .tsv, .txt)
- [x] Radio button lines horizontally
- [x] Change color of buttons in the data upload section to default (gray).
- [x] Deactivate the blue "Start Import" button until any data file has been uploaded (we should check for running ones only - those that have been added but for which the user never started the upload should maybe be ignored since there is currently no way to remove a file once it has been added).
- [x] Renaming the "Start Import" button into "Upload Data Set"
- [x] Add alert when people start uploading datasets when files are still queued.

jkmarx:
Resolves #1336


refinery/ui/source/js/data-set-import/state.js
```diff
@@ -10,16 +10,14 @@ function DataSetImportStates (
...
         templateUrl: '/static/partials/data-set-import/views/import.html',
-        controller: 'RefineryImportCtrl as import'
       },
       '/data_set_manager/import/'
     )
     .state(
-      'fileUpload', {
-        url: '/upload/',
+      'isaTabImport', {
+        url: '/isa-tab-import',
```
scottx611x: @jmarx We believe there should be a trailing slash here ^^^

refinery/ui/source/js/data-set-import/state.spec.js
```diff
@@ -56,10 +56,10 @@ describe('DataSetImport.state:', function () {
     });
 
     it(
-      'should be "fileUpload" when path is "/upload"',
+      'should be "isaTabImport" when path is "/isaTabImport"',
       function () {
-        goTo('/upload');
-        expect($state.current.name).toEqual('fileUpload');
+        goTo('/isa-tab-import');
```
scottx611x: @jkmarx same trailing slash here ^^^

PR #1379, opened 2016-08-30
Java and JS IGV

### LINKED ISSUE #1356: For the time being, support both old and new IGV
We want to continue to support the old IGV for a while longer. There will just be two buttons where now there is only one. Suggested wording, I think:
- Stand-alone IGV
- Web-based IGV

mccalluc:
Fix #1356

### DIFF
--- a/refinery/ui/source/js/data-set-explorer/controllers/igv.js
+++ b/refinery/ui/source/js/data-set-explorer/controllers/igv.js
@@ -129,7 +129,11 @@ function IgvCtrl (
     });
   };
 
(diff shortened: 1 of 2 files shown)


mccalluc:
@scottx611x : We decided not to merge this because the old java one would no longer work with your incoming changes. I think the idea was that you would continue your work until the java version would be working again, and then merge this. (Though we could approach it other ways.)

mccalluc:
@scottx611x : on slack you said "I’ve incorporated those changes in my branch", so closing and deleting this PR.

PR #1380, opened 2016-08-30
Mccalluc/pytest selenium

mccalluc:
- Is it ok to change the default behavior of create_user?
- Right now the neo4j warnings fill up the buffer, so if there were selenium errors, they aren't available in the travis UI
- Hard-coded time.sleeps() are the first thing that should be addressed in the next phase of development, but I didn't want to block on that.


selenium/basic_test.py
```diff
@@ -0,0 +1,119 @@
...
+                     'You must provide a Last Name',
+                     'You must provide an Affiliation',
+                     'This field is required')
+
+    stamp = str(time())  # Helps  prevent collisions when running locally.
+    selenium.find_element_by_name('username').send_keys(stamp)
+    selenium.find_element_by_name('first_name').send_keys('first')
+    selenium.find_element_by_name('last_name').send_keys('last')
+    selenium.find_element_by_name('affiliation').send_keys('affiliation')
+    selenium.find_element_by_name('email').send_keys('%s@example.edu' % stamp)
```
mccalluc: -> org

refinery/core/management/commands/create_user.py
```diff
@@ -37,20 +37,18 @@ def handle(self, username, password, email, first_name, last_name,
...
+def init_user(username, password, email, first_name, last_name, affiliation):
     """Create a new User instance with specified values.
     """
     user_object = User.objects.create_user(
         username, email=email, password=password)
     user_object.first_name = first_name
     user_object.last_name = last_name
     user_object.get_profile().affiliation = affiliation
-    user_object.is_active = is_active
+    user_object.is_active = True
```
mccalluc: revert this whole thing and make the user active from the command line.

mccalluc:
Simply added another optional arg for active. Looked at having flagged parameters, but it seemed more complicated, with little additional benefit: The is backwards compatible, which is the most important thing.

PR #1424, opened 2016-09-23
Jkmarx/public group bug fix

### LINKED ISSUE #1017: Easy-to-usify collaboration
Commit: df3ad330c524a01a68bd534c610a7347d918cfa5

### Steps to reproduce
1. Go to http://192.168.50.50:8000/collaboration
2. Create a new group
3. Click on the _cog wheel_ next to your name
4. Click on _demote_ or _remove_

### Observed behavior
1. Both clicks open the same dialog.
2. The first popup doesn't have any _X_ button on the top right and closes with an outside click while the second dialog contains an _X_ button and doesn't close with an outside click.
3. The second dialog is not really user-friendly. It only asks the question "Are you the only member or manager left?"

### Expected behavior
1. Both dialogs should behave the same
2. Instead of asking the user whether he / she is the only user it should tell the user that he / she is the only user.
3. Instead of only providing an _Okay_ button the second dialog should include an option to remove the group to save the user another 3 or 4 clicks.
4. The two buttons _demote_ and _remove_ could also be grayed out and disabled in the first place to avoid confusion.

### LINKED ISSUE #1304: Users should not be able to remove themselves from "Public" group
Commit: 33de42c5dcd1e22fe17e4

### Steps to reproduce
- Click "Collaboration" at the top
- In the "Groups" pane, click the gear beside "Public"
- Click "Leave"

### Observed behavior
Once this is done, the user will no longer have access to public datasets

### Expected behavior
Not sure? An error? or no gear beside "Public"?
(Before diving into this, confirm that we're modelling this appropriately... Maybe groups should only be used for restrictions, and a public dataset just has no groups? That might run into other problems, though.)

### LINKED ISSUE #1305: "Group Editor" dialog is missing cancel button
Commit: 33de42c5dcd1e22fe17e4

### Steps to reproduce
- Click "Collaboration" at the top
- In the "Groups" pane, click the gear beside "Public" (Or beside another group: The ability to edit the public group might be removed. See #1304)
- You'll get a dialog that looks like this:
screen shot 2016-08-08 at 11 37 13 am

### Observed behavior
There is no cancel button, though you can hit escape, or click in the background.

### Expected behavior
There is a cancel button.

jkmarx:
Resolves #1304  and resolves #1305 and resolves #1017

### DIFF
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -2063,6 +2063,11 @@ class ExtendedGroupResource(ModelResource):
         elif request.method == 'DELETE':
             # Removing yourself - must delete to leave if last member.
             if user.id == int(kwargs['user_id']):
+
+                # Prevent users from leaving default public group
+                if ext_group.name == 'Public':
+                    return HttpForbidden('Members may not leave Public group')
+
                 if ext_group.user_set.count() == 1:
                     return HttpForbidden('Last member - must delete group')
 
--- a/refinery/ui/source/js/collaboration/controllers/group-editor.js
+++ b/refinery/ui/source/js/collaboration/controllers/group-editor.js
@@ -18,6 +18,7 @@ function GroupEditorCtrl (
   this.group = group;
   this.authService = authService;
   this.sessionService = sessionService;
+  this.errorFlag = false;
 
   this.close = function () {
     this.$uibModalInstance.dismiss();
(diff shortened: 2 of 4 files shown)


jkmarx:
Remove a couple of alerts and opted for warning. But ui updates, should prevent user from leaving if last member.
screen shot 2016-09-23 at 2 11 47 pm

Final w/o warning for non-manager member.
screen shot 2016-09-23 at 4 00 15 pm

PR #1438, opened 2016-09-29
Mccalluc/remove sphinx

mccalluc:
@scottx611x : Failed with selenium error: https://travis-ci.org/parklab/refinery-platform/builds/163784934#L6183 ... Is this something you've seen before?

at the bottom is the link... maybe it is only supplied on error?: https://saucelabs.com/jobs/ac002077b2284f6db6df711fd533a293


mccalluc:
(Restarting build in case I just was unlucky.)

scottx611x:
@mccalluc Just seeing this today, looks like a restart of the build did the trick???

PR #1511, opened 2016-11-08
Scottx611x/django 1.7 upgrade

### LINKED ISSUE #614: Upgrade to Django 1.7

scottx611x:
Fixes #614


fabfile.py
```diff
@@ -86,19 +86,22 @@ def conf(mode=None):
...
     env.shell_after = \
         "export DJANGO_SETTINGS_MODULE=config.settings.{}".format(mode)
-    env.apache_before = "SetEnv DJANGO_SETTINGS_MODULE config.settings.*"
-    env.apache_after = \
-        "SetEnv DJANGO_SETTINGS_MODULE config.settings.{}".format(mode)
+    env.apache_before = "wsgi_.*.py"
+    if mode == "prod":
+        env.apache_after = "wsgi_prod.py"
+    else:
+        env.apache_after = "wsgi_dev.py"
```
scottx611x: how about "wsgi_{}".format(mode) ?

fabfile.py
```diff
@@ -138,8 +141,7 @@ def update_refinery():
         run("{refinery_app_dir}/manage.py collectstatic --clear --noinput"
             .format(**env))
         run("supervisorctl reload")
-    with cd(env.refinery_project_dir):
```
scottx611x: Do we need this???? (cd)

PR #1529, opened 2016-11-15
Use proper string formatting syntax

scottx611x:
Addresses Type error raised in `update_data_set_index()`


scottx611x:
[Type error raised in update_data_set_index()](https://trello.com/c/jqv84vE4/83-type-error-raised-in-update-data-set-index)

PR #1541, opened 2016-11-29
Jkmarx/data set import delete files

### LINKED ISSUE #1538: Data Import Error Handling
* Specific code commit: 826c479ecd9be6c3d068262bebd89448a4ebd2ef
* Version of the web browser and OS: 54.0.2840.98 and 10.11.6
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): Vagrant VM Prod

- [x] In the lag between 100% and green complete confirmation, the start and cancel button are enabled but don't perform any function
- [x] Add error handling for MD5 errors 
- [x] Allow user to re-add file and upload again.

***** Move to a separate issue **********
- [x] Add deletion capabilities when errors or cancel
- [ ] Add delete capabilities for the chunk upload (back end)

jkmarx:
Resolves #1538


jkmarx:
@scottx611x @hackdna Follow-up to the unusual data.files[0] vs data.files[data.index]
 From https://github.com/blueimp/jQuery-File-Upload/wiki/Frequently-Asked-Questions:
```The template will be rendered for each add call.
As long as the option singleFileUploads is set to true (which is the default), multiple selects/drops get split up into single add calls, so the index will always be 0.```

and 

from https://github.com/blueimp/jQuery-File-Upload/wiki/Options
```By default, each file of a selection is uploaded using an individual request for XHR type uploads.
Set this option to false to upload file selections in one request each.```

I'll add a comment to the code also.

Issue #1554, opened 2016-12-06
Version reference genomes on s3?


mccalluc:
With the work I've been doing for https://github.com/refinery-platform/get-reference-genomes/issues/9, there are some significant changes to the files we're saving on s3. I'm pretty sure I could just run the scripts live and it wouldn't hurt anything... but I could be wrong, and if we have real users we should start to be more cautious.

Right now the structure is: `data/igv-reference/hg19`. I propose:
- Creating `data/igv-reference/v2/hg19` (etc.)
- Running the scripts to populate those directories
- Updating the refinery code.
- Once the new code is deployed, clean up the old `data/igv-reference/hg19`

Does that sound ok? There are [two questions](https://github.com/refinery-platform/get-reference-genomes/issues/9#issuecomment-265273103) that I would like to clarify before I run the script... though we could also just run it and adjust afterwards.

mccalluc:
Question is answered: We will have version-number directories on s3.

PR #1561, opened 2016-12-09
Versioned directories in S3 for reference genomes

mccalluc:
Towards #1554.


mccalluc:
Have not gotten feedback about the new format for the genome files, so closing this for now.

PR #1562, opened 2016-12-12
Jkmarx/data set 2 attribute filter bugs

### LINKED ISSUE #849: allow same name for characteristic and factor value fields in study/assay file
**Reproduce**
Import am ISA-Tab data set that uses the same name both as a  characteristic and factor value fields, e.g. the attached ISA-Tab archive.

[isa_9505_764988_8-3.zip](https://github.com/parklab/refinery-platform/files/90634/isa_9505_764988_8-3.zip)
pasted image at 2016_01_13 10_12 pm

**Observed**
Two facets with the same name will show up in the file browser and facet filtering does not return any files if any of the values in the duplicate facet is selected for filtering.

screen shot 2016-01-13 at 7 14 03 pm

**Expected**
Enough information to distinguish the two facets (e.g. attach type "characteristic" or "facet value") and working facet filtering.

**Remarks**
Discussion with @sjhosui:

```
[10:03 AM] shosui It is both (a) reasonable and (b) correct to have the same attribute both as a factor value and a characteristic.
[10:11 AM] nils Ok
[10:14 AM] shosui i should also say though that it’s not necessary. It is much cleaner and probably more correct if it’s not duplicated(edited)
[10:16 AM] i think sometimes it’s confusing because the factor is defined as something that the experimenter can manipulate to see an outcome. so for example, genotype can be considered both a factor and a characteristic of the source material
[10:40 AM] nils i can see arguments for both cases. a major argument against allowing the same label to be used as a characteristic and a factor value is that is can lead to conflicts when different values are used for the characteristic and for the factor value
```

### LINKED ISSUE #1046: DataSet2 - Attribute Filter Bug
Commit:
6639135bd92cc7ad9dc69a73af742e149bca5628
### Steps to reproduce
1. In DataSet 2 browser, select Data Set 16329: Circulating Breast Tumor Cells Exhibit Dynamic Changes in Epithelial and Mesenchymal Cell Composition
2. Ensure disease state and disease fields are showing
3. Refresh
### Observed behavior

attributefilterbug
### Expected behavior

Attribute filters with more than one field should show.
The disease attributes have the same data.
screen shot 2016-04-26 at 9 11 59 am

jkmarx:
Resolves #1046 and resolves #849 in data set 2 browser.
Also resolves a new bug caused by duplicate field names.


refinery/ui/source/js/file-browser/partials/assay-filters.html
```diff
@@ -30,16 +30,15 @@ <h4 class="panel-title">
...
                 <div class="row">
                   <div class="col-xs-1">
                     <input
                       type="checkbox"
                       name="optionsCheckboxes"
-                      ng-model="FBCtrl.selectedField[field.name]"
-                      ng-change="FBCtrl.attributeSelectionUpdate(
-                      fields.internal_name, field.name)"
+                      ng-model="FBCtrl.attributeSelectedFields[fields.internal_name][field.name]"
+                      ng-change="FBCtrl.attributeSelectionUpdate(fields.internal_name, field.name)"
```
jkmarx: see if I can rename fields to attribute for clarity

refinery/ui/source/js/file-browser/partials/assay-filters.html
```diff
@@ -97,13 +96,13 @@ <h4 class="panel-title">
           <div class="panel-body">
             <div ng-repeat="fieldObj in fields.facetObj track by $index">
               <div class="checkbox container"
-                ng-show="showField(fieldObj, fields.internal_name, 'Analysis')">
+                ng-show="showField(fieldObj, attribute, 'Analysis')">
```
jkmarx: Double check if passing obj vs display_name

PR #1564, opened 2016-12-14
Wrap delete model to check for user permissions.

### LINKED ISSUE #1426: API V1: Group Members can Delete Group
- Specific code commit: b2eb04ebd8bace536b4ba0d5e709a2474c09e5a8
### Steps to reproduce

1. Enable the UI interface to enable the delete button or use curl.
2. Delete a group you are a member of
### Observed behavior

Group Deletes
### Expected behavior

Error, only managers can delete group.

jkmarx:
Resolves #1426

### DIFF
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -2023,6 +2023,17 @@ class ExtendedGroupResource(ModelResource):
         new_ext_group.manager_group.user_set.add(user)
         return bundle
 
+    def obj_delete(self, bundle, **kwargs):
+        user = bundle.request.user
+        ext_group = super(ExtendedGroupResource, self).obj_get(
+            bundle, **kwargs)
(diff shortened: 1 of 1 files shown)


refinery/core/api.py
```diff
@@ -2023,6 +2023,17 @@ def obj_create(self, bundle, **kwargs):
         new_ext_group.manager_group.user_set.add(user)
         return bundle
 
+    def obj_delete(self, bundle, **kwargs):
+        user = bundle.request.user
+        ext_group = super(ExtendedGroupResource, self).obj_get(
```
scottx611x: Don't believe the `super` call is necessary here, could be confusing for someone new to look at

try:
` ext_group = self.obj_get(bundle, **kwargs)`

PR #1570, opened 2016-12-20
Scottx611x/anonymous user fix

scottx611x:
Upon Django 1.7 branch seeing the automated neo4j pre-populated db install for the first time it was observed that the `import_annotations` mgmt. command failed due to there being no Anonymous User.

The AnonymousUser was being created with an `id=1`which led to it being overwritten by the `./manage.py loaddata superuser.json` command. 

Bumping django-guardian to v1.4.1 allows for the reintroduction of the `ANONYMOUS_USER_ID` setting where we can once again pin the AnonymousUser with `id=-1`

Didn't catch this in testing since I had been upgrading existing Django 1.6 instances to 1.7 and an Anonymous user (with id=-1) already existed in the db so the overwriting issue didn't expose itself.


refinery/config/settings/base.py
```diff
@@ -578,3 +578,6 @@ def get_setting(name, settings=local_settings):
     "REFINERY_AUXILIARY_FILE_GENERATION")
 
 REFINERY_TUTORIAL_STEPS = refinery_tutorial_settings
+
+
+ANONYMOUS_USER_NAME = "AnonymousUser"
```
scottx611x: @hackdna I think this is an outdated diff, I've removed that setting since

Issue #779, opened 2015-11-24
Configure Apache using Puppet module


hackdna:
https://forge.puppetlabs.com/puppetlabs/apache

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;OejRcN0q&#x2F;15-implement-single-instance-aws-deployment"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> implement single instance AWS deployment</a>

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;OejRcN0q&#x2F;15-implement-single-instance-aws-deployment"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> implement single instance AWS deployment</a>

hackdna:
Delaying this since not required for AWS deployment work.

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;WKWVJxrD&#x2F;261-configure-https-using-puppet"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Configure HTTPS using Puppet</a>

mccalluc:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;brcH2SDl&#x2F;26-779"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> #779</a>

PR #1572, opened 2017-01-04
Hackdna/large file upload config update

hackdna:
To implement #779


hackdna:
[Update CloudFormation template with settings necessary for large file uploads](https://trello.com/c/h2OdFcA2/357-update-cloudformation-template-with-settings-necessary-for-large-file-uploads)

PR #1575, opened 2017-01-06
Remove references to old IGV

mccalluc:
@scottx611x : I feel like we should get rid of the oldest vis code before trying to move to an even newer generation. It looks like there are migrations relating to the old IGV machinery, and I'm not sure what a good way of tackling that is:
```
$ python manage.py migrate --noinput
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
    utility.execute()
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute
    django.setup()
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/django/__init__.py", line 21, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/django/apps/registry.py", line 85, in populate
    app_config = AppConfig.create(entry)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/django/apps/config.py", line 87, in create
    module = import_module(entry)
  File "/opt/python/2.7.12/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: No module named visualization_manager
```

I'll try putting back the module just as a placeholder so migrations can run...


mccalluc:
@scottx611x : tests pass, but
- I think tests should be deterministic, but Saucelabs also seems to be useful, and I'm not sure how to resolve that tension. Maybe it doesn't need to be run every time: Surround it with an if-then is the bash?
- I pruned a whole lot of code, and have very little idea how most of it was used.

refinery/core/serializers.py
```diff
@@ -121,26 +118,6 @@ def _get_file_extension(self, obj):
     def _get_relative_url(self, obj):
         return obj.get_relative_file_store_item_url() or None
 
-    def _ready_for_igv_detail_view(self, obj):
```
scottx611x: I think this method may still be needed, but this is just off the top of my head.

refinery/core/tests.py
```diff
@@ -2117,8 +2117,7 @@ def setUp(self):
             "auxiliary_nodes": [],
             "is_auxiliary_node": False,
             "file_extension": None,
-            "auxiliary_file_generation_task_state": None,
-            "ready_for_igv_detail_view": None
```
scottx611x: Same in the tests: I think the idea behind `igv_detail_view` was that some FileTypes can be immediately visualized with IGV, but once you start zooming in things break due to the need of some type of index file being needed

refinery/data_set_manager/tests.py
```diff
@@ -1526,8 +1526,7 @@ def setUp(self):
             "auxiliary_nodes": [],
             "is_auxiliary_node": False,
             "file_extension": None,
-            "auxiliary_file_generation_task_state": None,
-            "ready_for_igv_detail_view": None
```
scottx611x: see above comments

jkmarx:
Closed due to reliance of data set 1 on old igv code.

PR #1584, opened 2017-01-12
Jkmarx/jshint rule change

jkmarx:
Update eslint to allow function use before defined. 
Reasons: https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#style-y033


jkmarx:
https://github.com/refinery-platform/refinery-platform/wiki/coding-style-guide#angularjs-and-javascript

PR #1585, opened 2017-01-12
Fix bug due to dependency order typo.

jkmarx:
Fix bug in data set 2 browser for launching IGV.


jkmarx:
We may want this as a hot fix.

PR #1592, opened 2017-01-25
Improve AWS resource tags handling

hackdna:


drj11:
I thought you wanted the tags in `config.yaml`?

hackdna:
@drj11 Yes, this is just a quick fix mainly to enable easier handling of the tags

drj11:
looks good, but I haven't tried launching one. I'll do that now.

deployment/stack.py
```diff
@@ -382,6 +378,22 @@ def main():
...
+    tags = {}
+    try:
+        with open("aws-config/tags.yaml") as f:
+            tags.update(yaml.load(f))
+    except (IOError, yaml.YAMLError) as exc:
+        sys.stderr.write("Error reading AWS resource tags: {}".format(exc))
+        raise ConfigError()
+
+    if 'owner' not in tags:
+        tags['owner'] = os.popen("git config --get user.email").read().rstrip()
```
drj11: I like this.

hackdna: Thanks but I copied this from your code :)

deployment/aws-config/tags.yaml.template
```diff
@@ -0,0 +1,9 @@
+# AWS resource tags
+# as per https://github.com/refinery-platform/refinery-platform/issues/839
+# Copy to tags.yaml before launching a stack
+
+department: dbmi
+environment: dev
+project:
```
drj11: an empty project currently means that using this template without changing it fails.

I think it would be nice if was filled in with something so that it can be used without changing it.

But then I was thinking that maybe it's a good thing that you have to change it.

The error is obscure though: 

```
$ make web-stack
python stack.py > web.json
Configuration saved to s3://refinery-gygexukbykxpm/refinery-config-20170126T1457
TZ=UTC date +refinery-web-%Y%m%dT%H%M > aws-config/stack-name
aws cloudformation create-stack --capabilities CAPABILITY_IAM --stack-name $(cat aws-config/stack-name) --template-body file://web.json

A client error (ValidationError) occurred when calling the CreateStack operation: [/Resources/RDSInstance/Type/Tags/0/Value] 'null' values are not allowed in templates
Makefile:7: recipe for target 'web-stack' failed
make: *** [web-stack] Error 255
```

drj11:
launched one. works.

Use your skill and judgement to address my comments (I'm happy if you decide to change nothing), then merge.

PR #1606, opened 2017-02-02
Add security group

drj11:
As per https://trello.com/c/5FpMkUM2/352-create-a-separate-security-groups-for-elb-ec2-and-rds now have separate security groups for EC2 and RDS instance (and that is separate from the ELB security group).

### DIFF
--- a/deployment/stack.py
+++ b/deployment/stack.py
@@ -373,12 +373,6 @@ def main():
                     "SourceSecurityGroupId": functions.get_att(
                         'ELBSecurityGroup', 'GroupId'),
                 },
-                {
-                    "IpProtocol": "tcp",
-                    "FromPort": "443",
-                    "ToPort": "443",
-                    "CidrIp": "0.0.0.0/0",
-                },
             ],
         })
     )
--- a/deployment/stack.py
+++ b/deployment/stack.py
@@ -124,6 +124,7 @@ def main():
         "AutoMinorVersionUpgrade": False,
         "AvailabilityZone": config['AVAILABILITY_ZONE'],
         "BackupRetentionPeriod": "0",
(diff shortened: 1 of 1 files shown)


deployment/stack.py
```diff
@@ -128,6 +128,7 @@ def main():
         "AutoMinorVersionUpgrade": False,
         "AvailabilityZone": config['AVAILABILITY_ZONE'],
         "BackupRetentionPeriod": "0",
+        "CopyTagsToSnapshot": True,
```
drj11: Not required for this PR, but seemed like a good idea.

drj11:
[Create a separate security groups for ELB, EC2 and RDS](https://trello.com/c/5FpMkUM2/352-create-a-separate-security-groups-for-elb-ec2-and-rds)

hackdna:
Just two things:
* EC2 security group: inbound HTTPS from ELB SG only (not from 0.0.0.0/0)
* RDS security group: no outbound traffic to 0.0.0.0/0 ([discussion](http://serverfault.com/questions/812907/aws-security-group-for-rds-outbound-rules))

drj11:
The circularity thing is I think why the first version that worked didn't have Egress rules. Fixing it now.

drj11:
- [ ]  ELB group: no outbound traffic to 0.0.0.0/0 but only for HTTP(80) to EC2 group
- [ ]  EC2 group: no inbound HTTPS(443)
- [ ]  RDS group: no outbound traffic (discussion)

drj11:
I think I lost ilya's comments by force push (I overwrote the last commit).

Ilya maybe we should drop the hack just to get an empty RDS egress rule set. I agree, especially as I was finding it very fiddly to get working (you have to specify a VPC too).

Issue #1580, opened 2017-01-11
DataSet deletion can break half-way


flekschas:
* Specific code commit: defd3f2bad89a113e21e2aace6ec61cc89347354
* Version of the web browser and OS: mac
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): vm

### Steps to reproduce
1. Delete a data set with multiple versions
2. Or do something else to break the data set deletion

### Observed behavior
* The deletion fails with an IntegrityError with ```<XYZ> could not be deleted. update or delete on table "core_dataset" violates foreign key constraint "content_object_id_refs_id_4e3e498cf2378c3a" on table "core_datasetuserobjectpermission"\nDETAIL:  Key (id)=(<XYZ>) is still referenced from table "core_datasetuserobjectpermission"```

This will cause severe trouble as the data set object still exists but the investigation link is already gone, which will trigger an error when data sets are retrieved via the data_sets API. The problem can't be fixed in the admin UI as the specific db entries need to be removed in PSQL.

### Expected behavior
* Half-way deletions should be avoided using [transactions](https://docs.djangoproject.com/en/1.10/topics/db/transactions/).

PR #1614, opened 2017-02-17
Wrap deletion operations in transactions

scottx611x:
Addresses #1580

### DIFF
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -12,6 +12,7 @@ from django.contrib.auth.models import User
 from django.contrib.sites.models import get_current_site, RequestSite, Site
 from django.core.exceptions import ImproperlyConfigured
 from django.core.urlresolvers import reverse
+from django.db import transaction, IntegrityError
 from django.http import (Http404, HttpResponse, HttpResponseBadRequest,
                          HttpResponseForbidden, HttpResponseNotFound,
                          HttpResponseRedirect, HttpResponseServerError)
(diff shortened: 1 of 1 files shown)


refinery/core/views.py
```diff
@@ -1094,7 +1095,12 @@ def delete(self, request, uuid):
                 content="User {} is not authenticated".format(request.user))
         else:
             try:
-                dataset_deleted = DataSet.objects.get(uuid=uuid).delete()
+                with transaction.atomic():
```
hackdna: There should be a comment explaining why this is necessary

refinery/core/views.py
```diff
@@ -1094,7 +1095,12 @@ def delete(self, request, uuid):
...
+                    dataset_deleted = DataSet.objects.get(uuid=uuid).delete()
```
hackdna: This should be split up into get and delete for readability and to handle `DataSet.DoesNotExist` exception.

refinery/core/views.py
```diff
@@ -1144,7 +1150,12 @@ def delete(self, request, uuid):
                 content="User {} is not authenticated".format(request.user))
         else:
             try:
-                analysis_deleted = Analysis.objects.get(uuid=uuid).delete()
+                with transaction.atomic():
```
hackdna: Same as above for DataSet

refinery/core/views.py
scottx611x: Doubdle check DataSet.delete() may be handling exceptions there already

refinery/core/views.py
scottx611x: double check Analysis.delete() as well

scottx611x:
Transaction handling should really happen in Dataset.delete() & Analysis.delete()

hackdna:
also "Avoid catching exceptions inside atomic!" https://docs.djangoproject.com/en/1.7/topics/db/transactions/

mccalluc:
@scottx611x : This is pretty old: is it still active?

scottx611x:
@mccalluc  Nope haven't touched this in awhile, the transaction handling needed here is much more involved than the simple fix in this PR

scottx611x:
Closing. This is really outdated.

PR #1620, opened 2017-02-28
Scottx611x/td api

scottx611x:
First iteration of ToolDefinitions API code.

- Allows for GETs and generation of sample ToolDefs w/ nested file_relationships
- [x] Ordering of imports
- [x] Usecase for many allowed filetypes?
- [x] DRF api required auth?
- [x] Placement of DRF urls?


refinery/tool_manager/models.py
```diff
@@ -0,0 +1,136 @@
...
+    name = models.TextField(max_length=100, blank=False,
+                            null=False)
+    description = models.TextField(max_length=500, blank=False,
+                                   null=False)
+    is_editable = models.BooleanField(default=False)
+    value_type = models.CharField(choices=VALUE_TYPES, max_length=25,
+                                  blank=False, null=False)
+    default_value = models.TextField(max_length=100, blank=False, null=False)
+    # ID of Galaxy tool that the parameter belongs to
+    galaxy_tool_id = models.TextField(max_length=300, blank=False, null=False)
```
hackdna: Galaxy-related fields should not be a part of the generic Parameter model since not all the tools will be coming from Galaxy.

scottx611x: How about: 
```
class GalaxyParameter(Parameter):
    # ID of Galaxy tool that the parameter belongs to
    galaxy_tool_id = models.TextField(max_length=300, blank=False,
                                        null=False)
    galaxy_tool_parameter = models.TextField(
        max_length=100, blank=False, null=False)
```

hackdna: I think that should work. Btw, no need for `null=False` - it's a default https://docs.djangoproject.com/en/1.7/ref/models/fields/#django.db.models.Field.null

refinery/tool_manager/models.py
```diff
@@ -62,16 +63,18 @@ class FileRelationship(models.Model):
...
-        max_length=100, choices=RELATIONSHIP_TYPES, blank=False, null=False)
-    file_relationship = models.ManyToManyField("self", symmetrical=False,
-                                               null=True, blank=True)
+    name = models.TextField(max_length=100)
+    value_type = models.CharField(max_length=100, choices=RELATIONSHIP_TYPES)
+
+    # NOTE: `symmetrical=False` is not very common. It's necessary for the
+    # self-referential M2M below. See: http://bit.ly/2mpPQfT
+    file_relationship = models.ManyToManyField(
+        "self", symmetrical=False, null=True, blank=True)
```
hackdna: Why is `symmetrical=False` necessary here?

refinery/tool_manager/models.py
```diff
@@ -122,15 +125,12 @@ class ToolDefinition(models.Model):
...
-                            null=False)
-    description = models.TextField(unique=True, max_length=500, blank=False,
-                                   null=False)
-    tool_type = models.CharField(max_length=100, choices=TOOL_TYPES,
-                                 blank=False, null=False)
-    file_relationship = models.ForeignKey("FileRelationship")
-    output_files = models.ManyToManyField("OutputFile")
-    parameters = models.ManyToManyField("Parameter")
+    name = models.TextField(unique=True, max_length=100)
+    description = models.TextField(unique=True, max_length=500)
```
hackdna: Why is `unique=True` necessary for `name` and `description` fields?

refinery/tool_manager/models.py
```diff
@@ -27,24 +29,23 @@ class Parameter(models.Model):
         (FILE, "file")
     )
     uuid = UUIDField(unique=True, auto=True)
-    name = models.TextField(max_length=100, blank=False,
-                            null=False)
-    description = models.TextField(max_length=500, blank=False,
-                                   null=False)
+    name = models.TextField(max_length=100)
+    description = models.TextField(max_length=500)
     is_editable = models.BooleanField(default=False)
```
hackdna: How about something like `is_user_adjustable` or just `user_adjustable`?

refinery/tool_manager/models.py
scottx611x: We want to provide some asymmetry to the FileRelationship nesting. Without this, the bottom-most FileRelationship could know about the top-most one, and things could get messy. This way `FileRelationships` know about their nested children only.

refinery/tool_manager/models.py
scottx611x: I think `name` would be good to keep unique so e don't have 100 "TEST" workflows show in the UI if someone is naming things the same Galaxy-side. Constraints on `description` can definitely go.

hackdna: OK but keep in mind that you will have to:
1. document this feature
1. write additional code to handle violations of this constraint

scottx611x: @hackdna sure, I'm "failing fast" now and not commiting any data upon the running of `generate_tool_definitions` command unless things look perfect. I'd like validation of workflow annotations to be polished well since I feel that with this flexibility we are giving users, we are also giving them more ways to potentially make errors.

Looking into workflow annotation validation using JSON Schema since we already have schemas defined in swaggerhub.

@jkmarx Also I just checked Galaxy and there is no enforcement of unique Workflow naming there.

PR #1650, opened 2017-03-27
Scottx611x/run selenium on travis only

scottx611x:


refinery/selenium_testing/tests.py
```diff
@@ -26,6 +29,10 @@ class SeleniumTestBase(StaticLiveServerTestCase):
 
     def setUp(self, site_login=True, initialize_guest=True,
               public_group_needed=True):
+
+        # Start a pyvirtualdisplay for geckodriver to interact with
+        self.display = Display(visible=0, size=(1900, 1080))
```
hackdna: Why didn't we have this before? How was the size chosen?

refinery/selenium_testing/tests.py
```diff
@@ -1,23 +1,26 @@
...
-# Start a pyvirtualdisplay for geckodriver to interact with
-display = Display(visible=0, size=(1900, 1080))
-display.start()
+    assert_body_text, assert_text_within_id, delete_from_ui, login,
+    MAX_WAIT, wait_until_class_visible, wait_until_id_clickable,
+    wait_until_id_visible
+)
 
 
+@skipIf("prod" not in os.getenv("DJANGO_SETTINGS_MODULE"),
```
hackdna: As I mentioned earlier, I think this is unnecessary as we have a way to run unit tests separately from Selenium tests. This also might trip us up down the road (why are Selenium tests are not running?). I think a more flexible solution is provided in Django 1.10 (https://docs.djangoproject.com/en/1.10/topics/testing/tools/#tagging-tests).

scottx611x: @hackdna We don't currently have way to run unit tests separately from Selenium tests. Sure, you can run individual app's test suites, but this isn't a good solution, especially in the case of Refinery, because our apps are tightly coupled and changes in one app can break tests in another.

I think this solution is simple and works fine. If one wants to explicitly run selenium tests locally, switch to prod mode. Otherwise, selenium based tests are run on Travis every time you push.

As for Django 1.10, yes I mentioned that tagging feature to you... but again the ability to use that is out of our reach for awhile.

refinery/selenium_testing/tests.py
scottx611x: We did have this before, See here: https://github.com/refinery-platform/refinery-platform/pull/1650/files/c733de5aea8aa4871c925bf8cd17bf705758b71b#diff-f4bed93e08ec88c8c01e26197a799d9bL17

I've moved this into the `setUp()` and the stopping of said virtual display into `tearDown()`

hackdna: OK, and how was the size parameter chosen?

scottx611x: @hackdna I just chose a screen resolution that I envisioned would be about the norm of what most people use nowadays. If we make the resolution too small, pages wont render properly, and tests will probably fail.

refinery/selenium_testing/tests.py
hackdna: I still think that having different test runner behavior based on an environment variable is potentially confusing. Even if someone forgets to run all the tests locally, the CI system will catch that.

PR #1653, opened 2017-03-28
Scottx611x/td api fixes

scottx611x:
Small fixes for ToolDefinition API code brought to light by writing documentation for it.


refinery/tool_manager/utils.py
```diff
@@ -282,11 +282,13 @@ def get_workflow_list():
     return workflow_list
 
 
-def mock_get_workflow_list():
+def mock_get_workflow_list(filename):
```
hackdna: This function simply reads a JSON file content and loads it into a Python object, so I'd name it accordingly and make it re-useable by allowing to specify any file path.

refinery/tool_manager/tests.py
```diff
@@ -413,13 +413,26 @@ def test_invalid_workflow_step_annotation_c(self):
             )
 
     def test_generate_tool_definitions_management_command(self):
-        with mock.patch('tool_manager.utils.get_workflow_list',
-                        side_effect=mock_get_workflow_list):
+        with mock.patch(
+            'tool_manager.utils.get_workflow_list',
+            side_effect=[
```
scottx611x: @hackdna: I could also  just nix the `mock_get_workflow_list()` and do the `json.loads()` in this `side_effects` array? thoughts?

hackdna: Good point, that's probably even better in this case.

refinery/tool_manager/tests.py
```diff
@@ -413,13 +412,57 @@ def test_invalid_workflow_step_annotation_c(self):
             )
 
     def test_generate_tool_definitions_management_command(self):
-        with mock.patch('tool_manager.utils.get_workflow_list',
-                        side_effect=mock_get_workflow_list):
+        invalid_workflows = open(
+            "tool_manager/test_data/invalid_galaxy_workflows.json"
```
hackdna: DRY: `tool_manager/test_data/` should be stored in a variable.

PR #1657, opened 2017-03-29
Fix typo in method call.

jkmarx:
On the collaboration page: 
Minor bug fix when there's a server error.


jkmarx:
@scottx611x ??

PR #1664, opened 2017-04-04
Scottx611x/docker puppet install

scottx611x:


deployment/Puppetfile
```diff
@@ -1,5 +1,6 @@
 forge "https://forgeapi.puppetlabs.com"
 
+mod 'garethr-docker', '5.3.0'
```
hackdna: Why not use the [puppetlabs/docker_platform](https://forge.puppet.com/puppetlabs/docker_platform)?

deployment/refinery-modules/refinery/manifests/docker_.pp
```diff
@@ -0,0 +1,5 @@
+class refinery::docker_ {
```
hackdna: Why not `refinery::docker`?

deployment/refinery-modules/refinery/manifests/init.pp
```diff
@@ -229,6 +226,7 @@
     Class["ui"],
     Class["solr"],
     Class["neo4j"],
+    Class["docker"],
```
hackdna: Why does supervisor require docker?

deployment/refinery-modules/refinery/manifests/selenium.pp
```diff
@@ -1,6 +1,6 @@
 class refinery::selenium {
   $geckodriver_version = 'v0.15.0'
-  $filename = "geckodriver-${geckodriver_version}-linux32.tar.gz"
+  $filename = "geckodriver-${geckodriver_version}-linux64.tar.gz"
```
hackdna: Why is this necessary?

deployment/refinery-modules/refinery/manifests/docker_.pp
```diff
@@ -0,0 +1,5 @@
...
+  class { 'docker':
```
hackdna: This should be `class { '::docker':` [to indicate global namespace explicitly](http://serverfault.com/questions/421023/why-does-my-namespace-and-class-name-conflict)

deployment/Puppetfile
scottx611x: Not sure, it looks exactly the same as: `garethr-docker`
Actually it depends on it: https://forge.puppet.com/puppetlabs/docker_platform/dependencies

deployment/refinery-modules/refinery/manifests/docker_.pp
scottx611x: Because there is already a class`docker` and Puppet complains about a naming conflict

deployment/refinery-modules/refinery/manifests/docker_.pp
scottx611x: Okay, should this be changed throughout our manifests? I was just following the syntax that we had used throughout

deployment/refinery-modules/refinery/manifests/selenium.pp
scottx611x: Because Geckodriver is not providing 32 bit builds in its newer releases anymore

deployment/refinery-modules/refinery/manifests/init.pp
scottx611x: I don't think supervisor explicitly needs docker, but i don't think we want to start up everything since our app will depend on docker working.

deployment/refinery-modules/refinery/manifests/docker_.pp
scottx611x: Also this answer is from 2012, is it still relevant?

deployment/Puppetfile
hackdna: The benefit of the puppetlabs modules is that they are provided by the Puppet team. Also, garethr-docker was last updated about a year ago and it supports only up to Ubuntu 14.04. The puppetlabs version was last updated a week ago and supports Ubuntu 16.04. Basically, it looks like it is probably going to replace garethr-docker.

deployment/refinery-modules/refinery/manifests/init.pp
hackdna: I'd remove that since there is no explicit dependency and docker installation will still happen by the time the Refinery manifest is applied.

deployment/refinery-modules/refinery/manifests/docker_.pp
hackdna: Yes, it's the reason you got the naming error when trying to use `refinery::docker` on the line above. Here's the source in Puppet docs (still relevant as of Puppet v3.5 and we are using v3.4.3): https://docs.puppet.com/puppet/3.5/lang_namespaces.html#workaround

hackdna: And yes, it should be changed in the rest of the manifests but in a separate branch.

deployment/refinery-modules/refinery/manifests/docker.pp
```diff
@@ -0,0 +1,5 @@
+class refinery::docker {
+  class { '::docker':
+    docker_users => [$app_user]
```
hackdna: `$::app_user`

hackdna: `$::app_user`

PR #1671, opened 2017-04-12
Hackdna/ec2 ips django settings

hackdna:
* EC2 instance health checks are done using the default vhost instead of the Refinery application vhost
* Access to the Refinery application is via ELB only


hackdna:
[Adding EC2 instance public and private IPs to ALLOWED_HOSTS is no longer necessary](https://trello.com/c/1Wr7Ur6F/134-adding-ec2-instance-public-and-private-ips-to-allowed-hosts-is-no-longer-necessary)

PR #1679, opened 2017-04-19
Hackdna/postgres conf update

hackdna:
Remove PostgreSQL server from EC2 instance, make it easier to connect to PostgreSQL server running on Vagrant via SSH from the host, and streamline dependency handling for Python.


deployment/refinery-modules/refinery/manifests/postgresql.pp
```diff
@@ -0,0 +1,32 @@
...
+  else {
+    class { '::postgresql::globals':
+      version  => '9.3',
+      encoding => 'UTF8',
+      locale   => 'en_US.utf8',
+    }
+    ->
+    class { '::postgresql::server':
+      # to make remote connections via SSH tunnel easier
+      ipv4acls => ["host all ${::app_user} 127.0.0.1/32 trust"],
```
drj11: Technically, the 127 net (RFC 1918) is a /8 (try, for example, ping 127.5.5.5).

I don't think that's likely to be an issue.

deployment/refinery-modules/refinery/manifests/postgresql.pp
```diff
@@ -0,0 +1,32 @@
+class refinery::postgresql {
+
+  if $::deployment_platform == 'aws' {
```
drj11: I had no idea you could do this with puppet.

hackdna: This is just a global variable that we set, like `$::app_user` for example. The plan is to replace all the globals with class parameters to make classes self-contained.

deployment/refinery-modules/refinery/manifests/postgresql.pp
hackdna: Right but in this case we only care about 127.0.0.1 (localhost).

PR #1680, opened 2017-04-19
Require STACK_NAME in aws-config

drj11:
`STACK_NAME` is now required in aws-config.

And it is used to name the stack and the ec2 instances.

There is more to do here, this is just a start.
Quite a bit of this will change when we get rid of the `Makefile`.

### DIFF
--- a/deployment/stack.py
+++ b/deployment/stack.py
@@ -26,8 +26,6 @@ Usually invoked via the Makefile: make web-stack
 # CloudInit
 #   https://help.ubuntu.com/community/CloudInit
 
-from __future__ import print_function
-
 import base64
 import datetime
 import json
@@ -60,7 +58,7 @@ def main(argv=None):
     # For now, just the "name" command,
     # which we plan to remove, when the Makefile goes.
     if len(argv) > 1 and argv[1] == "name":
-        print(stack_name)
+        sys.stdout.write("{}\n".format(stack_name))
         return 0
 
     # We discover the current git branch/commit so that the deployment script
--- a/deployment/stack.py
+++ b/deployment/stack.py
@@ -26,6 +26,8 @@ Usually invoked via the Makefile: make web-stack
 # CloudInit
 #   https://help.ubuntu.com/community/CloudInit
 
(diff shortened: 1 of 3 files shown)


drj11:
Merge at will.

deployment/stack.py
```diff
@@ -44,12 +46,22 @@ class ConfigError(Exception):
     pass
 
 
-def main():
+def main(argv=None):
+    if argv is None:
+        argv = sys.argv
```
hackdna: Why not use `sys.argv` directly?

deployment/stack.py
```diff
@@ -44,12 +46,22 @@ class ConfigError(Exception):
...
     derive_config(config)
 
-    unique_suffix = datetime.datetime.utcnow().strftime("%Y%m%dT%H%M")
+    stack_name = config['STACK_NAME']
+
+    # :todo: expand this into a proper command parser.
+    # For now, just the "name" command,
+    # which we plan to remove, when the Makefile goes.
+    if len(argv) > 1 and argv[1] == "name":
+        print(stack_name)
```
hackdna: What about `sys.stdout` instead of `print`?

deployment/stack.py
drj11: Guido's main

http://www.artima.com/weblogs/viewpost.jsp?thread=4829

it has the benefit that you import `stack.py` from an interactive interpreter and then call `main()` by hand with a specific set of arguments.
Admittedly not that useful in this case, but I have used it elsewhere.

So it's habit more than anything else.

deployment/stack.py
drj11: I don't mind. Is there a style recommendation (I admit to not checking).

hackdna: There is now :) https://github.com/refinery-platform/refinery-platform/wiki/Coding-Style-Guide

hackdna:
This is great, looking forward to merging this once requested changes are implemented.

drj11:
Good to go. Again.

PR #1685, opened 2017-04-24
Remove Makefile

drj11:
The `cloudformation create-stack` operation is moved into the Python script.

To create the stack:

`python stack.py create`

(the `Makefile` has been stubbed out, but works for old-timers like me; we can get rid of it soon)


drj11:
Have modified https://github.com/refinery-platform/refinery-platform/wiki/AWS-deployment to say `python stack.py create`

PR #1690, opened 2017-04-27
tag EC2 instances for backups

drj11:
Ensure EC2 instances are tagged so that their volumes get backed up (if EBS Snapshot Scheduler is running).

### DIFF
--- a/deployment/stack.py
+++ b/deployment/stack.py
@@ -120,6 +120,9 @@ def make_template(config, config_yaml):
 
     cft = core.CloudFormationTemplate(description="Refinery Platform main")
 
+    # This parameter tags the EC2 instances, and is intended to be used
+    # with the AWS Reference Implementation EBS Snapshot Scheduler:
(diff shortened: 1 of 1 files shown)


drj11:
Merge at Will

deployment/stack.py
```diff
@@ -143,6 +148,20 @@ def make_template(config, config_yaml):
 
     cft = core.CloudFormationTemplate(description="refinery platform.")
 
+    # This parameter tags the EC2 instances,
+    # and is intended to tbe used with the AWS Reference Implementation
+    # EBS Snapshot Scheduler:
```
hackdna: Looks like these three comment lines could be reformatted to become two :)

drj11: Probably been reading http://rhodesmill.org/brandon/2012/one-sentence-per-line/ too much. Will change.

drj11:
I've rebased the branch and `push -f`ed to address the review.

PR #1692, opened 2017-04-27
Scottx611x/tlc updates

scottx611x:
- Add TLC validation tests
- Integrate launching of HiGlass Vis Tool in tests
- Provide machinery for front end code


refinery/tool_manager/utils.py
```diff
@@ -375,3 +427,56 @@ def get_visualization_annotations_list():
...
+    nesting_types = nesting_dict[nesting_level]["types"]
+    nesting_contents = nesting_dict[nesting_level]["contents"]
+
+    for item in nested_structure:
+        nesting_types.add(type(item))
+        nesting_contents.append(item)
+
+    if len(nesting_types) != 1:
+        raise RuntimeError(
+            "LIST/PAIR structure is not balanced! {}".format(nesting_contents)
```
scottx611x: !

refinery/config/settings/base.py
```diff
@@ -600,3 +620,5 @@ def get_setting(name, settings=local_settings):
 
 # To avoid Port conflicts between LiveServerTestCases http://bit.ly/2pb64KN
 os.environ["DJANGO_LIVE_TEST_SERVER_ADDRESS"] = "localhost:10000-12000"
+
+DJANGO_DOCKER_ENGINE_URL_ROOT = "visualizations"
```
hackdna: May be `DJANGO_DOCKER_ENGINE_BASE_URL` to keep the nomenclature consistent with other similar uses? Also, why not `/visualizations` since it is a base URL?

refinery/file_store/models.py
```diff
@@ -556,6 +556,14 @@ def terminate_file_import_task(self):
                          self.import_task_id, e)
 
 
+@receiver(pre_delete, sender=FileStoreItem)
+def _remove_datafile(sender, instance, *args, **kwargs):
```
hackdna: There is already a method for this: `_delete_datafile()`

refinery/tool_manager/models.py
```diff
@@ -237,38 +243,88 @@ def __str__(self):
...
+                return JsonResponse(
+                    {
+                        "tool_url": self.get_relative_container_url()
+                    }
+                )
 
         if self.get_tool_type() == ToolDefinition.WORKFLOW:
             raise NotImplementedError
 
+    def get_django_docker_engine_manager(self):
```
hackdna: Does this need to be a class method?

refinery/tool_manager/models.py
```diff
@@ -237,38 +243,88 @@ def __str__(self):
...
+        """
+        Helper method to return the proper managerial class for
+        django_docker_engine
+        """
+        if "aws" in os.getenv("DJANGO_SETTINGS_MODULE"):
```
hackdna: This check essentially depends on a file name which makes it brittle.

refinery/tool_manager/models.py
```diff
@@ -237,38 +243,88 @@ def __str__(self):
...
+        :param matched_object: re.sub() MatchedObject instance
+        :return: Full url pointing to a Node's respective FileStoreItem's
+        datafile's location
+        """
+        try:
+            node = Node.objects.get(uuid=matched_object.group(0))
+        except (Node.DoesNotExist, Node.MultipleObjectsReturned):
+            # Raise these errors so that they propagate up the stack and
+            # properly nullify the current database transaction
+            raise
```
hackdna: There is no need to handle these exceptions if you are simply re-raising.

refinery/tool_manager/models.py
```diff
@@ -237,38 +243,88 @@ def __str__(self):
...
+            r'[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}',
+            self.populate_url_from_node_uuid,
+            self.file_relationships
+        )
+        self.file_relationships = populated_file_relationships
+        self.save()
 
-    def populate_url_from_node_uuid(self):
-        raise NotImplementedError
+    @staticmethod
```
hackdna: Why does this need to be a `@staticmethod`?

refinery/tool_manager/utils.py
```diff
@@ -108,6 +110,56 @@ def create_tool_definition(annotation_data):
...
+        # Create a unique container name that adheres to docker's specs
+        tool.container_name = "{}-{}".format(
+            tool.name.replace(" ", ""),
+            tool.uuid
+        )
+
+    if tool.get_tool_type() == ToolDefinition.WORKFLOW:
+        try:
+            tool.output_files = tool_launch_configuration["output_files"]
+        except KeyError:
```
hackdna: Probably no need to handle this if you are simply raising again

refinery/tool_manager/utils.py
```diff
@@ -300,24 +352,24 @@ def validate_tool_launch_configuration(tool_launch_config):
...
-        )
-    except ValueError as e:
-        raise RuntimeError(
-            "Tool launch configuration: {} is not valid JSON: {}".format(
-                tool_launch_config,
-                e
-            )
-        )
+        validate(tool_launch_config, schema, resolver=resolver)
     except ValidationError as e:
```
hackdna: Probably no need to handle this if you are simply raising again

refinery/tool_manager/utils.py
```diff
@@ -300,24 +352,24 @@ def validate_tool_launch_configuration(tool_launch_config):
...
         raise RuntimeError(
             "Tool launch configuration is not properly configured: {}".format(
                 e
             )
         )
+    # Assert that the data structure being sent over is able to be evaluated
+    #  as a Pythonic Data structure
+    try:
+        nesting = eval(tool_launch_config["file_relationships"])
+    except SyntaxError as e:
```
hackdna: Probably no need to handle this if you are simply raising again

refinery/tool_manager/utils.py
```diff
@@ -375,3 +427,56 @@ def get_visualization_annotations_list():
...
+                                    nesting_level=0):
+    """
+    Recursive method to ensure that a ToolLaunchConfiguration's
+    `file_relationships` string is a proper representation of a LIST/PAIR
+    structure
+    :raises: RuntimeError if an inappropriately configured `file_relationships`
+     string is detected
+    """
+
+    if nesting_dict is None:
```
hackdna: This probably should be combined with the try block below

refinery/tool_manager/utils.py
```diff
@@ -375,3 +427,56 @@ def get_visualization_annotations_list():
...
+        nesting_dict = {
+            nesting_level: {
```
hackdna: The arg `nesting_level=0` is an int but here it's a dict?

refinery/tool_manager/models.py
scottx611x: It made sense to me to keep it here rather than bloating `utils` because it is only used during `Tool.launch()`

refinery/tool_manager/models.py
scottx611x: @hackdna Do you a suggestion as to how I could improve this check?

refinery/tool_manager/models.py
scottx611x: This methods has a logical connection to the Tool class but does not need a reference to `self` to operate.

refinery/tool_manager/utils.py
scottx611x: `nesting_level` is just a key, no assignment being done here

refinery/tool_manager/utils.py
scottx611x: I think this re-raise could stay because we are provide a more meaningful message to the RuntimeError if a KeyError occurs

refinery/tool_manager/utils.py
scottx611x: Again, I think catching a specific error, then raising Runtime error with a useful message is helpful here

refinery/tool_manager/utils.py
scottx611x: Again, I think catching a specific error, then raising Runtime error with a useful message is helpful here

refinery/tool_manager/utils.py
hackdna: Yes, it is a key with a value of type dict (`{}`) being assigned here.

refinery/tool_manager/models.py
hackdna: I don't think it belongs in a class because it doesn't need to operate on any of the class properties.

refinery/tool_manager/models.py
hackdna: No but I'm sure you could find some options with a little bit of research :)

refinery/tool_manager/utils.py
scottx611x: I think its okay to keep them separate here. 

The  first `if` is just instantiating the `nesting_dict` to be reused in recursive calls.

Since each `nesting_level` could have many objects, I only want to set the `nesting_level` key once, which is why the KeyError check is helpful.

refinery/tool_manager/utils.py
scottx611x: @hackdna I'm not changing the value of `nesting_level` here. It is still an integer.

refinery/tool_manager/models.py
```diff
@@ -237,38 +243,88 @@ def __str__(self):
...
+    def populate_url_from_node_uuid(matched_object):
```
hackdna: The argument should be a UUID to match the name of the function and to make the code more readable and testable.

refinery/tool_manager/models.py
hackdna: This function takes a UUID and returns a file URL. What is the connection with the Tool class?

refinery/tool_manager/models.py
scottx611x: re.sub is providing me with a matched object instance in this case,  I think it would make more sense to rename the method rather than the argument to then still have to say:
`node = Node.objects.get(uuid=uuid.group(0))`

refinery/tool_manager/utils.py
hackdna: OK, is this message actually used anywhere?

refinery/tool_manager/models.py
scottx611x: This method is only called when updating the `file_relationships` field of the Tool class.
I think adding methods to utils.py files isn't always necessary unless they are designed to be of use in other places

refinery/tool_manager/utils.py
scottx611x: Yes in tests, Since RuntimeErros can be raised in many places during tool creation, its helpful to have a message associated so that I can assert that the proper exception is being hit.

refinery/file_store/migrations/0003_higlass_filetypes_and_fileextensions.py
```diff
@@ -0,0 +1,82 @@
+# encoding: utf8
+from django.db import migrations, transaction, IntegrityError
```
jkmarx: Are we case sensitive in the order? I can't remember the style guide.

refinery/tool_manager/migrations/0016_auto_20170424_1753.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
```
jkmarx: order

refinery/file_store/migrations/0003_higlass_filetypes_and_fileextensions.py
scottx611x: @jkmarx migrations are excluded from our flake8 checks because the auto generated files don't adhere to PEP8

refinery/tool_manager/migrations/0016_auto_20170424_1753.py
scottx611x: same

refinery/tool_manager/models.py
```diff
@@ -259,16 +269,53 @@ def get_relative_container_url(self):
...
+        their respective FileStoreItem's urls. No error handling here since
+        this method is only called in an atomic transaction.
+        """
+        node_uuids = re.findall(
+            r'[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}',
+            self.file_relationships
+        )
 
-    def populate_url_from_node_uuid(self):
+        for uuid in node_uuids:
```
hackdna: Why not keep this as a separate function?

refinery/config/settings/base.py
```diff
@@ -600,3 +619,7 @@ def get_setting(name, settings=local_settings):
 
 # To avoid Port conflicts between LiveServerTestCases http://bit.ly/2pb64KN
 os.environ["DJANGO_LIVE_TEST_SERVER_ADDRESS"] = "localhost:10000-12000"
+
+DJANGO_DOCKER_ENGINE_BASE_URL = "visualizations"
```
hackdna: Since it's a base URL it should probably start with a leading slash.

scottx611x: @hackdna Django url patterns always have the leading slash by default
See: https://docs.djangoproject.com/en/1.7/topics/http/urls/#example

Django yells at me if I do this (when specifying a leading slash prior):
```
url(
    r'^{}/'.format(settings.DJANGO_DOCKER_ENGINE_BASE_URL),
    include('django_docker_engine.urls')
)
```
Because the endpoint is then configured as `//visualizations/`

refinery/tool_manager/models.py
scottx611x: @hackdna What is the need for the separate function here?

hackdna: I think the functionality of getting a file URL from a Node UUID could be useful elsewhere down the road. Not necessary strictly speaking though.

refinery/data_set_manager/tests.py
```diff
@@ -1281,6 +1289,53 @@ def test_update_attribute_order_ranks_invalid(self):
         attribute_list = AttributeOrder.objects.filter(assay=self.assay)
         self.assertItemsEqual(old_attribute_list, attribute_list)
 
+    def test_get_file_url_from_node_uuid(self):
+        test_file_a = StringIO()
+        test_file_a.write('Coffee is great.\n')
+        file_store_item_a = FileStoreItem.objects.create(
+            datafile=InMemoryUploadedFile(
```
hackdna: Perhaps a better alternative is to delete all the FileStoreItems in a tearDown method explicitly and/or use [SimpleUploadedFile](http://stackoverflow.com/a/20508621) or [mock the File instance](http://stackoverflow.com/a/4547203).

refinery/data_set_manager/tests.py
```diff
@@ -1281,6 +1289,53 @@ def test_update_attribute_order_ranks_invalid(self):
         attribute_list = AttributeOrder.objects.filter(assay=self.assay)
         self.assertItemsEqual(old_attribute_list, attribute_list)
 
+    def test_get_file_url_from_node_uuid(self):
```
hackdna: This should be refactored into a class with three separate tests.

refinery/data_set_manager/tests.py
scottx611x: There is something in Refinery that still writes files to disk regardless of what `django.core.files.uploadedfile` class is used. Not sure what or where that is happening. 
I'm already deleting FileStoreitems explicitly in this TestCase

refinery/tool_manager/utils.py
```diff
@@ -300,24 +352,24 @@ def validate_tool_launch_configuration(tool_launch_config):
...
     except ValidationError as e:
         raise RuntimeError(
             "Tool launch configuration is not properly configured: {}".format(
                 e
             )
         )
+    # Assert that the data structure being sent over is able to be evaluated
+    #  as a Pythonic Data structure
+    try:
+        nesting = eval(tool_launch_config["file_relationships"])
```
hackdna: I'd would avoid using `eval()` and perform error handling in `parse_file_relationship_nesting()` instead. Also, this applies to another instance of eval above.

scottx611x: We are using `eval()` here because of our current limitations:
- Not being able to store proper JSON in our db due to out of date Django and 3rd party Libraries
- There is no concept of a tuple in the JSON format, so the `()` that represent our `PAIR`s are lost

I have notes to move away from this approach once we have alleviated some technical debt, but for the time being I've implemented error handling as best as I could and have provided full test coverage: 
https://codecov.io/gh/refinery-platform/refinery-platform/pull/1692/src/refinery/tool_manager/utils.py?before=refinery/tool_manager/utils.py#D1-432

https://codecov.io/gh/refinery-platform/refinery-platform/pull/1692/src/refinery/tool_manager/utils.py?before=refinery/tool_manager/utils.py#D1-362

PR #1710, opened 2017-05-02
Remove the explicit choice of AZ

### LINKED ISSUE #1709: Avoid explicitly setting Availability Zone.
Setting AZ for RDS and EC2 is not required,
and setting it explicitly sometimes leads to capacity errors on deployment.

So changing the stack to avoid picking an AZ will be more reliable and simpler.

drj11:
fixes #1709

### DIFF
--- a/deployment/stack.py
+++ b/deployment/stack.py
@@ -118,7 +118,6 @@ def make_template(config, config_yaml):
     rds_properties = {
         "AllocatedStorage": "5",
         "AutoMinorVersionUpgrade": False,
-        "AvailabilityZone": config['AVAILABILITY_ZONE'],
         "BackupRetentionPeriod": "0",
         "CopyTagsToSnapshot": True,
         "DBInstanceClass": "db.t2.small",       # todo:?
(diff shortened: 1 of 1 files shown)


drj11:
I think this may also fix the problem that EBS volumes and EC2 instances get deleted in the wrong order, because now there is a dependency between the volume and the instance (the volume depends on the instance being created first because it needs to know the instance's AZ).

hackdna:
[deleting AWS web stack fails due to error detaching data volume](https://trello.com/c/QFDr8ISz/156-deleting-aws-web-stack-fails-due-to-error-detaching-data-volume)

deployment/stack.py
```diff
@@ -147,10 +146,11 @@ def make_template(config, config_yaml):
         )
 
     volume_properties = {
-        'AvailabilityZone': config['AVAILABILITY_ZONE'],
         'Encrypted': True,
         'Size': config['DATA_VOLUME_SIZE'],
         'Tags': load_tags(),
+        'AvailabilityZone': functions.get_att(
+            'WebInstance', 'AvailabilityZone'),
```
hackdna: A nitpick: if an open parenthesis is on a line by itself then the closing one should be as well

deployment/stack.py
```diff
@@ -439,6 +437,8 @@ def make_template(config, config_yaml):
             'SecurityGroups': [
                 functions.get_att('ELBSecurityGroup', 'GroupId')],
             "Tags": instance_tags,  # todo: Should be different?
+            'AvailabilityZones': [
+                functions.get_att('WebInstance', 'AvailabilityZone')],
```
hackdna: A nitpick: if an open bracket is on a line by itself then the closing one should be as well

deployment/stack.py
drj11: Oh I did not know that. Shame `flake8` doesn't pick it up.

hackdna: Technically, it's not a part of PEP8 but we decided this would increase readability. I'll update the coding style guide.

Issue #1703, opened 2017-05-01
Search discrepancy: cdx vs. cdx2


scottx611x:
* Specific code commit: abca28511c9646c389bd13927898d8d40935afa6
* Version of the web browser and OS: `Chrome Version 57.0.2987.133 (64-bit) OS X 10.12`
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): 
`beta.stemcellcommons.org`

### Steps to reproduce
1. Search for `cdx` in the Dataset search bar:
![screen shot 2017-05-01 at 1 12 34 pm](https://cloud.githubusercontent.com/assets/5629547/25586983/ec6ef80c-2e6f-11e7-8161-cb440f4f836f.png)

2. Search for `cdx2`:
![screen shot 2017-05-01 at 1 12 51 pm](https://cloud.githubusercontent.com/assets/5629547/25586990/f63e7a2e-2e6f-11e7-9ee5-ada24700e778.png)

### Observed behavior
* Searching for `cdx2` yields three more results than `cdx` even though `cdx` is a substring.

### Expected behavior
* `cdx2` results should probably be shown when searching for `cdx`
* The current observance could be the expected behavior if the search relies on a lazy matching approach?

flekschas:
As far as I remember there is no prefix search across the data set description as the description can potentially be very long and prefix search can be slow (and resource intensive). The extra search results for `cdx2` are coming from exact matches in the description (that's why searching for `cdx` does not highlight `cdx2` in the data set description).
You can change that behavior by adjusting the way the data set description in stored in Solr. Title is a Edge n-gram field while the description is just a text field if I remember correctly.

scottx611x:
@flekschas Awesome, thanks for the quick response 🎊

scottx611x:
Related to #1702

scottx611x:
Addressed by #1702

Issue #1702, opened 2017-05-01
Revise indexing of data sets in Refinery


sjhosui:
Index on the following fields:
- [x] Study Person Last Name (i.e. PI)
    - The information has always been indexed but the last name could not be retrieved properly. For example, one could search for `Ramesh` but not `Shivdasani` before PR #1716 
- [x] Study Assay Measurement Type (i.e. technology type)
- [x] Study Identifier 
- [x] External ID (GSE)
- [x] index Description as an `EdgeNGram` field and test performance
 - [x] Are we indexing on both the title and description? **YES, see: https://github.com/refinery-platform/refinery-platform/blob/develop/refinery/core/search_indexes.py#L22-L37**

- [x] Look into this error:
`2017-05-03 15:24:56 ERROR    core.utils:54 update_data_set_index() - Could not update DataSetIndex: 'module' object has no attribute 'search_indexes'`

scottx611x:
@sjhosui @ngehlenborg 

I'm not seeing the `Study Person Last Name (i.e. PI)` information within a DataSet of any of its associated models. Is this something contained in the ISATAB standard that we aren't pulling out?

It's also not clear to me what a `Study Identifier` or `External ID` are in this context.

I think the real issue is that indexing may just be broken ( RE: `2017-05-03 15:24:56 ERROR core.utils:54 update_data_set_index() - Could not update DataSetIndex: 'module' object has no attribute 'search_indexes'`)

If dataset indexing has been broken for awhile this could explain many of the issues we've been seeing w/ search.

scottx611x:
Addressed by #1716

PR #1716, opened 2017-05-04
Scottx611x/indexing updates

scottx611x:
- Change `description` to an `EdgeNgramField` for better Dataset Description search by tokenizing on whitespace
- Fix `'module' object has no attribute 'search_indexes'` error to allow DataSet indexing to work again
- Add description to solr query fields
- Addresses: #1702
- Addresses: #1703 

Once merged a `./manage.py rebuild_index --using=core --batch-size=25` will need to be run

### DIFF
--- a/refinery/core/search_indexes.py
+++ b/refinery/core/search_indexes.py
@@ -76,6 +76,8 @@ class DataSetIndex(indexes.SearchIndex, indexes.Indexable):
                     "{}, {}".format(contact.last_name, contact.first_name)
                 )
 
+        # Cast to `list` looks redundant, but MultiValueField stores sets
+        # improperly, introducing a search bug. See: http://bit.ly/2pZLE5c
         return list(set(submitters))
 
     def prepare_measurement(self, object):
@@ -91,6 +93,8 @@ class DataSetIndex(indexes.SearchIndex, indexes.Indexable):
             for assay in study.assay_set.all():
                 measurements.append(assay.measurement)
 
+        # Cast to `list` looks redundant, but MultiValueField stores sets
+        # improperly, introducing a search bug. See: http://bit.ly/2pZLE5c
         return list(set(measurements))
 
     def prepare_technology(self, object):
@@ -106,6 +110,8 @@ class DataSetIndex(indexes.SearchIndex, indexes.Indexable):
             for assay in study.assay_set.all():
                 technologies.append(assay.technology)
 
+        # Cast to `list` looks redundant, but MultiValueField stores sets
+        # improperly, introducing a search bug. See: http://bit.ly/2pZLE5c
         return list(set(technologies))
 
     # from:
--- a/refinery/core/search_indexes.py
+++ b/refinery/core/search_indexes.py
@@ -65,13 +65,16 @@ class DataSetIndex(indexes.SearchIndex, indexes.Indexable):
         submitters = []
 
         for contact in investigation.contact_set.all():
-            submitters.append(contact.last_name + ", " + contact.first_name)
+            submitters.append(
+                "{}, {}".format(contact.last_name, contact.first_name)
+            )
 
         studies = investigation.study_set.all()
         for study in studies:
             for contact in study.contact_set.all():
                 submitters.append(
-                    contact.last_name + ", " + contact.first_name)
+                    "{}, {}".format(contact.last_name, contact.first_name)
+                )
 
         return set(submitters)
 
--- a/refinery/core/search_indexes.py
+++ b/refinery/core/search_indexes.py
@@ -26,7 +26,7 @@ class DataSetIndex(indexes.SearchIndex, indexes.Indexable):
     dbid = indexes.IntegerField(model_attr='id')
     uuid = indexes.CharField(model_attr='uuid')
     summary = indexes.CharField(model_attr='summary', null=True)
-    description = indexes.CharField(null=True)
+    description = indexes.EdgeNgramField(null=True)
     creation_date = indexes.DateTimeField(model_attr='creation_date')
     modification_date = indexes.DateTimeField(model_attr='modification_date')
     submitter = indexes.MultiValueField(null=True)
--- a/refinery/core/search_indexes.py
+++ b/refinery/core/search_indexes.py
@@ -76,7 +76,7 @@ class DataSetIndex(indexes.SearchIndex, indexes.Indexable):
                     "{}, {}".format(contact.last_name, contact.first_name)
                 )
 
-        return set(submitters)
+        return list(set(submitters))
 
     def prepare_measurement(self, object):
         investigation = object.get_investigation()
@@ -91,7 +91,7 @@ class DataSetIndex(indexes.SearchIndex, indexes.Indexable):
             for assay in study.assay_set.all():
                 measurements.append(assay.measurement)
 
-        return set(measurements)
+        return list(set(measurements))
 
     def prepare_technology(self, object):
         investigation = object.get_investigation()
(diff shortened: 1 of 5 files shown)


refinery/core/utils.py
```diff
@@ -21,6 +21,9 @@
 import core
 import data_set_manager
 
+from core.search_indexes import DataSetIndex
```
hackdna: Use explicit relative imports

refinery/core/utils.py
```diff
@@ -21,6 +21,9 @@
...
+from data_set_manager.search_indexes import NodeIndex
```
hackdna: `data_set_manager` is already imported above

hackdna:
Also, module imports should be fixed according to the style guide.

refinery/core/utils.py
scottx611x: @hackdna I would love to, but neither `core` nor `data_set_manager` have a reference to `search_indexes`

refinery/core/utils.py
scottx611x: @hackdna For example we were referencing the indexes before according to our style guide, but things were broken:
```
(refinery-platform)vagrant@refinery:/vagrant/refinery$ python
Python 2.7.6 (default, Oct 26 2016, 20:30:19)
[GCC 4.8.4] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import core
>>> core.search_indexes.DataSetindex
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'search_indexes'
>>> import data_set_manager
>>> data_set_manager.search_indexes.NodeIndex
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'search_indexes'
```

scottx611x: `core.utils` is also the file that introduces a nifty spiderweb of imports. 

There is not much we can do atm without some major refactoring/tackling of technical debt

hackdna: OK, this is due to mutual imports. Could you add a comment? Also, could you reference this in the circular imports Trello card?

refinery/ui/source/js/commons/data-sets/search-api.spec.js
```diff
@@ -68,7 +68,7 @@ describe('DataSet.search-api: unit tests', function () {
       'hl.simple.post': '%3C%2Fem%3E',
       'hl.simple.pre': '%3Cem%3E',
       q: _query,
-      qf: 'title%5E0.5+accession+submitter+text',
+      qf: 'title%5E0.5+accession+submitter+text+description',
```
flekschas: You now search twice on the description because the description is part of the `text` field. The `text` field is composed of multiple other fields and make up the core part of the document to be search for in Solr. Check out `templates/searches/indexes/core/dataset_text.txt`

flekschas: There are two possible solutions:
1. do not search on description explicitly
2. remove the description from the `text` field. On the other hand the `text` field should contain all the content of a document to be searched for.

flekschas: I also forgot to mention that using EdgeNGram increases the memory footprint of Solr quite a but because the description will be shredded into little pieces of overlapping n-grams which will all be stored explicitly as far as I remember.

Here's a useful post explaining a bit of what's going on: http://lucene.472066.n3.nabble.com/Solr-Wildcard-Search-for-large-amount-of-text-td4214392.html

scottx611x: @flekschas : I was finding that not all DataSet description information was included in this text field. 

For example: 
`With "EdgeNGramming"`:
screen shot 2017-05-04 at 2 43 18 pm

and Normally:
screen shot 2017-05-04 at 2 38 26 pm

This approach is more resource intensive, but it doesn't seem too bad at first glance.
Using `117` DataSets:

EdgeNGram:
```
{
  "solr_home":"/vagrant/refinery/solr/",
  "version":"5.3.1 1703449 - noble - 2015-09-17 01:48:15",
  "startTime":"2017-05-04T18:54:01.44Z",
  "uptime":"0 days, 0 hours, 0 minutes, 46 seconds",
  "memory":"76.3 MB (%15.6) of 490.7 MB"}
```
Normal:
```
{
  "solr_home":"/vagrant/refinery/solr/",
  "version":"5.3.1 1703449 - noble - 2015-09-17 01:48:15",
  "startTime":"2017-05-04T18:43:10.319Z",
  "uptime":"0 days, 0 hours, 6 minutes, 0 seconds",
  "memory":"74.5 MB (%15.2) of 490.7 MB"}
```

I'm probably not educated enough on this topic to see any repurcussions? 

@ngehlenborg: Would we be okay with taking this resource hit for better search results?

flekschas: @scottx611x You are finding different data sets because of the switch to EdgeNGram not because the information wasn't there. The memory foot print is expected to be the same (there is an overhead on indexing) but the size of the indices might increase. Using EdgeNGram is fine but now you search twice on `description` (you search on the description field alone and in combination with the text field). This might not have a huge impact but could slow things down in the future or artificially boost hits in the description.

I would also test few more search queries to see if side effects pop up (I ran into this problem after I added synonym search for example, but only for some queries). E.g.:
- ES
- iPS
- cel
- RNA
- ...

scottx611x: For additional context, the search index increased from `1.23Mb` -> `1.31Mb` after indexing with the DataSet description as an `EdgeNGram` field.

refinery/core/search_indexes.py
```diff
@@ -65,15 +65,18 @@ def prepare_submitter(self, object):
...
         studies = investigation.study_set.all()
         for study in studies:
             for contact in study.contact_set.all():
                 submitters.append(
-                    contact.last_name + ", " + contact.first_name)
+                    "{}, {}".format(contact.last_name, contact.first_name)
+                )
 
-        return set(submitters)
+        return list(set(submitters))
```
hackdna: Appears to be redundant since `submitters` is already a list

refinery/core/search_indexes.py
```diff
@@ -65,15 +65,18 @@ def prepare_submitter(self, object):
         submitters = []
 
         for contact in investigation.contact_set.all():
-            submitters.append(contact.last_name + ", " + contact.first_name)
+            submitters.append(
+                "{}, {}".format(contact.last_name, contact.first_name)
```
hackdna: In simple cases like these `format` is probably an overkill

refinery/core/search_indexes.py
scottx611x: This is not redundant. We cast to `set` to get a unique list of `submitters` and back to `list` to fix a search bug. 

Solr kept the `set` in its index like so: `submitters: "set(u'Armstrong, Scott')"`, but
breaks this field on whitespaces.
Due to some prefix searching going on, one could search for `Scott` and `set(u'Armstrong` but not for `Armstrong` alone

refinery/core/search_indexes.py
scottx611x: Why not keep in line with what our coding style guide states?

hackdna: The coding style guide only says to replace the old `%`-based syntax with the new `format`-based syntax. It doesn't actually say anything about replacing a simple string concatenation with `format`. In this case it comes down to readability vs separation of style from data. I'll let you decide.

refinery/core/search_indexes.py
hackdna: OK, could you add a short comment about this (to all three locations)?

refinery/core/search_indexes.py
scottx611x: I only saw this in our style guide regarding python string formatting: 
screen shot 2017-05-08 at 4 56 26 pm

PR #1723, opened 2017-05-10
Scottx611x/migration bug fix

scottx611x:
- "Add a new migration that adds the whole updated initial data set, then go back and comment out the code in the previous initial data migration since there's no point running it twice on new database setup" from https://www.caktusgroup.com/blog/2015/11/10/initial-data-django/

- Provide mechanism to allow for migrations to be properly applied when bringing up an AWS stack from existing backups regardless if prior migrations are `faked` or not


refinery/file_store/migrations/0002_filetypes_and_fileextensions.py
```diff
@@ -3,316 +3,11 @@
...
-
-    for file_extension in file_extensions:
-        # Attempt to save FileExtension objects. If an IntegrityError is raised
-        # upon saving an object we continue as it already exists.
-        try:
-            with transaction.atomic():
-                file_extension.save()
-        except IntegrityError:
-            continue
+    pass
```
hackdna: Since this is an empty migration now, I'd simply remove [unnecessary code](https://docs.djangoproject.com/en/1.7/topics/migrations/#data-migrations)

refinery/file_store/migrations/0003_filetypes_and_fileextensions.py
```diff
@@ -0,0 +1,378 @@
+# encoding: utf8
+
+from django.db import migrations, connection
+from django.db.models import Max
+
+def forwards(apps, schema_editor):
+    Filetype = apps.get_model("file_store", "Filetype")
+    FileExtension = apps.get_model("file_store", "FileExtension")
+
+    if Filetype.objects.count():
```
hackdna: Why is this check necessary?

refinery/file_store/migrations/0003_filetypes_and_fileextensions.py
```diff
@@ -0,0 +1,378 @@
+# encoding: utf8
+
+from django.db import migrations, connection
+from django.db.models import Max
+
+def forwards(apps, schema_editor):
+    Filetype = apps.get_model("file_store", "Filetype")
```
hackdna: The actual model name is FileType, so should probably be used for consistency.

refinery/file_store/migrations/0003_filetypes_and_fileextensions.py
```diff
@@ -0,0 +1,378 @@
...
+    FileExtension = apps.get_model("file_store", "FileExtension")
+
+    if Filetype.objects.count():
+        # Workaround for old-fixture based FileType data messing up
+        # Postgres' id sequence. See article here: http://bit.ly/2qsevm2 for similar issue
+        with connection.cursor() as cursor:
+            highest_id = Filetype.objects.all().aggregate(Max('id'))["id__max"]
```
hackdna: This should be outside the with block

refinery/file_store/migrations/0003_filetypes_and_fileextensions.py
```diff
@@ -0,0 +1,378 @@
...
+        Filetype(
+            description="BEDGRAPH file",
+            name="BEDGRAPH",
+            used_for_visualization=True
+        )
+    ]
+
+    existing_filetypes = [filetype.name for filetype in Filetype.objects.all()]
+
+    for filetype in filetypes:
```
hackdna: Why not simply use [bulk_create](https://docs.djangoproject.com/en/1.7/ref/models/querysets/#bulk-create) and pass on IntegrityError to avoid duplicates?

refinery/file_store/migrations/0003_filetypes_and_fileextensions.py
scottx611x: This issue doesn't affect "clean" instances without any prior FileType/Extension information, so theres no need to reset the postgres id sequences

refinery/file_store/migrations/0003_filetypes_and_fileextensions.py
scottx611x: One reason would be that we would lose the ability to do these manual:
` if filetype.name in existing_filetypes` checks. 

When migrations had been faked prior, the `unique` constraints on newly created Filetypes/FileExtensions fields were not enforced, thats why this check is necessary.

PR #1731, opened 2017-05-15
Jkmarx/import text update

### LINKED ISSUE #1729: Changes to Upload page text
- [x] Change "arbitrary" to "arbitrarily"
arbitrarycolumnorder

- [x] Alphabetize reserved words
alphareservedwords

- [x] Change "Upload Local Data Files" popover text to: Select **Yes** if the **Data File Column** contains paths to local files which now need to be uploaded. Select **No** if the **Data File Column** contains URLs of remote files already uploaded elsewhere.
uploadlocaldatafiles

jkmarx:
Resolves #1729

### DIFF
--- a/refinery/ui/source/js/data-set-import/partials/metadata-table.html
+++ b/refinery/ui/source/js/data-set-import/partials/metadata-table.html
@@ -36,9 +36,9 @@
           <li>
             Can not have column headers named the following reserved words:
             <br>
-            <em>assay_uuid, id, django_id, file_uuid, study_uuid, ,
-            type, is_annotation, species, genome_build, name,
-            or django_ct</em>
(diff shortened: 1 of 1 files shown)


refinery/ui/source/js/data-set-import/partials/metadata-table.html
```diff
@@ -31,14 +31,14 @@
             Can have additional columns used to characterize data
           </li>
           <li>
-            Can have the columns ordered arbitrary
+            Can have the columns ordered arbitrarily
```
hackdna: According to our copy style guide we should use active voice: "Can have arbitrary column order".

jkmarx: "Can have the columns ordered arbitrarily" is clearer, so would leave the change as is.

PR #1733, opened 2017-05-15
Jkmarx/generate launch configs

jkmarx:
- Generates launch configs
- Rename some services for clarity
- Update back end to accept '[uuid]' format vs '['uuid']'
- Unit test


refinery/tool_manager/tests.py
```diff
@@ -960,7 +961,7 @@ def test_node_uuids_get_populated_with_urls(self):
 
         post_data = {
             "tool_definition_uuid": td.uuid,
-            "file_relationships": "['{}', '{}']".format(
+            "file_relationships": "[{}, {}]".format(
```
hackdna: Should probably be on one line if it fits.

refinery/tool_manager/utils.py
```diff
@@ -157,6 +157,21 @@ def create_tool(tool_launch_configuration, user_instance):
...
+    try:
+        nesting = ast.literal_eval(tool.file_relationships)
+    except (SyntaxError, ValueError):
+        raise RuntimeError(
+            "ToolLaunchConfiguration's `file_relationships` could not be "
+            "evaluated as a Pythonic Data Structure: {}".format(
+                tool.file_relationships
+            )
+        )
+    else:
```
hackdna: Else is not needed because exception is (re-)raised.

Issue #1649, opened 2017-03-27
UI - Grid Selection Display


jkmarx:
- [x] For UI-Grid node selection, add event to trigger pop-over/drop down
    - [x] Component, Ctrl
    - [x] Unit Test

- [x] Selection template will show workflow form/template from selection service
    - [x] CSS to make it clear to user what is required

- [x] Add Nav Controls  (Group #, Left, Right, Add, Remove)
    - [x] Component, Ctrl
    - [x] Unit Test
    - [x] Sync up controls to Control Panel and Cart
    - [x] Drop down only shows empty groups or groups associated with row

- [x] Update selection service will maintain form updates per group
    - [x] Sync up service with ctrl panel and cart

- [x] Maintain row selection info upon grid reset, filtering, etc.

- [x] Add indicator column 
    - [x] Add template
    - [x] Display color
    - [x] Darken arrow if popover selected or row is selected
    - [x] Remove/remove all clears styling
    ~- [ ]  Display first 5 groups from Node Select Collection and then ....~
    - [x] Adjust css to show ... for overflow
    - [x]  Ensure display during cont scrolling and filtering (any resets)

- [x] Hide Selection column unless tool is selected

~- [ ] Handle batch selection (hide when not possible)~

- [x] Add parameters

- [x] Clean up File Browser Ctrl
    - [x] Remove unused or unnecessary functionalities

- [x] Track collapse service

- [x] For lists: Remove input details

- [x] For lists: Remove display in popover and show Item #? in Group #?

PR #1737, opened 2017-05-17
Jkmarx/data set 2 add selection nav

jkmarx:
- Add navigation to popover Ref #1649 
- Add validation, inner group must be filled before moving to a new group (pairs require 2 nodes, lists require 1 node)
- Add unit tests

nav-with-validation


refinery/ui/source/js/tool-launch/partials/input-control-inner-nav.html
```diff
@@ -0,0 +1,27 @@
+ <div>
+  <button
+    class="btn btn-default btn-xs outline"
+    ng-disabled="$ctrl.currentGroup[$ctrl.currentGroup.length-2] == 0"
```
hackdna: 2 looks like a magic number

jkmarx: This is fixed in my current branch which addresses who currentGroup is being used.

PR #1738, opened 2017-05-19
update README

mccalluc:
These tweaks were necessary for me, but even if they belong, perhaps they should be in Troubleshooting?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ The above step should take about 15 minutes depending on the speed of your machi
 $ vagrant provision
 ```
 
+(If you are updating an older checkout, `refinery/config/config.json` will not be regenerated automatically: Either clear it, or add the missing keys from the error message.)
+
 Open <http://192.168.50.50:8000/> in your web browser.
 
 ### Configure Deployment Environment on the Host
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,6 +23,8 @@ The above step should take about 15 minutes depending on the speed of your machi
 $ vagrant provision
 ```
 
+(If you are updating an older checkout, `refinery/config/config.json` will not be regenerated automatically: Either clear it, or add the missing keys from the error message.)
+
```
hackdna: This info is [already available](https://github.com/refinery-platform/refinery-platform/wiki/Troubleshooting#improperly-configured-configjson)

README.md
```diff
@@ -42,6 +44,12 @@ Use `fabricrc.sample` to update or initialize Fabric configuration, for example:
 $ cp fabricrc.sample ~/.fabricrc
 ```
 
+You may need to start ssh-agent:
+``` bash
+$ eval `ssh-agent`
+$ ssh-add ~/.ssh/id_rsa
+```
+
```
hackdna: This should be added to the troubleshooting section of the troubleshooting wiki page.

mccalluc:
Added a note about ssh-agent to the wiki, and will close.

Issue #1734, opened 2017-05-15
Handle AttributeError in DataSet REST API


hackdna:
* Specific code commit: 7367df5
* Version of the web browser and OS: N/A
* Environment where the error occurred: AWS instance

### Steps to reproduce
Unknown: traceback received by email, possibly as a result of a data set import

### Observed behavior
```
Internal Server Error: /api/v1/data_sets/
Traceback (most recent call last):
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/tastypie/resources.py", line 202, in wrapper
    response = callback(request, *args, **kwargs)
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/tastypie/resources.py", line 433, in dispatch_list
    return self.dispatch('list', request, **kwargs)
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/tastypie/resources.py", line 465, in dispatch
    response = method(request, **kwargs)
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/tastypie/resources.py", line 1298, in get_list
    bundles.append(self.full_dehydrate(bundle, for_list=True))
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/tastypie/resources.py", line 861, in full_dehydrate
    bundle = self.dehydrate(bundle)
  File "/srv/refinery-platform/refinery/core/api.py", line 541, in dehydrate
    bundle.data["version"] = bundle.obj.get_version_details().version
AttributeError: 'NoneType' object has no attribute 'version'

Request repr():
<WSGIRequest
path:/api/v1/data_sets/,
GET:<QueryDict: {u'order_by': [u'-modification_date'], u'offset': [u'50'], u'limit': [u'50'], u'format': [u'json']}>,
POST:<QueryDict: {}>,
COOKIES:{'_ga': 'GA1.2.2126546047.1492614465',
'_gat': '1',
'_gid': 'GA1.2.844249910.1494882001',
'csrftoken': <redacted>,
'sessionid': <redacted>},
META:{'CONTEXT_DOCUMENT_ROOT': '/var/www',
'CONTEXT_PREFIX': '',
u'CSRF_COOKIE': <redacted>,
'DOCUMENT_ROOT': '/var/www',
'GATEWAY_INTERFACE': 'CGI/1.1',
[...]

PR #1739, opened 2017-05-19
Scottx611x/dataset api error handling

scottx611x:
Addresses #1734


scottx611x:
@hackdna I've only been able to reproduce this issue by hitting `/api/v1/datasets` in the middle of a tabular dataset upload, or by manually creating a minimal dataset in the shell: just with `DataSet.objects.create()`

See here for where a Dataset exists for a time without `InvestigationLinks` etc.
https://github.com/refinery-platform/refinery-platform/blob/develop/refinery/data_set_manager/tasks.py#L371-L385

hackdna:
OK, in other words, this happens when a data set is not ready to be used because it has not been fully created, so we should probably return HTTP 404.

scottx611x:
@hackdna Why should we return a 404 for `/api/v1/datasets` response if one DataSet isn't fully created? I think this would be very confusing for users

scottx611x:
@hackdna I'm just saying that we shouldn't disallow users from browsing the datasets api and return a 404 just because one dataset hasn't been fully created. 

I think the idea: `Until a data set is fully created it should not be returned.` is sufficient, but a 404 would be odd.

hackdna:
Why would returning a 404 for one data set prevent users from browsing other data sets?

scottx611x:
@hackdna the tastypie endpoint that we hit on the Main page of Refinery is: `/api/v1/data_sets/`

When we do this we grab all of the current Datasets due to how the `DataSetResource`'s `queryset` attribute is set, and the `dehydrate` method is called upon each DataSet instance returned from the queryset. 

With this error we are seeing, the actual `DataSet` object has already been created, and only its related objects haven't been created yet. Since the Tastypie resource doesn't care about these related objects (until we add them in `dehydrate()`) it still includes the dataset in it's response. 

I've provided `AttributeError` handling in the `dehydrate` method, which is where we are adding custom fields to each DataSet object. Here is where I'm assuming you'd want to return a 404? 

Returning a 404 in the middle of the dehydration cycle would return the actual `HttpResponse` object which gets serialized and included in the api response causing the Dashboard to react oddly.

See: 
screen shot 2017-05-24 at 11 51 14 am
screen shot 2017-05-24 at 11 51 01 am

Providing a 404 response at a higher level is possible, but then no datasets would be returned from this endpoint if we are in the middle of creating a single one.

hackdna:
Ideally:
* for a single data set API request: return a 404 if the data set has not been initialized completely yet (with an appropriate logger message)
* for a data set list API request: return a list of data sets excluding the ones that have not been initialized completely yet

It looks like you'll need to use a [custom ListSerializer](http://www.django-rest-framework.org/api-guide/serializers/#customizing-listserializer-behavior) to implement this.

scottx611x:
@hackdna This isn't a DRF endpoint, so I'm not sure if there is a great way to do this in tastypie, but I can look.

Regardless, the original issue here was only to handle the uncaught attribute error here. Considering this is getting to be more involved, should there be a separate issue?

hackdna:
I wouldn't close this because I don't think that the current error handling approach is the correct one and this branch should not be merged into develop without at least estimating the effort to implement the correct approach.

refinery/core/api.py
```diff
@@ -498,6 +498,13 @@ class Meta:
         }
 
     def dehydrate(self, bundle):
+        version_details = bundle.obj.get_version_details()
```
hackdna: `get_version_details()` is not named correctly because it actually returns an InvestigationLink instance, so I would rename the variable holding the return value to something like `investigation_link`

refinery/core/api.py
```diff
@@ -498,6 +498,13 @@ class Meta:
...
+
+        if version_details is None:
```
hackdna: Is this actually necessary if you already have added checks to `obj_get()` and `obj_get_list()`?

scottx611x: I don't think its _super_ necessary, but it does cover the edge-case where an importing DataSet is fully created inbetween the calls to `dehydrate` and `obj_get`/`obj_get_list`

This is less of an edge case when we are fetching many hundreds of DataSets

refinery/core/api.py
```diff
@@ -831,6 +843,54 @@ def get_study_assays(self, request, **kwargs):
...
+        """
+        Handle incomplete DataSet instances when a list view is accessed.
+        :param datasets: a list of DataSet objects
+        :return: a list of DataSet objects sans ones that don't have the
+        proper information associated yet.
+        """
+        complete_datasets = []
+
+        for dataset in datasets:
+            if dataset.get_version_details() is None:
```
scottx611x: I thought that it was okay to split these two up since they really are providing different functions:

`handle_incomplete_dataset` happens during a detail view, takes a single dataset and returns a 404 if its not fully created.

`handle_incomplete_datasets` happens during a list view, takes a list of datasets and excludes datasets from its response if they are not fully created, logging the errors rather than providing a 404 for the entire response.

refinery/core/api.py
hackdna: [Tastypie docs](http://django-tastypie.readthedocs.io/en/latest/resources.html#flow-through-the-request-response-cycle) say that serialization happens after objects are retrieved, so it appears to be impossible to have an incomplete data set at this point. Even if that's somehow not the case, returning an incomplete dataset is not the answer.

refinery/core/api.py
hackdna: Actually, I would go further and re-use `handle_incomplete_dataset()` in `get_obj_list()` directly:
```python
    for dataset in datasets:
        try:
            complete_datasets.append(handle_incomplete_dataset(dataset))
        except NotFound:
            pass
```
Then move the functionality of `make_incomplete_dataset_error_message()` into `handle_incomplete_dataset()`.
Then remove both `make_incomplete_dataset_error_message()` and `handle_incomplete_datasets()` completely.

refinery/core/api.py
scottx611x: Its very possible to have an "incomplete" DataSet at that point:

`dehydrate()` is hit before `obj_get()` and `obj_get_list()`
As far as TastyPie cares, its queryset includes: `DataSet.objects.all()` so "Incomplete" DataSets are always included in the response.

I only found one example in our codebase so far, but I'm sure we are doing things like:
```
ds = DataSet.objects.create(name="DataSet")
####
# Some code happens here and a user hits `/api/v1/datasets` during this time
###

# Other objects are then associated 
i = Investigation.objects.create()
s = Study.objects.create(investigation=i)
i_l = InvestigationLink.objects.create(
        data_set=dataset,
        investigation=i,
        version=1,
        date=datetime.now()
    )
ds.save() # DataSet is "complete"
```

refinery/core/api.py
hackdna: Probably even better to add a validation method to the `DataSet` class that would return a boolean. Then you can simply use that in `get_obj()` and `get_object_list()` instead of `handle_incomplete_dataset()`. It could also be easily re-used elsewhere.

refinery/core/api.py
```diff
@@ -678,24 +671,34 @@ def filter_by_group(self, request, obj_list):
         return obj_list
 
     def obj_get(self, bundle, **kwargs):
-        dataset = SharableResourceAPIInterface.obj_get(self, bundle, **kwargs)
-        return self.handle_incomplete_dataset(dataset)
+        try:
+            dataset = DataSet.objects.get(uuid=kwargs["uuid"])
```
hackdna: I'd replace this whole block with:
```python
dataset = super(ModelResource, self).obj_get(bundle, **kwargs)
```
and no exception handling necessary since this will raise `NotFound` if there are any problems

refinery/core/api.py
```diff
@@ -678,24 +671,34 @@ def filter_by_group(self, request, obj_list):
...
+            logger.error("Couldn't properly fetch DataSet with UUID: %s %s",
+                         kwargs["uuid"], e)
+            raise NotFound(e)
+
+        if not dataset.is_valid():
+            raise NotFound(
+                "DataSet with UUID: {} is invalid, and most likely is "
+                "still being created".format(dataset.uuid)
+            )
+        return SharableResourceAPIInterface.obj_get(self, bundle, **kwargs)
```
hackdna: and technically, this should be:
```python
return super(SharableResourceAPIInterface, self).obj_get(bundle, **kwargs)
```

refinery/core/api.py
```diff
@@ -678,24 +671,34 @@ def filter_by_group(self, request, obj_list):
...
 
     def obj_get_list(self, bundle, **kwargs):
-        datasets = SharableResourceAPIInterface.obj_get_list(
-            self,
-            bundle,
-            **kwargs
-        )
-        complete_datasets = []
+        datasets = SharableResourceAPIInterface.obj_get_list(self, bundle,
```
hackdna: probably need to call `get_obj_list()` of `ModelResource` and remove invalid data sets from the bundle then
```python
datasets = super(SharableResourceAPIInterface, self).obj_get_list(bundle, **kwargs)
```

refinery/core/api.py
```diff
@@ -678,24 +671,34 @@ def filter_by_group(self, request, obj_list):
...
-                    self.handle_incomplete_dataset(dataset)
+            if dataset.is_valid():
+                valid_datasets.append(dataset)
+            else:
+                logger.error(
+                    "DataSet with UUID: {} is invalid, and most likely is "
+                    "still being created".format(dataset.uuid)
                 )
-            except NotFound:
                 pass
```
hackdna: unnecessary

refinery/core/api.py
scottx611x: Making this proposed change yields: 
`{AttributeError}'super' object has no attribute 'obj_get'`

refinery/core/api.py
scottx611x: I think we are avoiding the super call to ModelResource because of the wackyness mentioned in ShareableResourceApiInterfaces `obj_get()`

Regardless, calling `obj_get()` on an invalid dataset bundle is what leads to the `AttributeErrors` we've seen in the dehydrate method in the first place

refinery/core/api.py
scottx611x: Making this proposed change yields:
`AttributeError: 'super' object has no attribute 'obj_get_list'`

refinery/core/api.py
scottx611x: I've since removed this `try/except`

refinery/core/api.py
hackdna: of course, it should be:
```python
return super(DataSetResource, self).obj_get(bundle, **kwargs)
```
also, `SharableResourceAPIInterface` should be the first parent in `DataSetResource` class definition to enable the correct method resolution order

refinery/core/api.py
hackdna: right, it actually should be:
```python
dataset = super(DataSetResource, self).obj_get(bundle, **kwargs)
```
with `SharableResourceAPIInterface` as the first parent in `DataSetResource` class definition to enable the correct method resolution order

refinery/core/api.py
```diff
@@ -282,6 +282,12 @@ def obj_get(self, bundle, **kwargs):
         if not user.is_authenticated() and res and not res.is_public():
             return HttpUnauthorized()
 
+        if not res.is_valid():
+            raise NotFound(
+                "DataSet with UUID: {} is invalid, and most likely is "
```
hackdna: This is `SharableResourceAPIInterface` class. Why is there a reference to DataSet here?

refinery/core/models.py
```diff
@@ -654,37 +654,37 @@ def get_version(self):
...
+        """
+        Try to fetch the latest InvestigationLink related to the DataSet
+        instance. If a version is provided, try to fetch the latest
+        InvestigationLink for said version.
+        :param version: integer
+        :returns: an InvestigationLink Instance or None if Exception occurs
+        """
         try:
             if version is None:
                 version = (
```
hackdna: Unnecessary parentheses

refinery/core/models.py
```diff
@@ -654,37 +654,37 @@ def get_version(self):
...
                 )
 
-            return (
-                InvestigationLink.objects.filter(
-                    data_set=self, version=version).get()
-            )
+            return InvestigationLink.objects.filter(
+                data_set=self,
+                version=version
+            ).get()
```
hackdna: Can we simply do:
```python
return InvestigationLink.objects.get(data_set=self, version=version)
```
?

refinery/core/api.py
scottx611x: The `super` calls (or regular calls we were making prior) being made to`obj_get` with a uuid of an invalid DataSet are followed directly by a call to `dehydrate()`. This was the root cause of the `AttributeErrors` we were seeing originally. 

The logic to handle this has to be in `DataSetResource.dehydrate()`, `DataSetResource.obj_get()`, or `SharableResourceAPIInterface.obj_get()` AFAIK.

I introduced logic to handle for these invalid DataSets which I provided in the `deyhdrate()` method originally like so:
``` python
class DataSetResource(ModelResource, SharableResourceAPIInterface):
...
    def dehydrate(self, bundle):
        if not bundle.obj.is_valid():
            # Dataset has not been associated with its InvestigationLink so
            # we are not able to provide a fully created bundle. Returning
            # the bundle in this state will properly exclude it from the api
            # response
            return bundle
    ...
```
Then I was asked to move it out of `dehydrate()`, so I moved it into `obj_get()` like so:
``` python
class DataSetResource(ModelResource, SharableResourceAPIInterface):
...
    def obj_get(self, bundle, **kwargs):
            try:
                dataset = DataSet.objects.get(uuid=kwargs["uuid"])
            except (DataSet.DoesNotExist, DataSet.MultipleObjectsReturned) as e:
                logger.error("Couldn't properly fetch DataSet with UUID: %s %s",
                             kwargs["uuid"], e)
                raise NotFound(e)

            if not dataset.is_valid():
                raise NotFound(
                    "DataSet with UUID: {} is invalid, and most likely is "
                    "still being created".format(dataset.uuid)
                )
            return SharableResourceAPIInterface.obj_get(self, bundle, **kwargs)
```

Then, being asked to remove that code block forced me to place the logic in the parent class that your are seeing here.

PR #1744, opened 2017-05-24
Jkmarx/remove node group

### LINKED ISSUE #1643: Node Groups Repurpose
- ~[ ] Update UI: Node Group stores all filtered uuids (not selected nodes)~
- [x] Remove Node Groups API and associated unit tests
- [x] Remove Node Groups Service and other references in UI
- [x] Remove Current Selection in UI
- [x] Remove Current Selection in data base and api

jkmarx:
Resolves #1643 
- Removes node group services, references, helper methods from UI
- Removes node group API, model, and serializer


jkmarx:
@scottx611x Node Groups was only used in the data set two browser. It was used initially for launching analyses in the data set 2.

PR #1745, opened 2017-05-24
Jkmarx/add launch validation

jkmarx:
- Move needMoreNodes helper method to service so both group and launch tool could use it
- Update unit tests accordingly
launch-validation


refinery/ui/source/js/tool-launch/ctrls/input-control-inner-nav-ctrl.js
```diff
@@ -30,36 +32,10 @@
...
-          moreFlag = areInputFileTypesEmpty();
-        } else {
-          moreFlag = true;
-        }
-      } else {
-        // list
-        moreFlag = areInputFileTypesEmpty();
-      }
-      return moreFlag;
+      return toolLaunchService.checkNeedMoreNodes();
```
hackdna: I'd leave the function name the same. I think it's more natural to read/write `if needMoreNodes()`.

refinery/ui/source/js/tool-launch/services/tool-launch-service.js
```diff
@@ -150,6 +151,17 @@
       return finalSetStr;
     }
 
+    // helper method to check if each group-inputType are empty
+    function areInputFileTypesEmpty () {
```
hackdna: I'd call it `inputFileTypesEmpty()` for the same reason as the comment above.

Issue #1735, opened 2017-05-16
Handle Exceptions during `rebuild_index` mgmt. command


scottx611x:
* Specific code commit: 7367df586c57aae5d622800032de12458003073b
* Version of the web browser and OS: `OSX 10.12 Chrome Version 58.0.3029.110 (64-bit)`
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): 
Both `Vagrant` & `AWS - beta.stemcellcommons.org`

### Steps to reproduce
Please list all the actions and the input data used:

1.  Create A `DataSet` without an `Investigation` such that: `<DataSet>.get_investigation()` yields `None`
2. Create a DataSet with a `Contact` who has non ascii chars in their name: e.g. `ä`
> NOTE: The Beta site's backups include both of these scenarios
3. Run: `./manage.py rebuild_index --using=core --batch-size=25`

### Observed behavior
**RE: 1**
```
  File "/srv/refinery-platform/refinery/core/search_indexes.py", line 49, in prepare_description
    return object.get_investigation().get_description() or ""
AttributeError: 'NoneType' object has no attribute 'get_description'
```
**RE: 2**
```
  File "/srv/refinery-platform/refinery/core/search_indexes.py", line 76, in prepare_submitter
    "{}, {}".format(contact.last_name, contact.first_name)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 4: ordinal not in range(128)
```
### Expected behavior
* For the index to be able to be built regardless of the above scenarios

### General notes:
- I ran the `rebuild_index` command as a background process before, and due to the `batch-size` option, a large subset of the `DataSets` still were updated with the newly introduced indexing changes which led me to believe things were all ok.
- In the methods mentioned in the above Tracebacks, Exceptions should be handled in a way that allows solr's index to still be built.
- People probably won't always search for names using the proper chars Re: `ä` so it may be good to build the index with equivalent chars as well. (So we index: `"Dän Räymond"`, as well as: `"Dan Raymond"`)

scottx611x:
Regarding ascii/non-ascii chars:
https://cwiki.apache.org/confluence/display/solr/CharFilterFactories

PR #1746, opened 2017-05-25
Handle errors that have occured when rebuilding solr's index

scottx611x:
Addresses #1735


refinery/core/search_indexes.py
```diff
@@ -66,14 +74,20 @@ def prepare_submitter(self, object):
 
         for contact in investigation.contact_set.all():
             submitters.append(
-                "{}, {}".format(contact.last_name, contact.first_name)
+                unicode("{}, {}").format(
```
hackdna: We should avoid `unicode()` because it does not exist in Python 3. Either add `u''` prefix or import `unicode_literals`: http://python-future.org/compatible_idioms.html#unicode
Thoughts?

refinery/core/search_indexes.py
```diff
@@ -45,8 +45,16 @@ def index_queryset(self, using=None):
...
+    def prepare_description(self, object, description=""):
+        try:
+            description = object.get_investigation().get_description()
+        except AttributeError as e:
+            logger.error(
+                "Could not fetch Investigation for DataSet with UUID: %s %s",
+                object.uuid,
+                e
+            )
+        return description
```
hackdna: Adding unnecessary arguments to function signatures should be avoided. One simple solution is to put `return object.get_investigation().get_description()` in the try block and add `return ""` at the very end.

PR #1749, opened 2017-05-26
Mccalluc/all files empty ui start api

mccalluc:
Mocking up the UI, and starting on the supporting API for the list of all files...


refinery/templates/core/all_files.html
```diff
@@ -0,0 +1,251 @@
...
+  <div class="page-header">
+    <!-- DIFF: Title -->
+    <h1>All Files</h1>
+  </div>
+{% endblock %}
+
+{% block content %}
+  <!-- tab headers -->
+
+<div ng-controller="DataSetNavCtrl" ng-cloak>
```
jkmarx: Not Required: DataSetNavCtrl - required for the file tabbing, DataSetUiModeCtrl - required for analyses state

Create your own controller for the all-files-view

refinery/templates/core/all_files.html
```diff
@@ -0,0 +1,251 @@
...
+  <div ng-controller="DataSetUiModeCtrl">
+
+    <!-- DIFF: Only one tab, so remove -->
+
+    <div class="refinery-panel refinery-panel-content jquery-full-height
+    handle-overflow" id="data-set-view">
+      <!-- DIFF: Remove "if data_set" -->
+      <div ui-view>TODO</div>
```
jkmarx: ui-view? using a router? 
Inject your all-file component

refinery/templates/core/all_files.html
```diff
@@ -0,0 +1,251 @@
...
+      var REFINERY_USER_AUTHENTICATED = true;
+    {% else %}
+      var REFINERY_USER_AUTHENTICATED = false;
+    {% endif %}
+
+    var csrf_token = "{{ csrf_token }}";
+  </script>
+
+  <!-- Script for popup content in the tool status and analyses icon-->
+  <script type="text/ng-template" id="nodeselectionpopover.html">
```
jkmarx: remove

refinery/templates/core/all_files.html
```diff
@@ -0,0 +1,251 @@
...
+    <rp-node-selection-popover-detail>
+    </rp-node-selection-popover-detail>
+  </script>
+
+  <script type="text/javascript" src="{% static "js/refinery/solr/solr_utilities.js" %}"></script>
```
jkmarx: These libraries and scripts below need to be reviewed. I'll put it on my agenda before leaving.

mccalluc:
Most recent travis PR build failed in NPM rather than actually getting to the tests... waiting to see if the Push build has the same problem...

mccalluc:
I would like to get this reviewed, but will continue in work in #1759.

refinery/data_set_manager/utils.py
```diff
@@ -503,7 +505,38 @@ def _index_annotated_nodes(node_type, study_uuid, assay_uuid=None,
...
+     """
+
+    assay_uuids = []
+    datasets = get_objects_for_user(user, "core.read_dataset")
+    for dataset in datasets:
+        version_details = dataset.get_version_details()
+        study = Study.objects.get(
+            investigation=version_details.investigation
+        )
+        assay = Assay.objects.get(
```
mccalluc: fit on one line?

refinery/data_set_manager/utils.py
```diff
@@ -557,7 +599,9 @@ def generate_solr_params(params, assay_uuid):
         solr_params = ''.join([solr_params, split_facet_fields])
     else:
         # Missing facet_fields, it is generated from Attribute Order Model.
-        attributes_str = AttributeOrder.objects.filter(assay__uuid=assay_uuid)
+        attributes_str = AttributeOrder.objects.filter(
+            assay__uuid__in=assay_uuids  # TODO: Confirm this syntax
+        )
```
mccalluc: Scott writes:
> regarding the `Thing.objects.filter(name__in=name_list)` syntax there are performance considerations, but they aren't applicable to our use case:

refinery/templates/core/user_files.html
```diff
@@ -0,0 +1,237 @@
...
+    {% if REFINERY_REPOSITORY_MODE %}
+      var REFINERY_REPOSITORY_MODE = true;
+    {% else %}
+      var REFINERY_REPOSITORY_MODE = false;
+    {% endif %}
+
+    {% if user.is_authenticated %}
+      var REFINERY_USER_AUTHENTICATED = true;
+    {% else %}
+      var REFINERY_USER_AUTHENTICATED = false;
```
mccalluc: Filed #1765: I don't think it's a security risk, though it might be cruft.

refinery/ui/source/js/user-file-browser/directives/user-file-browser-filters.js
```diff
@@ -0,0 +1,127 @@
+(function () {
+  'use strict';
+  angular
+    .module('refineryUserFileBrowser')
+    .component('rpUserFileBrowserFilters', {
+      controller: 'UserFileBrowserCtrl',
```
mccalluc: **TODO**

refinery/ui/source/js/user-file-browser/views/user-file-browser.html
```diff
@@ -0,0 +1,18 @@
+<div class="row y-scrollable m-t-1-2 p-t-1">
+    <div class="col-lg-2 col-md-2 col-sm-3">
+      <div id="table-filter" style="height: initial;"><!-- TODO: Why is the default CSS the way it is? -->
+        <div id="filter-panel-body" style="height: initial; overflow-y: initial;">
+            <rp-user-file-browser-filters></rp-user-file-browser-filters>
+        </div>
+      </div>
+    </div>
+    <div class="col-lg-10 col-md-10 col-sm-9" id="main-area">
+      <div style="margin-right: 10px">
```
mccalluc: **TODO**

refinery/user_files_manager/tests.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class UserFilesAPITests(TestCase):
+    pass
+    # TODO: Fails, because solr isn't running.
+    # Should I mock it, or make this an integration test?
+
+    # def test_get(self):
+    #     response = self.client.get('/api/v2/user/files/')
+    #     self.assertEqual(response.status_code, 200)
```
mccalluc: **TODO**

refinery/user_files_manager/views.py
```diff
@@ -0,0 +1,21 @@
...
+from django.http import HttpResponseNotAllowed
+from rest_framework.response import Response
+from data_set_manager.utils import (
+    search_solr, format_solr_response,
+    generate_solr_params_for_user)
+
+logger = logging.getLogger(__name__)
+
+
+def user_files(request):
```
mccalluc: **TODO**

mccalluc: **TODO**

mccalluc:
(Just pushed the easier parts and will address the rest tomorrow.)

refinery/user_files_manager/tests.py
mccalluc: That doesn't seem to be one that hits solr, which is the problem I'm dealing with.

refinery/user_files_manager/views.py
mccalluc: not needed with class-based view.

mccalluc:
Hit the NPM error again on the last build, and the log was dumped. I'll follow up on that.

mccalluc:
@scottx611x : ok to merge? The failure is the erratic NPM problem: We now have the logs, and the problem is the same class as those discussed at https://github.com/npm/npm/issues/9418 and https://github.com/npm/npm/issues/12484 : The suggested fixes are to upgrade npm, and (more contentiously) to just use http instead of https for connections to the registry.

I'll look into the npm upgrade, but I don't think that's necessary here.

refinery/user_files_manager/views.py
```diff
@@ -1,21 +1,21 @@
 import logging
-from django.http import HttpResponseNotAllowed
+
+from rest_framework.views import APIView
 from rest_framework.response import Response
-from data_set_manager.utils import (
-    search_solr, format_solr_response,
-    generate_solr_params_for_user)
+
+from data_set_manager.utils import (search_solr, format_solr_response,
```
scottx611x: Check out https://github.com/refinery-platform/refinery-platform/wiki/coding-style-guide#python

The imports  in parens should be in alphabetical order

refinery/core/views.py
```diff
@@ -78,6 +78,11 @@ def statistics(request):
                               context_instance=RequestContext(request))
 
 
+def user_files(request):
```
scottx611x: Should this be in `user_files_manager.views`?

hackdna:
@mccalluc we are using [node_6.x](https://github.com/refinery-platform/refinery-platform/blob/046c8ac130f364c7c9ebe991f9af24ed735f3aad/deployment/refinery-modules/refinery/manifests/init.pp#L85) while the latest is [node_8.x](https://github.com/nodesource/distributions#installation-instructions), so it's probably a good idea to upgrade anyway.

refinery/data_set_manager/utils.py
```diff
@@ -503,7 +505,44 @@ def _index_annotated_nodes(node_type, study_uuid, assay_uuid=None,
...
+        facet_filter - adds params to facet fields&fqs for filtering on fields
+        facet_pivot - list of fields to pivot
+        sort - Ordering include field name, whitespace, & asc or desc.
+        fq - filter query
+     """
+
+    assay_uuids = []
+    datasets = get_objects_for_user(user, "core.read_dataset")
+    for dataset in datasets:
+        version_details = dataset.get_version_details()
```
scottx611x: Generally we try to keep the code in `try` blocks to a minimum

refinery/user_files_manager/views.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class UserFiles(APIView):
+    def get(self, request):
+        params = request.query_params
+
+        solr_params = generate_solr_params_for_user(params, user=request.user)
```
hackdna: This call can now raise exceptions - they should be handled.

refinery/data_set_manager/utils.py
mccalluc: I agree in general, but I think that here, breaking out the checks line by line will require a more complicated structure than checking the whole set. That said, I might be looking at it wrong, so I'll go in the direction you suggest and we see if that's better.

hackdna: Separating checks allows for much easier debugging in production when all you have to go on is the log.

refinery/user_files_manager/views.py
mccalluc: I might be missing something: `*.DoesNotExist` does not reflect a real error, so we can capture it and it won't reach this level. On the other hand, MultipleObjectsReturned does represent a real unexpected error, and letting this go all the way up and manifest as a 500 seems appropriate.

hackdna: DRF handles only a few exceptions. You have to handle all other ones inside the views and probably return a more useful response than a 500 error (e.g., a descriptive JSON-encoded error message).
http://www.django-rest-framework.org/api-guide/exceptions/

refinery/data_set_manager/utils.py
```diff
@@ -522,8 +522,18 @@ def generate_solr_params_for_user(params, user):
...
+    try:
+        user = User.objects.get(uuid=user_uuid)
+    except User.DoesNotExist:
+        pass
+
+    if user:
+        datasets = get_objects_for_user(user, "core.read_dataset")
+    else:
+        datasets = []
+
```
mccalluc: A different approach would be:
```python
datasets = []
try:
    user = User.objects.get(uuid=user_uuid)
    datasets = get_objects_for_user(user, "core.read_dataset")
except User.DoesNotExist:
    pass
```
but my sense is that limiting the scope of the `try` is really important to us?

refinery/user_files_manager/views.py
```diff
@@ -0,0 +1,31 @@
+import logging
+
+from rest_framework.views import APIView
+from rest_framework.response import Response
+
+from data_set_manager.utils import (search_solr, format_solr_response,
```
scottx611x: check alphabetical order of imports here again

PR #1757, opened 2017-06-01
Jkmarx/data set 2 list display changes

jkmarx:
Ref #1649 
- Move input group column template to own directive
- Adjust template to show color and correct icon
- Add border with a row is active or in a group
- Remove control panel 'input details' for lists
screen shot 2017-06-01 at 9 36 51 am
screen shot 2017-06-01 at 9 36 30 am
screen shot 2017-06-01 at 9 36 21 am

### DIFF
--- a/refinery/ui/source/js/file-browser/partials/input-groups-column-template.html
+++ b/refinery/ui/source/js/file-browser/partials/input-groups-column-template.html
@@ -1,33 +1,40 @@
-<div class="ngCellText grid-input-groups">
-  <div
-    ng-if="grid.appScope.nodeSelectCollection[row.entity.uuid].groupList.length > 0"
-    class="selected-node"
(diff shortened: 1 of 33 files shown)


refinery/ui/source/js/file-browser/partials/input-groups-column-template.html
```diff
@@ -0,0 +1,33 @@
+<div class="ngCellText grid-input-groups">
+  <div
+    ng-if="grid.appScope.nodeSelectCollection[row.entity.uuid].groupList.length > 0"
```
hackdna: Would it makes sense to factor out `grid.appScope.nodeSelectCollection[row.entity.uuid]`? It is used several times just in this section.

jkmarx: Some of the variable aren't available until the ui-grid generation.

PR #1759, opened 2017-06-02
Mccalluc/user files integration

mccalluc:
Continuing the work from #1749.

[Should solr be running?](1760)


mccalluc:
Replaced by #1772.

PR #1763, opened 2017-06-05
Flekschas/isa tab import no annotated node explosion

flekschas:
- Extend ISA Mode by _Metabolite Assignment File_
- Heuristic to avoid exponential creation of annotated nodes (https://github.com/refinery-platform/refinery-platform/compare/flekschas/isa-tab-import-no-annotated-node-explosion?expand=1#diff-ee8c7d170403d51a67e071466c692f1dR462)
- Removed unused indices that bloated up the db for no good reason
- Removed code duplication
- Lots of PEP8 and readability issues fixed


flekschas:
Has anyone seen the TravisCI `KeyError: u"Migration core.0001_initial dependencies reference nonexistent parent node (u'data_set_manager', u'0001_initial')"` before? It's odd because my other two branches have merged the exact same code from _develop_ and run just fine.

scottx611x:
@fritz regarding the migration error above, I haven't had time to look into it deeply, but the new migration you've introduced looks like a `south` SchemMigration which isn't used in django 1.7+

`./manage.py makemigrations` is the new command to run

Also I would try to keep the migration #'s in order so the `0006` should be `0002`

flekschas:
@scottx611x Maybe that's the issue. I'll have a look into that.

scottx611x:
This looks ok to me @flekschas 

I'd be more comfortable merging if there was a test to cover it. I'm willing to write one if you gave me some plain english as to the outcome that we now expect from this code. 

i.e. `When isa-tab X is imported we expect less than N annotated nodes`

flekschas:
@scottx611x 

A test should catch the follow case:

When an ISA-Tab with N raw data files (nodes) with A unique attributes is imported and all M nodes (of any type, e.g., _derived data file_) have M * A unique attributes the isatab import should be cancelled. In other words, when every node represents all attributes, then the import should be canceled.

The above scenario can happen if all initial attributes of nodes are inherited to every child node.

Some MetaboLights ISA-Tab that should be unimportable are:
- MTBLS1
- MTBLS30
- MTBLS112

scottx611x:
@flekschas I've written some tests regarding the isatabs you've posted, but to me they all seem to be importable. This is the test case I've provided:
```
class AnnotatedNodeExplosionTestCase(TestCase):
    def setUp(self):
        self.username = 'coffee_lover'
        self.password = 'coffeecoffee'
        self.user = User.objects.create_user(
            self.username,
            '',
            self.password
        )

    def test_metabolights_isatab_wont_import_a(self):
        call_command(
            "process_isatab",
            self.user.username,
            "data_set_manager/test-data/MTBLS1.zip"
        )
        self.assertEqual(DataSet.objects.count(), 0)

    def test_metabolights_isatab_wont_import_b(self):
        call_command(
            "process_isatab",
            self.user.username,
            "data_set_manager/test-data/MTBLS30.zip"
        )
        self.assertEqual(DataSet.objects.count(), 0)

    def test_metabolights_isatab_wont_import_c(self):
        call_command(
            "process_isatab",
            self.user.username,
            "data_set_manager/test-data/MTBLS112.zip"
        )
        self.assertEqual(DataSet.objects.count(), 0)

```

hackdna:
[Improve data model of data sets](https://trello.com/c/uPviGMWp/199-improve-data-model-of-data-sets)

PR #1766, opened 2017-06-06
Scottx611x/cleanup tool containers

scottx611x:
Add `post_delete` signal to handle removal of Tool Docker containers

### DIFF
--- a/refinery/tool_manager/models.py
+++ b/refinery/tool_manager/models.py
@@ -10,7 +10,9 @@ from django.http import HttpResponseServerError, JsonResponse
 
 from django_extensions.db.fields import UUIDField
 from django_docker_engine.container_managers.local import LocalManager
-from django_docker_engine.docker_utils import DockerContainerSpec
+from django_docker_engine.docker_utils import (DockerClientWrapper,
+                                               DockerContainerSpec)
+
 from docker.errors import APIError
 
 from core.models import Analysis, OwnableResource, WorkflowEngine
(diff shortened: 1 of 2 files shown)


refinery/tool_manager/tests.py
```diff
@@ -1169,6 +1170,33 @@ def test_bad_POST_transaction_rollback(self):
...
+            tool_annotation = [json.loads(f.read())]
+
+        with mock.patch(self.mock_vis_annotations_reference,
+                        return_value=tool_annotation):
+            call_command("generate_tool_definitions", visualizations=True)
+
+        td = ToolDefinition.objects.all()[0]
+        post_data = {
+            "tool_definition_uuid": td.uuid,
+            "file_relationships": "['www.coffee.com/cool_file.txt']"
```
hackdna: Should be a IANA-managed Reserved Domain: https://github.com/refinery-platform/refinery-platform/wiki/Copy-Style-Guide#miscellaneous

refinery/tool_manager/models.py
```diff
@@ -305,6 +307,19 @@ def update_file_relationships_string(self):
         self.save()
 
 
+@receiver(post_delete, sender=Tool)
```
hackdna: If container is created after tool instance, I'd make it a pre_delete handler, so container would be stopped before tool instance is deleted to preserve logical order.

refinery/tool_manager/tests.py
```diff
@@ -1192,12 +1192,10 @@ def test_tool_container_removed_on_deletion(self):
         self.view(post_request)
 
         tool = Tool.objects.get(tool_definition__uuid=td.uuid)
+        tool_uuid = tool.uuid
```
hackdna: Why is `tool_uuid` variable necessary?

scottx611x: When I delete the Tool I lose access to its members, and I need its UUID for the DockerClientWrapper call below

hackdna: I see. Actually, there is no need to test external libraries: `DockerClientWrapper().purge_by_label()` in `remove_tool_container()` should be mocked and then tested if it was called then there is no need to save the tool UUID.

scottx611x: If I mock the purging of the container, then its never actually purged in the test

Issue #877, opened 2016-02-09
Separate requirements.txt into dev and prod versions


hackdna:

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;L5rFZqwx&#x2F;96-separate-requirements-txt-into-dev-and-prod-versions"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Separate requirements.txt into dev and prod versions</a>

hackdna:
There's now also `deployment/aws-requirements.txt`

mccalluc:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;HJJG10uq&#x2F;36-877"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> #877</a>

PR #1771, opened 2017-06-07
isort prehook

### LINKED ISSUE #1770: use isort to reorder imports, and add to flake8
Misordering imports can hold up a PR, for example https://github.com/refinery-platform/refinery-platform/pull/1749#pullrequestreview-42446364

Computers are better about alphabetical order than I am. Suggest adding https://github.com/gforcada/flake8-isort to our flake8 checks, and as needed, running isort to get things in the correct order.

This might touch a lot of files, and it would be nice to minimize conflicts with outstanding branches.

mccalluc:
Fix #1770

Developers will need to install `isort`: With the default settings, the imports are sorted in place, and the pre-commit will fail, and you then need to `git add` and `git commit` again. 

There are a lot of [config options](https://github.com/timothycrosley/isort/wiki/isort-Settings), but there doesn't seem to be a way to keep our blank lines between libraries.

It seems that this only checks the files which were touched in the commit. I don't think it would be that useful to do a bulk reformat of everything retrospectively. I've touched the files under `user_files_manager` just as a demonstration.

I think the increment of build complexity is worth not having to worry about import order, but it's not really a big gain, either.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,3 +9,8 @@
     hooks:
     -   id: eslint
         files: ui/source/js/.+\.js
+
+-   repo: git://github.com/hvdklauw/pre-commit-isort
+    sha: 3cddb8006cd2188e4af72454f2faeec25b040821
+    hooks:
+    -   id: isort
--- a/refinery/user_files_manager/tests.py
+++ b/refinery/user_files_manager/tests.py
@@ -2,7 +2,6 @@ import logging
 
 from django.test import TestCase
 
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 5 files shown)


scottx611x:
@hackdna I've personally had PR reviews that have taken 3 or 4 days for someone to get around to reviewing be delayed for another day just for simple import ordering fixes.

I think this could be helpful and take a small burden off of ourselves and future contributors.

mccalluc:
@hackdna : I'll go ahead and create a requirements-dev.txt in this PR if that allays your concerns?

hackdna:
@mccalluc thanks but there is already one: #877

mccalluc:
@hackdna : Sorry for the miscommunication: Unless I'm missing something, `requirements-dev.txt` does not exist... Or is your point that `requirements-aws.txt` is already doing the job? Is there anything you would like to see here that would make you comfortable with a merge?

hackdna:
Just approved this PR.

There is no `requirements-dev.txt` yet but we should create one at some point. That's what #877 is about.

PR #1772, opened 2017-06-07
Mccalluc/user files next steps

mccalluc:
This continues the work on user file listing begun in #1749. This PR replaces https://github.com/refinery-platform/refinery-platform/pull/1759. This is generally about better or more graceful error handling.

### DIFF
--- a/refinery/data_set_manager/utils.py
+++ b/refinery/data_set_manager/utils.py
@@ -811,8 +811,6 @@ def format_solr_response(solr_response):
         solr_response_json = json.loads(solr_response)
     except TypeError:
         return "Error loading json."
-    assert solr_response_json['responseHeader']['status'] == 0, \
-        solr_response_json['error']['msg']
 
     # Reorganizes solr response into easier to digest objects.
     try:
--- a/refinery/data_set_manager/utils.py
+++ b/refinery/data_set_manager/utils.py
@@ -818,9 +818,10 @@ def format_solr_response(solr_response):
     try:
         order_facet_fields_joined = (solr_response_json
                                      ['responseHeader']['params']['fl'])
-        order_facet_fields = order_facet_fields_joined.split(',')
     except KeyError:
         order_facet_fields = []
+    else:
+        order_facet_fields = order_facet_fields_joined.split(',')
 
     if solr_response_json.get('facet_counts'):
         facet_field_counts = solr_response_json.get('facet_counts').get(
--- a/refinery/data_set_manager/utils.py
+++ b/refinery/data_set_manager/utils.py
@@ -615,12 +615,7 @@ def _generate_solr_params(params, assay_uuids=[]):
                   'facet.limit=-1'
                   ])
 
-    solr_params = 'fq=({})'.format(
-        ' OR '.join(map(
-            lambda id: 'assay_uuid:{}'.format(id),
-            assay_uuids
-        ))
-    )
+    solr_params = 'fq=assay_uuid:({})'.format(' OR '.join(assay_uuids))
 
     if facet_field:
         facet_field = facet_field.split(',')
@@ -806,6 +801,8 @@ def format_solr_response(solr_response):
         solr_response_json = json.loads(solr_response)
     except TypeError:
         return "Error loading json."
+    assert solr_response_json['responseHeader']['status'] == 0, \
+        solr_response_json['error']['msg']
 
     # Reorganizes solr response into easier to digest objects.
     order_facet_fields = solr_response_json.get('responseHeader').get(
--- a/refinery/data_set_manager/utils.py
+++ b/refinery/data_set_manager/utils.py
@@ -804,8 +804,13 @@ def format_solr_response(solr_response):
         solr_response_json['error']['msg']
 
     # Reorganizes solr response into easier to digest objects.
-    order_facet_fields = solr_response_json.get('responseHeader').get(
-            'params').get('fl').split(',')
+    order_facet_fields_joined = solr_response_json.get('responseHeader')\
+        .get('params').get('fl')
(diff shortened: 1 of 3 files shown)


refinery/data_set_manager/utils.py
```diff
@@ -544,21 +544,19 @@ def generate_solr_params_for_user(params, user_uuid):
...
             )
         except Study.DoesNotExist:
-            continue
-            # It's not an error not to have data,
-            # but there's nothing more to do here.
+            logger.error('Expected at least one Study for %s', investigation)
+            raise
         except Study.MultipleObjectsReturned:
-            logger.error('Expected only one study for %s', investigation)
+            logger.error('Expected only one Study for %s', investigation)
```
jkmarx: except (Study.DoesNotExist, Study.MultipleObjectsReturned) as e:

We tend to include e in the logger.

refinery/data_set_manager/utils.py
```diff
@@ -544,21 +544,19 @@ def generate_solr_params_for_user(params, user_uuid):
...
 
         try:
             assay = Assay.objects.get(study=study)
         except Assay.DoesNotExist:
-            continue
-            # Again, it's not an error not to have data,
-            # but there's nothing more to do here.
-        except:
-            logger.error('Expected only one assay for %s', study)
+            logger.error('Expected at least one Assay for %s', study)
```
scottx611x: same as Jennifer's remarks above

refinery/data_set_manager/utils.py
```diff
@@ -615,12 +613,8 @@ def _generate_solr_params(params, assay_uuids=[]):
                   'facet.limit=-1'
                   ])
 
-    solr_params = 'fq=({})'.format(
-        ' OR '.join(map(
-            lambda id: 'assay_uuid:{}'.format(id),
-            assay_uuids
-        ))
-    )
+    assert len(assay_uuids) > 0, 'At least one assay must be accessible'
```
scottx611x: I don't think `assert` is inline with what we've been doing. We generally have been using try/excepts, but i do enjoy its simplicity. Maybe it could be up for debate?

Presumably 
AssertionError is raised? so at the very least that should be handled.

refinery/data_set_manager/utils.py
```diff
@@ -806,10 +800,17 @@ def format_solr_response(solr_response):
         solr_response_json = json.loads(solr_response)
     except TypeError:
         return "Error loading json."
+    assert solr_response_json['responseHeader']['status'] == 0, \
```
scottx611x: same as above

refinery/data_set_manager/utils.py
```diff
@@ -806,10 +800,17 @@ def format_solr_response(solr_response):
...
+        solr_response_json['error']['msg']
 
     # Reorganizes solr response into easier to digest objects.
-    order_facet_fields = solr_response_json.get('responseHeader').get(
-            'params').get('fl').split(',')
+    order_facet_fields_joined = solr_response_json.get('responseHeader')\
```
scottx611x: We've been generally avoiding `\`

scottx611x: Also I think the chaining of `get()`s  would be cleaner like so:
```
try:
    order_facet_fields_joined = solr_response_json["response_header"]["params"]["fl"]
except KeyError as e:
    # handle error
else:
    # forge ahead
```

refinery/user_files_manager/views.py
```diff
@@ -21,9 +24,14 @@ class UserFiles(APIView):
     def get(self, request):
         params = request.query_params
 
-        solr_params = generate_solr_params_for_user(
-            params,
-            user_uuid=request.user.uuid)
+        try:
+            solr_params = generate_solr_params_for_user(
```
scottx611x: Minor styling: 
```
solr_params = generate_solr_params_for_user(
        params,
        user_uuid=request.user.uuid
)
```

refinery/data_set_manager/utils.py
mccalluc: Nothing will go wrong at the next line if `assya_uuids == []`, except that the solr query constructed is invalid: so it seems best to catch it here. That said, perhaps what we really want is a custom error, that can be caught a couple layers up.

mccalluc: Just committed a fix I like: We're inside a method which is building the query string. rather than using errors, we just return None is there query string would be invalid, and then at the top level that can be handled.

refinery/data_set_manager/utils.py
mccalluc: @scottx611x : I would like to leave this, actually: If solr is returning error responses to us, then we really do have a problem, and letting it bubble up and manifest as a 500 is appropriate: It's not something we can or should recover from.

This assertion isn't necessary for my code, but it will make it easier for any developer in the future by pointing them in the right direction.

mccalluc:
@scottx611x : see if this addresses everything.

refinery/data_set_manager/utils.py
hackdna: @mccalluc It may be useful to return HTTP 500 response in a case like this but exceptions should not be left unhandled: "Care must be taken when handling exceptions to ensure proper application cleanup while maintaining useful error reporting." https://doughellmann.com/blog/2009/06/19/python-exception-handling-techniques/

mccalluc: I'm not on the same page right now, but this line shouldn't block the PR. I'll revert it; meanwhile, I've filed #1789 to try to explain where I'm coming from. Sorry for the distraction.

mccalluc:
@scottx611x : I think that one line was the last remaining problem? Let me know if you see anything else.

PR #1784, opened 2017-06-09
Mccalluc/user files continued

mccalluc:
This fleshes out the UI with mock data and adds tests of the end points. I'll want to get some changes from develop merged into this before merging.

### DIFF
--- a/refinery/config/urls.py
+++ b/refinery/config/urls.py
@@ -4,7 +4,6 @@ from django.conf import settings
 from django.conf.urls import include, patterns, url
 from django.conf.urls.static import static
 from django.contrib import admin
-
 from haystack.forms import FacetedSearchForm
 from haystack.query import SearchQuerySet
 from haystack.views import FacetedSearchView
@@ -18,23 +17,21 @@ from core.api import (AnalysisResource, DataSetResource, ExtendedGroupResource,
                       NodeRelationshipResource, NodeResource,
                       NodeSetListResource, NodeSetResource, ProjectResource,
                       StatisticsResource, UserAuthenticationResource,
-                      UserProfileResource,
-                      WorkflowInputRelationshipsResource, WorkflowResource)
+                      UserProfileResource, WorkflowInputRelationshipsResource,
+                      WorkflowResource)
 from core.forms import RegistrationFormWithCustomFields
-from core.models import DataSet, AuthenticationFormUsernameOrEmail
+from core.models import AuthenticationFormUsernameOrEmail, DataSet
 from core.urls import core_router
 from core.views import CustomRegistrationView
 from data_set_manager.api import (AssayResource, AttributeOrderResource,
                                   AttributeResource, InvestigationResource,
-                                  ProtocolReferenceResource,
                                   ProtocolReferenceParameterResource,
-                                  ProtocolResource, PublicationResource,
-                                  StudyResource)
+                                  ProtocolReferenceResource, ProtocolResource,
+                                  PublicationResource, StudyResource)
 from data_set_manager.urls import data_set_manager_router
-from user_files_manager.urls import user_files_router, user_files_url
 from file_store.urls import file_store_router
-
-from tool_manager.urls import tool_manager_router, django_docker_engine_url
+from tool_manager.urls import django_docker_engine_url, tool_manager_router
+from user_files_manager.urls import user_files_router, user_files_url
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


refinery/user_files_manager/tests.py
```diff
@@ -1,16 +1,43 @@
...
 
 
-class UserFilesAPITests(TestCase):
-    pass
-    # TODO: Fails, because solr isn't running.
-    # Should I mock it, or make this an integration test?
+# def fake_search_solr():
+#     return {}
+#
+# @mock.patch("user_files_manager.utils.search_solr",
```
scottx611x: I'd remove comments, but I'm actually more interested if you are able to mock a function call for the entirety of a test suite, this would be really useful for me.

refinery/user_files_manager/tests.py
```diff
@@ -1,16 +1,43 @@
...
+#                 fake_search_solr)
+class UserFilesAPITests(APITestCase):
+    def setUp(self):
+        self.factory = APIRequestFactory()
+        self.view = UserFiles.as_view()
+        self.url_root = '/api/v2/user/files/'
+        self.user = User.objects.get(pk=-1)  # anonymous user
```
scottx611x: I think there is a `get_anonymous_user()` provided by `guardian` already

refinery/user_files_manager/views.py
```diff
@@ -23,8 +23,10 @@ def get(self, request):
 
         solr_params = generate_solr_params_for_user(
             params,
-            user_uuid=request.user.uuid)
+            user_id=request.user.id)
         solr_response = search_solr(solr_params, 'data_set_manager')
+        # import pdb
```
scottx611x: remove comments

refinery/data_set_manager/utils.py
```diff
@@ -536,17 +536,20 @@ def generate_solr_params_for_user(params, user_uuid):
 
     assay_uuids = []
     for dataset in datasets:
-        investigation_links = dataset.get_investigation_links()
+        investigation_link = dataset.get_latest_investigation_link()
+        if not investigation_link:
```
scottx611x: I think `if investigation_link is None:` is more appropriate here

mccalluc:
(The usual npm build problem, and one I haven't seen before? Restarted both on Travis.)

mccalluc:
@scottx611x : Could you re-review this? I think your concerns are addressed.

refinery/data_set_manager/utils.py
```diff
@@ -9,22 +9,19 @@
 import time
 import urlparse
 
+import core
```
scottx611x: `import core` should be below with our other homemade django apps

mccalluc: I can look into customizing the behavior of isort if it really matters, but my sense is that we want some consistent order, but which particular order it is doesn't matter so much?

mccalluc: I think `known_first_party` would do it, but we'd have to list all the modules we wanted handled that way. Significant maintenance cost / little benefit. (Also: I have got it to complain about syntax errors in the config file, but I haven't actually seen it make any difference: I'm doing something wrong, and it feels like another thing that might break in the future.)

hackdna: We won't be able to automate everything and it's OK because we do have a coding style guide where [Python module import order](https://github.com/refinery-platform/refinery-platform/wiki/coding-style-guide#python) is described. It is also a part of [PEP8](https://www.python.org/dev/peps/pep-0008/#imports).

refinery/user_files_manager/tests.py
```diff
@@ -1,15 +1,37 @@
 import logging
+from urlparse import urljoin
 
-from django.test import TestCase
+import requests
+from django.contrib.staticfiles.testing import StaticLiveServerTestCase
+from guardian.utils import get_anonymous_user
+from rest_framework.test import (APIRequestFactory, APITestCase,
+                                 force_authenticate)
+from views import UserFiles
```
scottx611x: Minor but: `from .views import UserFiles`

refinery/config/urls.py
```diff
@@ -31,7 +31,7 @@
                                   ProtocolResource, PublicationResource,
                                   StudyResource)
 from data_set_manager.urls import data_set_manager_router
-from user_files_manager.urls import user_files_router
+from user_files_manager.urls import user_files_router, user_files_url
```
scottx611x: Shouldn't this import be below: `from file_store.urls import file_store_router` ? (not sure how isort is doing this)

mccalluc:
@scottx611x : As discussed, isort only checks those files modified in a commit, but as a demonstration, here are the two files you mentioned, sorted.

PR #1790, opened 2017-06-12
Jkmarx/generalize attribute filter

jkmarx:
(Most of the code changes moves the methods around.)
This pull request untangles the attribute/analysis filter from the main ui-grid ctrl. 
- Creating an assay filter ctrl and moved variables/methods from the main ctrl
- Created services to share/track files params and the attribute/analysis filter data structure used by ui
- Added jsDoc comments to new files
- Added unit tests

### DIFF
--- a/refinery/ui/source/js/file-browser/ctrls/assay-filters-ctrl.js
+++ b/refinery/ui/source/js/file-browser/ctrls/assay-filters-ctrl.js
@@ -38,6 +38,7 @@
      * attributeInternalName: {fieldName: boolean, fieldName: boolean} */
     vm.uiSelectedFields = selectedFilterService.uiSelectedFields;
     vm.updateFiltersFromUrlQuery = updateFiltersFromUrlQuery;
+    vm.updateFilterDOM = false;
 
     activate();
    /*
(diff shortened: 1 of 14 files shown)


refinery/ui/source/js/file-browser/ctrls/assay-filters-ctrl.js
```diff
@@ -0,0 +1,212 @@
+/**
+ * Assay Filters Ctrl
+ * @namespace AssayFiltersCtrl
```
jkmarx: agreed....perhaps move it to common services or a separate module. It's really dependent on a specific api data structure. Let's wait until it works out for the user/files module and then discuss moving it.

refinery/ui/source/js/file-browser/partials/assay-filters.html
```diff
@@ -1,5 +1,4 @@
-
-<div ng-if="FBCtrl.attributeFilter">
+<div ng-if="ASCtrl.attributeFilter">
```
jkmarx: We added a ctrl for the directive, so no longer need to use the parent ctrl.

refinery/ui/source/js/file-browser/services/assay-filters-service.spec.js
```diff
@@ -0,0 +1,94 @@
...
+              'Harslem/Heafner': 1,
+              McConnell: 6,
+              'Crocker + McConnell': 4,
+              Crocker: 4,
+              'Postel/Cerf': 1,
+              Cotton: 1
+            }
+          }
+        };
+      }));
```
jkmarx: Point of discussion, but I like to mock real data.

PR #1794, opened 2017-06-12
Upgrade node and npm. Hopefully fix NPM failures? Fix #1768

### LINKED ISSUE #1768: Upgrade NPM to sporadic Travis errors
https://github.com/refinery-platform/refinery-platform/pull/1749#issuecomment-306521020

- First step would be to check what version Travis is giving us by default, and only if that is particularly old, try to update it. 
- Some folks reported that changing to http fixed things: There are good reasons not to do that in production, but I would be ok doing that just for Travis... though objections would be reasonable, too.
- Simply retrying might also get around the problem: If errors are independent, trying twice would radically reduce the already low error rate.

If neither of these seem to work, or seem appropriate, I think it would also be ok to close this issue, since we are pretty confident it has nothing to do with our code.

(In response to [Ilya](https://github.com/refinery-platform/refinery-platform/pull/1749#issuecomment-306537024), the advice seems to be upgrading npm, rather than node itself, though that might be a good thing, too.)

mccalluc:
Here are the corresponding node and npm versions: https://nodejs.org/en/download/releases/

The Travis line wouldn't need to be so complicated, but I want to be sure that we remove it when it is no longer needed.

### DIFF
--- a/deployment/refinery-modules/refinery/manifests/init.pp
+++ b/deployment/refinery-modules/refinery/manifests/init.pp
@@ -82,7 +82,7 @@ class ui {
   apt::source { 'nodejs':
     ensure      => 'present',
     comment     => 'Nodesource NodeJS repo.',
-    location    => 'https://deb.nodesource.com/node_6.x',
+    location    => 'https://deb.nodesource.com/node_8.x',
     release     => 'trusty',
     repos       => 'main',
     key         => {
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,10 @@ services:
   - postgresql
   - neo4j
   - docker
-
 install:
+  # If Travis does not have a current npm, install.
+  # (If it does, exit: This line should then be removed.)
+  - if [[ `npm -v` != 5.* ]]; then npm install -g npm@5; else exit 1; fi
   - pip install -r requirements.txt --quiet
   # Install geckodriver required for selenium testing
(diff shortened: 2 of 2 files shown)


deployment/refinery-modules/refinery/manifests/init.pp
```diff
@@ -82,7 +82,7 @@
   apt::source { 'nodejs':
     ensure      => 'present',
     comment     => 'Nodesource NodeJS repo.',
-    location    => 'https://deb.nodesource.com/node_6.x',
+    location    => 'https://deb.nodesource.com/node_8.x',
```
hackdna: This should be thoroughly tested before merging into develop:
* both on AWS (launch stacks with and without snapshots) and Vagrant (provision existing and brand new VMs)
* deployment logs should be examined for any errors or warnings in all cases
* at least the home page should be loaded and browser console should be checked for any errors or warnings in all cases

mccalluc:
@scottx611x : Maybe the next thing with this would to be to try it on AWS instances as Ilya suggests? Would you like to give me pointers on that? (That is, assuming these errors are worth the trouble, and we don't have any other work in progress which would make this redundant?)

mccalluc:
(We decided that this the same vein as the Travis upgrade, which needs to happen sooner, so we'll do that first, and then come back here and see if it's still relevant.)

mccalluc:
Closing this in favor of #1930

PR #1797, opened 2017-06-14
Jkmarx/data set 2 tools param

jkmarx:
Ref: #1649
- Add parameters panel
- Display parameters
- Form to update parameters
- Unit tests/Comments
- Add form validation for integers and floats

parameters


refinery/ui/source/js/tool-launch/directives/tool-params.js
```diff
@@ -0,0 +1,20 @@
...
+  angular
+    .module('refineryToolLaunch')
+    .component('rpToolParams', {
+      controller: 'ToolParamsCtrl',
+      require: {
+        displayCtrl: '^rpToolDisplay'
+      },
+      templateUrl: ['$window', function ($window) {
+        return $window.getStaticUrl('partials/tool-launch/partials/tool-params.html');
+      }]
```
mccalluc: Could you explain what's going on here, and why a plain string wouldn't work?

jkmarx: A plain string would work here. A few weeks ago Ilya added a method to generate the static url based on django settings. So we use that method to generate the static urls as seen above.

refinery/ui/source/js/tool-launch/partials/tool-params.html
```diff
@@ -0,0 +1,129 @@
...
+                   </h4>
+                   <input
+                     autofocus
+                     type="text"
+                     name="{{param.uuid}}"
+                     ng-class="{'form-validation-error': toolParamForm[param.uuid].$error.pattern}"
+                     placeholder="{{ param.default_value }}"
+                     ng-model="$ctrl.paramsForm[param.uuid]"
+                     ng-maxlength="32"
+                     ng-pattern="/^\d+$/">
```
mccalluc: Is this field allowed to be blank? Leading zeros probably won't cause problems?

refinery/ui/source/js/tool-launch/partials/tool-params.html
```diff
@@ -0,0 +1,129 @@
...
+                   </h4>
+                   <input
+                     autofocus
+                     type="text"
+                     name="{{param.uuid}}"
+                     ng-class="{'form-validation-error': toolParamForm[param.uuid].$error.pattern}"
+                     placeholder="{{ param.default_value }}"
+                     ng-model="$ctrl.paramsForm[param.uuid]"
+                     ng-maxlength="32"
+                     ng-pattern="/^[0-9]+(\.[0-9]{1,2})?$/">
```
mccalluc: should handle negatives? More than two decimal points? why [0-9] instead of \d?

refinery/ui/source/js/tool-launch/services/tool-params-service.spec.js
```diff
@@ -0,0 +1,59 @@
...
+          file_relationship: [],
+          input_files: [],
+          uuid: mocker.generateUuid(),
+          name: 'List of FASTQ Files',
+          value_type: 'LIST'
+        },
+        output_files: [],
+        parameters: [
+          {
+            galaxy_workflow_step: 1,
```
scottx611x: Do we need this `galaxy_workflow_step` in tests?

refinery/ui/source/styles/refinery-style.less
```diff
@@ -1838,3 +1838,9 @@ ul.paragraph {
     float: left;
   }
 }
+
+.form-validation-error {
+  border: solid 1px rgba(246, 36, 0, 1);
```
mccalluc: Should we pull the color from a variable?

jkmarx: consistency because the shadow needs a transparency.

refinery/ui/source/js/tool-launch/partials/tool-params.html
jkmarx: No idea, but this pattern can be changed once we have actual parameter specs.

refinery/ui/source/js/tool-launch/partials/tool-params.html
jkmarx: Yes, there are default values for params. No idea, but this pattern can be changed once we have actual parameter specs.

refinery/ui/source/js/tool-launch/ctrls/tool-params-ctrl.js
```diff
@@ -0,0 +1,48 @@
...
+  ];
+
+
+  function ToolParamsCtrl (
+    $scope,
+    toolParamsService
+  ) {
+    var paramsService = toolParamsService;
+    var vm = this;
+    vm.isToolParamsCollapsed = false; // tracks the parameters panel
```
scottx611x: Thoughts on starting this collapsed? 
I think if people are really dying to edit a tools params they will, otherwise they'll probably run them with the defaults at first.

refinery/ui/source/js/tool-launch/services/tool-params-service.spec.js
```diff
@@ -0,0 +1,59 @@
...
+            default_value: '1337'
+          }
+        ],
+        uuid: mocker.generateUuid(),
+        name: 'FASTQC Quality Control',
+        description: 'Runs FASTQC quality control on a set of FASTQ files.',
+        tool_type: 'WORKFLOW',
+        image_name: '',
+        container_input_path: '',
+        galaxy_workflow_id: 'c0279aab05812500',
```
scottx611x: Same for`galaxy_workflow_id`

refinery/ui/source/js/tool-launch/services/tool-params-service.spec.js
jkmarx: I like to mock real-like data, but if we'd like to have a standard let's discuss.

PR #1799, opened 2017-06-14
Scottx611x/tool manager updates

scottx611x:
- Updates necessary to be in line with newer django_docker_engine
- Add volume association with `extra_directories`

### DIFF
--- a/refinery/tool_manager/migrations/0017_tooldefinition_extra_directories.py
+++ b/refinery/tool_manager/migrations/0017_tooldefinition_extra_directories.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tool_manager', '0016_auto_20170424_1753'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='tooldefinition',
+            name='extra_directories',
+            field=models.CharField(max_length=500, blank=True),
+            preserve_default=True,
+        ),
+    ]
--- a/refinery/tool_manager/models.py
+++ b/refinery/tool_manager/models.py
@@ -6,10 +6,10 @@ from django.conf import settings
 from django.db import models
 from django.db.models.signals import post_delete, pre_delete
 from django.dispatch import receiver
-from django.http import HttpResponseServerError, JsonResponse
+from django.http import (HttpResponseBadRequest, HttpResponseServerError,
+                         JsonResponse)
 
 from django_extensions.db.fields import UUIDField
-from django_docker_engine.container_managers.local import LocalManager
 from django_docker_engine.docker_utils import (DockerClientWrapper,
                                                DockerContainerSpec)
 
@@ -161,6 +161,10 @@ class ToolDefinition(models.Model):
         max_length=500,
         blank=True
     )
+    extra_directories = models.CharField(
+        max_length=500,
+        blank=True
+    )
     galaxy_workflow_id = models.CharField(
         max_length=250,
         blank=True
@@ -255,12 +259,16 @@ class Tool(OwnableResource):
                         self.file_relationships
                     )
                 },
-                manager=get_django_docker_engine_manager()
+                extra_directories=ast.literal_eval(
+                    self.tool_definition.extra_directories
+                )
             )
             try:
-                container.run()
+                DockerClientWrapper().run(container)
             except APIError as e:
                 return HttpResponseServerError(content=e)
+            except AssertionError as e:
+                return HttpResponseBadRequest(content=e)
             else:
                 return JsonResponse(
                     {
@@ -318,16 +326,3 @@ def remove_tool_container(sender, instance, *args, **kwargs):
         except APIError as e:
             logger.error("Couldn't purge container for Tool with UUID: %s %s",
                          instance.uuid, e)
-
-
-def get_django_docker_engine_manager():
-    """
-    Helper method to return the proper managerial class for
-    django_docker_engine
-    """
-    # Travis CI runs on EC2, but we want our tests running against a local
-    # docker engine there
-    if settings.DEPLOYMENT_PLATFORM == "aws":
-        raise NotImplementedError
-    else:
-        return LocalManager()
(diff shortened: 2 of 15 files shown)


refinery/tool_manager/migrations/0017_tooldefinition_extra_directories.py
```diff
@@ -0,0 +1,20 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='tooldefinition',
+            name='extra_directories',
+            field=models.CharField(max_length=500, blank=True),
+            preserve_default=True,
+        ),
+    ]
```
mccalluc: Out of my depth here, but is it correct that this is plural? Values are delimited? Would we want to model it differently?

refinery/tool_manager/models.py
```diff
@@ -255,12 +259,16 @@ def launch(self):
...
+                    self.tool_definition.extra_directories
+                )
             )
             try:
-                container.run()
+                DockerClientWrapper().run(container)
             except APIError as e:
                 return HttpResponseServerError(content=e)
+            except AssertionError as e:
+                return HttpResponseBadRequest(content=e)
```
mccalluc: I like assertions... but I'm not comfortable catching them: Over time, how do you know that it is the assertion you had in mind that's causing the trouble now? Perhaps better to have a new error class, if it's really necessary?

refinery/tool_manager/test_data/visualizations/higlass.json
```diff
@@ -3,7 +3,8 @@
   "tool_type": "VISUALIZATION",
   "annotation": {
     "container_input_path": "/data/input.json",
-    "image_name": "scottx611x/refinery-higlass-docker:v0.0.2",
+    "extra_directories": ["/refinery-data"],
+    "image_name": "scottx611x/refinery-higlass-docker:v0.1.0",
```
mccalluc: Semantics are different when we're at 0: could this just be 0.0.3?

refinery/tool_manager/tests.py
```diff
@@ -637,6 +636,39 @@ def test_workflow_list_not_enough_inputs(self):
             )
             self.assertEqual(ToolDefinition.objects.count(), 0)
 
+    def test_visualization_annotation_extra_dirs_valid(self):
+        with open(
+            "{}/visualizations/LIST_visualization_good_extra_directories.json"
+            .format(TEST_DATA_PATH)
```
mccalluc: os.path.join slightly preferable: portability, and makes intentions clear.

refinery/tool_manager/tests.py
```diff
@@ -637,6 +636,39 @@ def test_workflow_list_not_enough_inputs(self):
...
+        ) as f:
+            visualization_annotation = json.loads(f.read())
+            create_tool_definition(visualization_annotation)
+            self.assertEqual(ToolDefinition.objects.count(), 1)
+
+    def test_visualization_annotation_extra_dirs_invalid(self):
+        with open("{}/visualizations/"
+                  "LIST_visualization_bad_extra_directories_structure.json"
+                  .format(TEST_DATA_PATH)) as f:
```
mccalluc: ditto

refinery/tool_manager/tests.py
```diff
@@ -637,6 +636,39 @@ def test_workflow_list_not_enough_inputs(self):
...
+                RuntimeError,
+                validate_tool_annotation,
+                visualization_annotation
+            )
+            self.assertEqual(ToolDefinition.objects.count(), 0)
+
+    def test_visualization_annotation_extra_dirs_missing(self):
+        with open("{}/visualizations/"
+                  "LIST_visualization_missing_extra_directories.json"
+                  .format(TEST_DATA_PATH)) as f:
```
mccalluc: ditto

refinery/tool_manager/tests.py
mccalluc: would you like to add "visualizations" to TEST_DATA_PATH? It seems to be repeated a lot.

refinery/tool_manager/tests.py
```diff
@@ -1046,6 +1078,69 @@ def test_bad_POST_transaction_rollback(self):
         self.assertIn("LIST/PAIR structure is not balanced",
                       self.post_response.content)
 
+    def test_bad_extra_directories_path(self):
+        Tool.objects.all().delete()
+        with open("{}/visualizations/"
+                  "LIST_visualization_bad_extra_directories_path.json"
+                  .format(TEST_DATA_PATH)) as f:
```
mccalluc: ditto.

refinery/tool_manager/tests.py
```diff
@@ -1046,6 +1078,69 @@ def test_bad_POST_transaction_rollback(self):
...
+            visualization_annotation = json.loads(f.read())
+            create_tool_definition(visualization_annotation)
+
+        td = ToolDefinition.objects.get(
+            name=visualization_annotation["name"]
+        )
+
+        tool_launch_configuration = {
+            "tool_definition_uuid": td.uuid,
+            "file_relationships": str(["www.example.com"])
```
mccalluc: Not sure how it's used, but this seems like a url, not a file? and is a serialized python array the best choice? Maybe represent each separately, but if this is best, perhaps the name should be more explicit?

refinery/tool_manager/test_data/visualizations/higlass.json
mccalluc: Never mind, I was confusing this with django_docker_engine versions.

refinery/tool_manager/migrations/0017_tooldefinition_extra_directories.py
scottx611x: @mccalluc I think pluralization is okay for this implementation, but I've been thinking about this precisely for the items we currently have to `ast.literal_eval()`. 

We can't natively store JSON db-side until we get to a later Django version, but I think it would be adequate to have one `TextField` that I store a Json string in which can deserialize and interact with later.

refinery/tool_manager/models.py
scottx611x: Yeah, good call. I just did this to handle for the assertion we added for having absolute filesystem paths. 
Catching something less specific would probably be better here.

refinery/tool_manager/tests.py
scottx611x: We have `workflows` under `TEST_DATA_PATH` as well so I made the distinction a directory above.  
I have it on my list to clean up these long test strings (although they are quite descriptive)

refinery/tool_manager/tests.py
scottx611x: This is one of the other fields i'd encapsulate in the json I mentioned above 👍

PR #1805, opened 2017-06-16
Scottx611x/tool manager updates

scottx611x:
- Test refactoring
- Provide fields on `Tool` & `ToolDefinition` to dump json in rather than the `ast.literal_eval()` approach from before
- @mccalluc & @jkmarx: `tool_manager.tests.py`  has a lengthy diff, but most of it is just moving existing members into places that make more sense

### DIFF
--- a/refinery/tool_manager/tests.py
+++ b/refinery/tool_manager/tests.py
@@ -65,7 +65,7 @@ class ToolManagerTestBase(TestCase):
             }
         )
 
-        self.tools_url_root = '/api/v2/tools'
+        self.tools_url_root = '/api/v2/tools/'
         self.tool_defs_url_root = '/api/v2/tool_definitions/'
 
     def tearDown(self):
--- a/refinery/tool_manager/tests.py
+++ b/refinery/tool_manager/tests.py
@@ -14,14 +14,14 @@ from django.http import HttpResponseBadRequest, JsonResponse
 from django.test import TestCase
 from rest_framework.test import (APIRequestFactory, APITestCase,
                                  force_authenticate)
-from selenium_testing.utils import (MAX_WAIT, SeleniumTestBaseGeneric,
-                                    wait_until_class_visible)
 
 from core.models import (DataSet, ExtendedGroup, InvestigationLink,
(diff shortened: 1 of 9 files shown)


refinery/tool_manager/utils.py
```diff
@@ -1,15 +1,13 @@
...
 import os
 
+from bioblend.galaxy.client import ConnectionError
 from django.conf import settings
 from django.contrib import admin
 from django.db import transaction
-
-from bioblend.galaxy.client import ConnectionError
-from jsonschema import RefResolver, validate, ValidationError
+from jsonschema import RefResolver, ValidationError, validate
```
jkmarx: is that from isort? caps before lowercase?

scottx611x: Yes,  isort puts `class`es first in alph. order

refinery/tool_manager/tests.py
```diff
@@ -1,64 +1,150 @@
...
+        )
+        self.tool_defs_view = ToolDefinitionsViewSet.as_view(
+            {
+                'get': 'list',
+                'post': 'create'
+            }
+        )
 
-        self.url_root = '/api/v2/tool_definitions/'
+        self.tools_url_root = '/api/v2/tools'
```
jkmarx: trailing / ?

refinery/tool_manager/tests.py
```diff
@@ -368,6 +444,54 @@ def test_list_list_pair_workflow_tool_def_generation(self):
...
+    def test_list_pair_list_workflow_tool_def_validation(self):
+        with open(
+                "{}/workflows/LIST:PAIR:LIST.json".format(TEST_DATA_PATH)
+        ) as f:
+            workflow_annotation = json.loads(f.read())
+            self.assertIsNone(
+                validate_tool_annotation(workflow_annotation)
+            )
+
+    def test_list_pair_list_workflow_tool_def_generation(self):
```
jkmarx: a lot of assetEquals, perhaps break into assert the different types separately.

Issue #1813, opened 2017-06-20
Upgrade all deployment platforms to Ubuntu 14.04.5


hackdna:
We are using Ubuntu 14.04.2 on Vagrant and AWS at the moment. [Travis CI is in the process of moving to the latest version of 14.04.5](https://blog.travis-ci.com/2017-06-19-trusty-updates-2017-Q2) ~~and it is causing builds to fail~~.

Ubuntu 14.04.5 is the latest in the 14.04 LTS series, while 14.04.2 is almost 2.5 years old at this point (https://wiki.ubuntu.com/Releases).

We should standardize on Ubuntu 14.04.5 across all deployment platforms (Vagrant, Travis and AWS).

We should probably [constrain](https://www.vagrantup.com/docs/boxes/versioning.html#version-constraints) Vagrant environment to a specific version of the Ubuntu box using the Vagrantfile by specifying the `config.vm.box_version` option.

mccalluc:
Huh: So my last build was on the new version, and there weren't any problems. Don't know whether I wasn't building the right way before, or if they put in some last minute fixes.
- Removing "critical" here
- and I'll delete my experimental branches.

hackdna:
FWIW, my [build](https://travis-ci.org/refinery-platform/refinery-platform/builds/245479188) just failed:
```
$ bower install --config.interactive=false --quiet
module.js:472
    throw err;
    ^

Error: Cannot find module 'internal/fs'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at evalmachine.<anonymous>:18:20
    at Object.<anonymous> (/home/travis/.nvm/versions/node/v7.4.0/lib/node_modules/bower/lib/node_modules/graceful-fs/fs.js:11:1)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)


The command "bower install --config.interactive=false --quiet" failed and exited with 1 during .

Your build has been stopped.
```

mccalluc:
Oh, yeah, that's the error I had seen when I tried the preview. Maybe it wasn't deployed across the whole pool yet. who knows. 

- "critcal" back on
- but I'll also make a PR to use the old version, now that it's available.

mccalluc:
[...]

PR #1817, opened 2017-06-21
Scottx611x/analysis manager updates

scottx611x:
@jkmarx @mccalluc Sorry for the long diff in analysis_manager.tests, but there wasn't a `test.py` file there and I thought it was due time.

> Note: The long diffs in `analysis_manager/utils.py` are just due to me moving some methods from analysis_manager around. I have no plans on touching them, but I've provided tests as best as I could.

- Provide feature parity with our old ways of launching analyses as I introduce the `Tool` based approach

- Provide test coverage for `analysis_manager` `utils` and `views` that were never tested beforehand

- Add schema-based approach for validating analysis configs

- Move some `analysis_manager` methods into places that make more sense. (Utility methods out of `.views` .etc)

PR #1818, opened 2017-06-21
Revert to previous image until we have a real fix for #1813

mccalluc:
https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch

If we don't actually need this, great. I don't really have a sense for myself of whether all our builds actually are failing.


scottx611x:
Includeded in https://github.com/refinery-platform/refinery-platform/pull/1817

PR #1820, opened 2017-06-21
Add depthNames data and replace types.

jkmarx:
- Variable to store depth names
- Display depth names instead of depth type
- Adjust language for consistency


scottx611x:
@jkmarx This looks good to me, but you might want to pull latest develop in to get the workaround for our travis builds.

Issue #1815, opened 2017-06-20
Add new facets which don't care about diambiguation?


mccalluc:
For the user files browser, we want to be able to facet just like we can within a single dataset. This is hard because each dataset adds a uuid so its fields are named differently. I would like something like `facet.field=*`, [but that was proposed, and shot down](https://issues.apache.org/jira/browse/SOLR-247). If I can't do that, then I guess I would need to iterate over the entire found set to collection facet field names, and then hit solr again for the facets, and even with that, I would still need to figure out what to do when the user clicked on a single facet that corresponded to multiple underneath.

Alternate proposal: We add new indexes to solr that will be consistent across all datasets: If a particular record uses a given field twice, that's fine: it doesn't need to be unique.

Thoughts? cc: @jkmarx @scottx611x

scottx611x:
Should we add an entirely new index? Or just a another field to the current `DataSetIndex`? 

I think adding a new `prepare` method to the index to normalize this data sounds like it would be sufficient, but I could not be understanding the problem fully.

mccalluc:
Fixed in #1824

PR #1824, opened 2017-06-22
Mccalluc/add fields to datasetmanager index

mccalluc:
Towards #1815: I want facets that will work across datasets. I'm not sure who has the most experience here, so listing everyone.

The test needs to be stronger, and after that the next step would be to use it in the UI.

### DIFF
--- a/refinery/data_set_manager/search_indexes.py
+++ b/refinery/data_set_manager/search_indexes.py
@@ -60,17 +60,14 @@ class NodeIndex(indexes.SearchIndex, indexes.Indexable):
             uuid += "_" + str(object.assay.id)
         # create dynamic fields for each attribute
         for annotation in annotations:
-            if annotation.attribute_subtype is None:
-                name = annotation.attribute_type
-            else:
-                name = annotation.attribute_subtype + "_" + \
-                       annotation.attribute_type
-            if annotation.attribute_value_unit is None:
-                value = annotation.attribute_value
-            else:
-                value = annotation.attribute_value + " " + \
-                        annotation.attribute_value_unit
-            # replace problematic characters in field names
+            name = annotation.attribute_type
+            if annotation.attribute_subtype is not None:
+                name = annotation.attribute_subtype + "_" + name
+
+            value = annotation.attribute_value
+            if annotation.attribute_value_unit is not None:
+                value += " " + annotation.attribute_value_unit
+
             name = re.sub(r'\W',
                           settings.REFINERY_SOLR_SPACE_DYNAMIC_FIELDS,
                           name)
--- a/refinery/data_set_manager/search_indexes.py
+++ b/refinery/data_set_manager/search_indexes.py
@@ -5,7 +5,7 @@ Created on Jul 2, 2012
 '''
 
 import logging
-import string
+import re
 
 from django.conf import settings
 from haystack import indexes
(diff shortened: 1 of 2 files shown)


mccalluc:
(Tests still pass, which indicates that we need better coverage here.)

refinery/data_set_manager/search_indexes.py
```diff
@@ -60,33 +60,20 @@ def prepare(self, object):
...
+            value = annotation.attribute_value
+            if annotation.attribute_value_unit is not None:
+                value += " " + annotation.attribute_value_unit
+
+            name = re.sub(r'\W',
+                          settings.REFINERY_SOLR_SPACE_DYNAMIC_FIELDS,
+                          name)
+
+            uniq_key = name + "_" + uuid + "_s"
+            generic_key = name + "_generic_s"
```
mccalluc: Adding "generic" fields to solr is the reason for this PR. Is this a reasonable approach?

refinery/data_set_manager/tests.py
```diff
@@ -1572,3 +1571,62 @@ def test_get_basic_node(self):
...
+class NodeIndexTests(APITestCase):
+
+    def setUp(self):
+        investigation = Investigation.objects.create()
+        study = Study.objects.create(investigation=investigation)
+        assay = Assay.objects.create(study=study)
+
+        test_file = StringIO()
+        test_file.write('Coffee is great.\n')
+        file_store_item = FileStoreItem.objects.create(
```
scottx611x: I've found that something in our codebase is still writing temp files to disk for FileStoreItems regardless of the class you use for the `datafile` param

We have a pre_delete signal for FileStoreItems that deletes these datafiles, so I've been doing this in my tests:

```
def tearDown(self):
    FileStoreItem.objects.all().delete()
```

refinery/data_set_manager/search_indexes.py
scottx611x: @mccalluc I wouldn't know myself without doing more research about solr. 
I think @flekschas  would have more of an idea than any other team member.

scottx611x:
@mccalluc Also regarding the travis test failures I found overriding the `maxDiff` attribute of TestCases to be helpful:
```
# Some members in assertions are truncated if they are too long, but we
# want to test them in their entirety
maxDiff = None
```

refinery/data_set_manager/tests.py
```diff
@@ -1,18 +1,20 @@
...
 from django.core.files.uploadedfile import (InMemoryUploadedFile,
                                             SimpleUploadedFile)
-
 from django.http import QueryDict
 from django.test import TestCase
-
 from rest_framework.test import APIClient, APIRequestFactory, APITestCase
 
-from .models import Assay, AttributeOrder, Study, Investigation, Node
+from core.models import DataSet, ExtendedGroup, InvestigationLink
```
jkmarx: code style guide: 
```Use explicit relative imports:
from __future__ import absolute_import

from .models import WaffleCone
from .forms import WaffleConeForm
from core.views import FoodMixin```

jkmarx: Which needs to be updated?

mccalluc: isort wants to put relative imports at the end, which doesn't seem bad to me, but I'll look into overriding that: #1830.

refinery/data_set_manager/search_indexes.py
mccalluc: Talked this over with Nils and he thinks its reasonable.

Issue #1712, opened 2017-05-02
Be able to Launch a `ToolDefinition: Workflow` with `tools` endpoint


scottx611x:
Steps needed:
``` python 
elif tool.get_tool_type() == ToolDefinition.WORKFLOW:
    workflow_tool.launch()
    # POST -> `/run` (analysis_manager.views.run)
    # CURRENTLY REQUIRES:
    # ->  workflow_uuid
    #  ->  study_uuid
    #  ->  node_set_uuid
    #  ->  node_group_uuid
    #  ->  node_relationship_uuid
    #  ->  custom_name
    # RETURNS: HttpResponse w analysis.uuid

    workflow_tool.analysis = Analysis.objects.get(
        uuid=HttpResponse["analysis_uuid"]
    )
    workflow_tool.save()

    #Then the run_analysis() task is called:
    # - Files are imported into Refinery
    # - Files are imported into Galaxy (think about the use
    #     case where we have many galaxy instances running in
    #     containers...)

    # - DATASET COLLECTIONS WILL BE CREATED HERE

    # - Export Galaxy results back to Refinery
    # - Attaching of outputs/derived files happens here:
    #      analysis.attach_outputs_dataset()
    #      analysis.attach_outputs_downloads()
```

General Notes:
- Workflow objects still should be created for backwards compatibility (workflow visualization .etc) ToolDefinition generation should create this.

scottx611x:
Related to #444

PR #1825, opened 2017-06-22
Scottx611x/workflow launching

scottx611x:
Closes https://github.com/refinery-platform/refinery-platform/issues/1712

- Allow for workflow based-Tools to be launched from the ui.
- @jkmarx @mccalluc A lot of the diff is due to moving methods into places that made more sense, and necessary updates to different tests.


refinery/tool_manager/utils.py
```diff
@@ -466,3 +455,84 @@ def parse_file_relationship_nesting(nested_structure, nesting_dict=None,
         parse_file_relationship_nesting(
             item, nesting_dict=nesting_dict, nesting_level=nesting_level
         )
+
+
+def validate_tool_annotation(annotation_dictionary):
```
scottx611x: validation seems very similar across these methods

Issue #1826, opened 2017-06-22
Nodes, and Attributes, and Annotations (oh my)


mccalluc:
More docs around these might help, or there might be things to clean up. Here's what I understand, and what seems weird:

In `single_file_column_parser:258` it looks like the data from extra columns is stored:
```
                    Attribute.objects.create(
                        node=sample_node, type=Attribute.CHARACTERISTICS,
                        subtype=self.headers[column_index].strip().lower(),
                        value=row[column_index].strip()
                    )
```
Downstream, though, in `search_indexes.py:56`, where I think the data is pulled out, it’s talking about “annotations”:
```
annotations = AnnotatedNode.objects.filter(node=object)
```

What is the relationship between these? Looking at the code:
- `Node` has `is_annotation`
- `AnnotatedNode` has `is_annotation` as well. It also has FKs for `node` and `attribute`, and several `node_*` and `attribute_*` properties, which might be denormalized for some reason, but I'm not sure that's what's going on.
- `Attribute` has `node`, and `is_attribute`, which seems tautological?
- `AttributeDefinition` and `AttributeOrder` are important for non ISAtab files?

mccalluc:
- Confirmed that `AnnotatedNode` is all about denormalization: it's a bridge to the index.

mccalluc:
@scottx611x : follow-up question: core/searh_indexes.py:
```python
            studies = investigation.study_set.all()
            for study in studies:
                assays = study.assay_set.all()
                for assay in assays:
                    nodes = list(AnnotatedNode.objects.filter(
                        study__uuid=study.uuid,
                        assay__uuid=assay.uuid,
                        node_type__in=Node.FILES
                    ).distinct(
                        'node_name',
                        'attribute_type',
                        'attribute_subtype',
                        'attribute_value',
                        'attribute_value_unit'
                    ))
            ...
            data['text'] = template.render(
                Context({
                    'object': data_set,
                    'nodes': nodes
                })
            )
```
To me, it looks like nodes is being reset each time through the loop?
- If that's correct, and the desired behavior, could we just set it once taking the last assay and the last study?
- If that's not the desired behavior, could we get rid of it, because it doesn't seem to have been bothering us?

mccalluc:
closing. Questions found some answers.

PR #1827, opened 2017-06-23
Mccalluc/cleanup data set manager models

mccalluc:
I would like to understand better how the data_set_manager models work. As I think I understand better how the pieces work, I'll clean up and add tests in this branch. Trying to answer the questions in #1826.


mccalluc:
@ngehlenborg or @scottx611x : The primary motivation was learning, but some of the tweaks here may be useful. I think I'll set this aside right now rather than adding to it.

mccalluc:
I think this is good, but since it's a clean-up which doesn't introduce anything we need, I'll hold off on merging until start of the next release.

mccalluc:
Single failure is `test_visualization_container_launch_IGV`, which we've seen before. No changes besides merges since earlier approval by scott.

PR #1831, opened 2017-06-23
Sort django above other third party, and local above first part

### LINKED ISSUE #1830: finalize isort config
In the wiki, we state that the order should be:
>
    standard library imports
    imports from core Django
    imports from third party libraries and apps
    imports from the apps that you created as part of your Django project

[isort's default](https://github.com/timothycrosley/isort) is:
>
    Future
    Python Standard Library
    Third Party
    Current Python Project
    Explicitly Local (. before import, as in: from . import x)
    Custom Separate Sections (Defined by forced_separate list in configuration file)
    Custom Sections (Defined by sections list in configuration file)

We currently override this in `.isort.cfg` to get closer to the behavior we want.
- Listing known first-party imports
- and otherwise defaulting to thirdparty.
(I'm actually not clear on how isort should normally distinguish current project imports from third party, but this seems to work for us.)

There remain a few differences between isort and our wiki guidelines:
- isort is separating explicit relative imports
- isort is not treating django specially

I think the treatment of relative imports is actually not a bad thing, but I'll look at segregating django. Thoughts?

mccalluc:
Fix #1830. With these settings, imports will come out looking like:
```
from __future__ import foo

import json
import re

from django.test import TestCase

from .models import Assay, AttributeOrder, Investigation, Node, Study
from .views import Assays, AssaysAttributes

from core.models import DataSet, ExtendedGroup, InvestigationLink
from file_store.models import FileStoreItem
```
(Besides running as part of the pre-commit, you can also just use it by itself on the command-line to see how it works. `-d` will dump rather than modifying in-place.)

@jkmarx / @hackdna : Does this do what you want? Or would you prefer a different behavior? I don't think it's a priority to apply this recursively on all files, but if the behavior file-by-file isn't what you want, I'd like to get that clarified now.


hackdna:
Looks good though I'd put the local relative imports (e.g., `.models`) last. We should update the [wiki](https://github.com/refinery-platform/refinery-platform/wiki/Coding-Style-Guide#python) too.
It's not a bad idea to apply these rules to all files for better consistency and readability but we might run into issues with circular imports.

jkmarx:
Agree w/ Ilya. Most of the file imports were updated according to the wiki above a while ago with local relative imports last.

mccalluc:
Updated the wiki to put local imports last. @jkmarx : not sure what "main conversation page" means here. Let me know if you want anything else here before merge?

(I'll make a separate branch to apply this across the board: I'd be reluctant to fit it in in this release.)

(Travis hit the NPM problem again: restarting.)

mccalluc:
Just incorporated the lessons learned from #1844, so as individual imports are reordered, we're moving in the right direction, and when we do make the bulk change, more files will already be in the right shape.

PR #1834, opened 2017-06-26
Jkmarx/missing nav bar bug

jkmarx:
- Minor bug
- Adjust css

### DIFF
--- a/refinery/ui/source/styles/file-browser.less
+++ b/refinery/ui/source/styles/file-browser.less
@@ -267,6 +267,10 @@
   }
 }
 
+#node-selection-popover {
+  width: 400px;
+}
+
 #param-help-popover {
   p.param-field {
     text-transform: none;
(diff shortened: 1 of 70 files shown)


refinery/ui/source/styles/file-browser.less
```diff
@@ -267,8 +267,10 @@
   }
 }
 
-#node-selection-popover {
-  width: 400px;
+#assay-files-table {
```
scottx611x: I think you can just do:
```
#assay-files-table.popover {
 width: 400px !important;
}
```

But thats just syntax preference, not sure what we prefer

refinery/ui/source/styles/file-browser.less
```diff
@@ -347,3 +349,9 @@ ul.plain {
   list-style: none;
   padding-left: 0;
 }
+
+.refinery-subheader {
```
scottx611x: same syntax here

PR #1838, opened 2017-06-26
Mccalluc/user files facets

mccalluc:
This brings the client side code to a place where I could hit the API, and fixes a bug that had been in the solr indexing. (Credit the fix to Scott.)

I've spent some time looking at the code around AttributeOrder. I still think it would be possible to do it that way, but it's looking a lot messier, and likely to break something else, than just setting up a new solr core that will only have the facets I need. I'll try to start on that in a separate branch which doesn't need to build from this one.

### DIFF
--- a/refinery/data_set_manager/search_indexes.py
+++ b/refinery/data_set_manager/search_indexes.py
@@ -124,14 +124,11 @@ class NodeIndex(indexes.SearchIndex, indexes.Indexable):
         except(FileStoreItem.DoesNotExist,
                FileStoreItem.MultipleObjectsReturned) as e:
             logger.error("Couldn't properly fetch FileStoreItem: %s", e)
+            file_store_item = None
+        if file_store_item:
+            data[NodeIndex.FILETYPE_PREFIX + "_" + uuid + "_s"] =\
+                file_store_item.get_filetype()
         else:
-            if file_store_item:
-                data[NodeIndex.FILETYPE_PREFIX + "_" + uuid + "_s"] =\
-                    file_store_item.get_filetype()
-            else:
-                logger.warning(
-                    "Unable to get file store item " + str(object.file_uuid) +
-                    ". No file type available.")
-                data[NodeIndex.FILETYPE_PREFIX + "_" + uuid + "_s"] = ""
+            data[NodeIndex.FILETYPE_PREFIX + "_" + uuid + "_s"] = ""
 
-            return data
+        return data
--- a/refinery/ui/source/js/user-file-browser/ctrls/user-file-browser-filters-ctrl.js
+++ b/refinery/ui/source/js/user-file-browser/ctrls/user-file-browser-filters-ctrl.js
@@ -29,5 +29,11 @@
     };
 
     vm.hidden = {};
+
+    vm.filters = {};
+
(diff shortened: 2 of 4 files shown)


refinery/ui/source/js/user-file-browser/ctrls/user-file-browser-filters-ctrl.js
```diff
@@ -23,6 +23,20 @@
       $log.error('/user/files/ request failed');
       promise.reject();
     });
+
+    vm.togglePanel = function (attribute) {
+      vm.hidden[attribute] = ! vm.hidden[attribute];
+    };
+
+    vm.hidden = {};
```
jkmarx: vm variables up top

refinery/data_set_manager/search_indexes.py
```diff
@@ -124,14 +124,11 @@ def prepare(self, object):
...
+            file_store_item = None
+        if file_store_item:
+            data[NodeIndex.FILETYPE_PREFIX + "_" + uuid + "_s"] =\
+                file_store_item.get_filetype()
         else:
-            if file_store_item:
-                data[NodeIndex.FILETYPE_PREFIX + "_" + uuid + "_s"] =\
-                    file_store_item.get_filetype()
-            else:
-                logger.warning(
```
scottx611x: Do we still want a logger message somewhere?

refinery/ui/source/js/user-file-browser/services/factory.js
```diff
@@ -124,9 +124,11 @@
             });
           }
         });
-        filters[filterName] = {
-          facetObj: facetCounts
-        };
+        if (facetCounts.length > 0) {
```
jkmarx: Design choice: once a user starts filtering, they would not see the available fields.  Also, I believe solr has a min count filter, so you can add it to the query instead of filtering client side. https://wiki.apache.org/solr/SimpleFacetParameters#facet.mincount
short-zero

refinery/data_set_manager/search_indexes.py
mccalluc: We'll only get to the `else` if we were in the `except`:
```python
except(FileStoreItem.DoesNotExist,
            logger.error("Couldn't properly fetch FileStoreItem: %s", e)
            file_store_item = None
        if file_store_item:
            data[NodeIndex.FILETYPE_PREFIX + "_" + uuid + "_s"] =\
                file_store_item.get_filetype()
        else:
            data[NodeIndex.FILETYPE_PREFIX + "_" + uuid + "_s"] = ""
```

refinery/ui/source/js/user-file-browser/services/factory.js
mccalluc: Thanks! I'll revisit this when the front and back ends are connected.

refinery/ui/source/js/user-file-browser/ctrls/user-file-browser-filters-ctrl.js
mccalluc: fixed. Thanks.

mccalluc:
(Both Travis runs failed on test_visualization_container_launch_IGV, which is unrelated.)

PR #1840, opened 2017-06-26
/api/v2/user/files/ now returns the facets I need

mccalluc:
Rather than going too deeply into the existing logic, this provides a work-around. API output looks like:
```
{
    "facet_field_counts": {
        "technology_Characteristics_generic_s": [
            {
                "count": 2,
                "name": "ChIP-seq"
            }
        ],
        "antibody_Characteristics_generic_s": [],
        "organism_Characteristics_generic_s": [
            {
                "count": 32,
                "name": "Homo sapiens"
            },
            {
                "count": 2,
                "name": "Mouse"
            }
        ],
        "organism_Factor_Value_generic_s": [],
        "genotype_Characteristics_generic_s": [
            {
                "count": 1,
                "name": "XYZ"
            }
        ],
        "antibody_Factor_Value_generic_s": [
            {
                "count": 25,
                "name": "GATA6 + none + CDX2 + HNF4A"
            },
            {
                "count": 2,
                "name": "CDX2"
            },
...
```


refinery/data_set_manager/utils.py
```diff
@@ -629,7 +632,14 @@ def _generate_solr_params(params, assay_uuids):
         return None
     solr_params = 'fq=assay_uuid:({})'.format(' OR '.join(assay_uuids))
 
-    if facet_field:
+    if facets_from_config:
+        # Twice as many facets as necessary, but easier than the alternative.
```
scottx611x: This might take a second for someone to figure out. How about:
```
template = '&facet.field={0}_Characteristics_generic_s' + \
        '&facet.field={0}_Factor_Value_generic_s'
solr_params += ''.join([template.format(s) for s in settings.USER_FILES_FACETS.split(",")])
```

mccalluc: @scottx611x : Thanks! With that change in place, is this approved?

mccalluc:
@scottx611x : Could you re-review this? My sloppiness with `format` wasn't caught by any tests, and when I got into that, I realized the logic we had for pulling out the assays connected to a user was wrong.

PR #1841, opened 2017-06-27
Scottx611x/workflow launching updates

scottx611x:
- Create and associate `Workflow` objects upon creation of `Workflow-based ToolDefinitions`
- Pull Docker images for `Visualization-based ToolDefinitions` upon `ToolDefinition` Creation, rather than when a `Tool` is launched from them.
- Update mock Galaxy response


refinery/tool_manager/models.py
```diff
@@ -202,32 +197,14 @@ def get_extra_directories(self):
...
-                workflow_engine=self.workflow_engine
-            )
-        except(Workflow.DoesNotExist,
-               Workflow.MultipleObjectsReturned) as e:
-            raise RuntimeError("Couldn't fetch Workflow: {}".format(e))
-
 
 @receiver(pre_delete, sender=ToolDefinition)
-def delete_parameters_and_output_files(sender, instance, *args, **kwargs):
+def remove_associated_objects(sender, instance, *args, **kwargs):
```
mccalluc: Is there a consistent practice around `delete_` vs `remove_`? Just noticing that the next method down is still `delete_*`.

refinery/tool_manager/models.py
```diff
@@ -356,13 +341,19 @@ def _launch_workflow(self):
             pointing to the Analysis' status page
         :raises: RuntimeError
         """
+        if self.get_tool_type() != ToolDefinition.WORKFLOW:
+            raise NotImplementedError(
+                "Tool must be of type: {} to launch a workflow".format(
+                    ToolDefinition.WORKFLOW
+                )
```
mccalluc: Judgement call, but you could just say "Not a workflow" while hardly inconveniencing the user (and saving 4 lines)... but if you did hit this in the log, you'd probably also want the ID and the tool_type? Something to think about in general, but not worth worrying about here.

refinery/tool_manager/tests.py
```diff
@@ -869,6 +868,14 @@ def test_get_extra_directories_workflow_tool_def(self):
         with self.assertRaises(NotImplementedError):
             self.td.get_extra_directories()
 
+    def test_related_workflow_inactive_after_deletion(self):
+        self.create_workflow_tool_definition()
+        self.assertTrue(self.td.workflow.is_active)
+        self.td.delete()
+        self.assertFalse(
+            Workflow.objects.all()[0].is_active
+        )
```
mccalluc: Would you like to make an assertion about `self.td` here? It's not immediately clear what the relationship to `Workflow.objects.all()[0]` would be...

refinery/tool_manager/utils.py
```diff
@@ -31,7 +33,13 @@
...
 # Allow JSON Schema to find the JSON pointers we define in our schemas
 JSON_SCHEMA_FILE_RESOLVER = RefResolver(
-    "{}{}{}".format('file://', os.path.abspath("tool_manager/schemas"), '/'),
+    "{}{}{}".format(
+        'file://',
+        os.path.join(
+            settings.BASE_DIR,
+            "refinery/tool_manager/schemas"
+        ),
+        '/'),
```
mccalluc: Not worth worrying about, but the static strings could go in the template, and that would make the intention more clear at the get-go: `'file://{}/''.format(...)`

refinery/tool_manager/utils.py
```diff
@@ -128,27 +136,53 @@ def create_tool_definition(annotation_data):
...
+        #  pull down its docker image
+        try:
+            image_name, version = tool_definition.image_name.split(":")
+        except ValueError:
+            logger.debug(
+                "`image_name` has no specified version, and the `latest` "
+                "version of %s will be pulled",
+                tool_definition.image_name
+            )
+            DockerClientWrapper().pull(tool_definition.image_name)
```
mccalluc: Oh: I would not continue here: I don't think the default behavior is ever helpful, and debug level logging may very well not be turned on .

refinery/tool_manager/tests.py
scottx611x: Not particularly interested in `self.td` other than what happens after it gets deleted.
Since `ToolDefinitions` now have a foreign key to `Workflows` I want to assert that Workflows are set to an inactive state after a ToolDefinition is removed, I just can't access the Workflow object through the definition post-deletion, hence the ORM stuff

PR #1844, opened 2017-06-27
Mccalluc/isort everywhere

mccalluc:
Extending the work from #1831.

This is the result of `isort -rc .`, but
- excluding migrations
- reverting a comment deletion in `refinery/config/settings/aws.py` (I think it interprets a comment in the imports as a section header, and we do not have those turned on.)

Questions:
- Some packages we often put in the django section, and perhaps they should be kept there? `mock`, `mockcache`, `django_extensions`
- Are you happy with the handling of `from` vs `import`?
- Should migrations be explicitly excluded here? Or should they be re-ordered, too? Or just ignore them?
- If folks preferred a different style for the long lists of imports, that could be tweaked.

I don't have strong opinions about any of these. I do think we should wait till the start of the next release before merging this.


refinery/analysis_manager/tasks.py
```diff
@@ -6,20 +6,20 @@
 import logging
 import urlparse
 
-from bioblend import galaxy
 import celery
+import requests
```
hackdna: Why is `requests` before `celery.result`?

refinery/analysis_manager/tasks.py
```diff
@@ -6,20 +6,20 @@
...
+from bioblend import galaxy
```
hackdna: Why is `bioblend` after `celery`?

refinery/analysis_manager/utils.py
```diff
@@ -4,11 +4,12 @@
 import os
 import re
 
-import requests
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.http import HttpResponseServerError
 from django.utils import timezone
+
+import requests
```
hackdna: Why is `requests` before `jsonschema`?

refinery/analysis_manager/tests.py
```diff
@@ -1,10 +1,11 @@
 import json
 import uuid
 
-import mock
 from django.contrib.auth.models import User
 from django.http import HttpResponseBadRequest, HttpResponseNotAllowed
 from django.test import RequestFactory, TestCase
+
+import mock
```
hackdna: Why is `mock` before `guardian`?

hackdna:
The comment in the import section of `aws.py` is superfluous and can be removed. However, there appear to be multiple issues with ordering. I've commented on the first few that I found. Third-party packages like `mock` should be grouped with other third-party packages. Migration are generated automatically so should probably be ignored (they often don't even confirm to PEP8).

mccalluc:
> However, there appear to be multiple issues with ordering. I've commented on the first few that I found.

The default behavior is to separate the different types of imports; I overridden that default, to both kinds of imports are sorted together.

> Third-party packages like mock should be grouped with other third-party packages.

So no action needed there.

> Migration are generated automatically so should probably be ignored (they often don't even confirm to PEP8).

Add an ignore to cover all migrations

mccalluc:
Close in favor of #1927

PR #1850, opened 2017-06-28
Mccalluc/refactor search indexes

mccalluc:
I think this might be easier to read? There was just a lot of unnecessary repetition before. It's subjective, and I could be wrong.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,9 @@ before_script:
 script:
   - set -e # Any error will cause travis to exit early and report a failure.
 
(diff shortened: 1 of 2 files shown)


mccalluc:
(Pre-commit and flake8 in Travis aren't using the same rules: #1851)

mccalluc:
Ugh: Add one ignore, and instead of getting quieter, now I have dozens of style errors being reported. Maybe shelve this for the moment and get #1851 under control first?

mccalluc:
PR failure is known IGV timeout.

PR #1854, opened 2017-06-28
Scottx611x/run analysis refactoring

scottx611x:
- Refactor monolithic `run_analysis` for easier testing, and extensibility moving forward w/ the implementation of new `Tool`-based `Analyses`


refinery/analysis_manager/utils.py
```diff
@@ -61,7 +61,7 @@ def create_nodeset_analysis(validated_analysis_config):
     :return: an Analysis instance
     :raises: RuntimeError
     """
-    custom_name = validated_analysis_config["custom_name"]
+    custom_name = validated_analysis_config.get("custom_name")
```
mccalluc: In general, if we can avoid having None, it's a good thing. Here, we only use `custom_name` to set `temp_name`:
```python
if not custom_name:
         temp_name = current_workflow.name + " " + get_aware_local_time() \
             .strftime("%Y-%m-%d @ %H:%M:%S")
     else:
temp_name = custom_name
```

Could you just set `temp_name` directly, and use the optional second arg instead of the `if`? (One difference would be the handling of empty strings... not sure if that's a case we care about?)

refinery/core/models.py
```diff
@@ -1285,6 +1286,17 @@ def running(self):
...
+    def get_tool(self):
+        try:
+            return tool_manager.models.Tool.objects.get(analysis=self)
+        except(tool_manager.models.Tool.DoesNotExist,
+               tool_manager.models.Tool.MultipleObjectsReturned) as e:
+            logger.debug(
+                "Couldn't fetch Tool for Analysis with UUID: %s %s",
+                self.uuid,
+                e
+            )
```
mccalluc: The caller does something appropriate with `None`?

refinery/analysis_manager/views.py
```diff
@@ -189,7 +189,7 @@ def run(request):
     analysis_status.save()
 
     # call function via analysis_manager
-    run_analysis.delay(analysis.uuid)
+    AnalysisRunner(analysis.uuid)
```
mccalluc: The new way puts the logic in a constructor, but we don't actually have any need for the resulting object? Do we ever?

refinery/analysis_manager/tasks.py
```diff
@@ -322,22 +407,47 @@ def start_galaxy_analysis(analysis_uuid):
...
-        connection.workflows.run_workflow(
-            workflow_id=analysis.workflow_galaxy_id,
-            dataset_map=ds_map,
-            history_id=analysis.history_id
-        )
-    except galaxy.client.ConnectionError as exc:
-        error_msg += "error running Galaxy workflow for analysis '%s': %s"
-        logger.error(error_msg, analysis.name, exc.message)
-        start_galaxy_analysis.update_state(state=celery.states.FAILURE)
+    tool = analysis.get_tool()
```
scottx611x: This doesn't need to be in PR `invoke_workflow` is in later bioblend

refinery/analysis_manager/tasks.py
```diff
@@ -349,18 +459,35 @@ def get_galaxy_download_tasks(analysis):
...
-            analysis.history_id)
-    except galaxy.client.ConnectionError as exc:
-        error_msg = "Error downloading Galaxy history files for analysis " \
-                    "'%s': %s"
-        logger.error(error_msg, analysis.name, exc.message)
-        analysis.set_status(Analysis.FAILURE_STATUS, error_msg)
-        analysis.galaxy_cleanup()
-        return task_list
+
+    tool = analysis.get_tool()
```
scottx611x: doesn't need to be in this PR

refinery/core/models.py
scottx611x: get_tool() won't be necessary

refinery/tool_manager/models.py
```diff
@@ -278,6 +278,11 @@ def __str__(self):
             self.uuid
         )
 
+    def get_input_file_uuid_list(self):
```
scottx611x: Not in this PR

refinery/analysis_manager/tasks.py
```diff
@@ -46,64 +48,240 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
         analysis.set_status(Analysis.FAILURE_STATUS, error_msg)
 
 
-@task(base=AnalysisHandlerTask, max_retries=None)
-def run_analysis(analysis_uuid):
-    """Manage analysis execution"""
-    RETRY_INTERVAL = 5  # seconds
-
+def _get_analysis(analysis_uuid):
```
hackdna: Name is misleading because method not only retrieves an `Analysis` object but can also change the state of the `run_analysis` task.

refinery/analysis_manager/tasks.py
```diff
@@ -46,64 +48,240 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
-                     analysis_uuid, exc)
+                     analysis_uuid, e)
         run_analysis.update_state(state=celery.states.FAILURE)
         return
 
-    # if cancelled by user
-    if analysis.failed():
-        return
 
+def _get_analysis_status(analysis_uuid):
```
hackdna: Name is misleading because method not only retrieves an `AnalysisStatus` object but can also change the state of the `run_analysis` task.

refinery/analysis_manager/tasks.py
scottx611x: @hackdna Do you have any proposal for name changes?

refinery/analysis_manager/tasks.py
scottx611x: @hackdna Do you have any proposal for name changes?

refinery/analysis_manager/tasks.py
hackdna: I'd remove the logic for updating the task state change or even simply get rid of the function completely because it is used only once.

refinery/analysis_manager/tasks.py
hackdna: I'd remove the logic for updating the task state change or even simply get rid of this function completely because it is used only once.

scottx611x: Both `_get_analysis()` and `_get_analysis_status()` are used many times each throughout the methods that `run_analysis()` encapsulates.

Considering that they are both specific to this module, I think its okay to keep the logic to update the task state. 

As for naming I could append `_or_fail`, but that seems odd to me. I don't have any other good names coming of the top of my head.

hackdna: Up to you but keep in mind that this breaks [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) which limits code reuse, makes code less readable, testable and maintainable. Difficulty of naming is just a symptom that simply points to this underlying problem.

PR #1857, opened 2017-06-28
Mccalluc/make the facets do something

mccalluc:
Building on the earlier #1855: I now have a hard coded fq in user-file.js ... Next step is to understand how to connect this to the facets panel.

(Added a couple TODOs in existing code where things didn't seem right.)

### DIFF
--- a/refinery/ui/source/js/user-file-browser/services/user-file.js
+++ b/refinery/ui/source/js/user-file-browser/services/user-file.js
@@ -15,6 +15,7 @@
         query: {
           method: 'GET',
           params: {
+            limit: 100, // Default is 100,000. Immutability make it hard in python.
             fq: function () {
               var filters = Object.keys(userFileFiltersService).map(function (key) {
                 var values = userFileFiltersService[key];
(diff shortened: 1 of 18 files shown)


refinery/ui/source/js/user-file-browser/services/user-file.js
```diff
@@ -4,15 +4,24 @@
     .module('refineryApp')
     .factory('userFileService', userFileService);
 
-  userFileService.$inject = ['$resource', 'settings'];
+  // TODO: userFileBrowserFiltersCtrl.filters is what I need, but maybe I need a new service?
```
mccalluc: @jkmarx : This is the point I'd like to fix when I come back to this tomorrow: userFileBrowserFiltersCtrl.filters has the information I need, but my guess is that I need to pull that out and put it in a service so I can access it here? (A controllers shouldn't be a dependency here?)

Let me know if I'm not thinking about this correctly?

jkmarx: Correct. A ctrl should not be a dependency here, but you can set service data from a ctrl.

mccalluc:
On Friday, I did not cleanly separate the data from the rest of the grid configuration. I would like to come back to that, but I thought it was good to get this much in, with Scott's go-ahead.

PR #1859, opened 2017-06-29
Fix concat bug when offset is 0.

jkmarx:
- Fix bug when cont scroll will concat below 0.

### DIFF
--- a/refinery/ui/source/js/file-browser/services/factory.js
+++ b/refinery/ui/source/js/file-browser/services/factory.js
@@ -197,7 +197,7 @@
         assayFilesTotalItems.count = response.nodes_count;
 
         // Not concat data when under minimun file order, replace assay files
-        if (assayFilesTotalItems.count < maxFileRequest && params.offset === 0) {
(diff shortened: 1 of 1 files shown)


refinery/ui/source/js/file-browser/services/factory.js
```diff
@@ -197,7 +197,7 @@
         assayFilesTotalItems.count = response.nodes_count;
 
         // Not concat data when under minimun file order, replace assay files
-        if (assayFilesTotalItems.count < maxFileRequest && params.offset === 0) {
+        if (assayFilesTotalItems.count < maxFileRequest || params.offset === 0) {
```
mccalluc: This makes sense, but I'm not sure I understand when this branch would be taken: If the total files is less the the max files per request, this should only be hit on the initial page load, and there won't be any loading-on-scroll to worry about?

jkmarx: Yes or you're at the top of list.

PR #1863, opened 2017-07-05
Masking errors not helpful

mccalluc:
Masking the error like this made it harder for me to debug the root issue. AJAX response was 200, but body could not be parsed.


scottx611x:
@mccalluc This looks good to me. Masking a non-JSON solr response by returning a string seemed troublesome before.

I think that if solr isn't returning JSON, then something really odd is going on and letting an admin get an email with the traceback in production is okay.

PR #1866, opened 2017-07-06
Jkmarx/data set2 css update

jkmarx:
- Adjust css


scottx611x:
Push build failure just to errant igv test

PR #1875, opened 2017-07-06
I think we were only indexing the values we got the last time through

mccalluc:
Follow-up to https://github.com/refinery-platform/refinery-platform/issues/1826#issuecomment-313508341

If this is in fact a fix, should add a test that would fail in the old state.


mccalluc:
@scottx611x : let's see if the tests pass, but I believe this is a fix.

refinery/core/search_indexes.py
```diff
@@ -142,7 +141,7 @@ def prepare(self, data_set):
             for study in studies:
                 assays = study.assay_set.all()
                 for assay in assays:
-                    nodes = list(AnnotatedNode.objects.filter(
+                    nodes += list(AnnotatedNode.objects.filter(
```
mccalluc: This is the actual bug: fixing this motivates everything else.

refinery/core/tests.py
```diff
@@ -2413,6 +2415,16 @@ def setUp(self):
         self.good_dataset = create_dataset_with_necessary_models()
         self.bad_dataset = DataSet.objects.create()
 
+    def test_prepare(self):
+        data = self.dataset_index.prepare(self.good_dataset)
+        self.assertRegexpMatches(
+            data['text'],
+            re.compile(
+                r'AnnotatedNode-\d.*AnnotatedNode-\d',
```
mccalluc: This depends on the names the factory fills in as a default. Widely separated dependencies on the same string aren't reassuring.

refinery/factory_boy/utils.py
```diff
@@ -66,17 +70,36 @@ def create_dataset_with_necessary_models():
...
+        description="This is a great DataSet"
+    )
+
+    for i in xrange(2):
+        assay_uuid = str(uuid_builtin.uuid4())
+        assay = AssayFactory(
+            uuid=assay_uuid,
+            study=study
+        )
+        attribute = AttributeFactory()
```
scottx611x: @mccalluc I think `Attribute` creation has to follow `Node` creation here since it has node as a foreignKey

mccalluc: @scottx611x : huh, but right below it looks like like NodeFactory needs an attribute? This is the chicken-and-egg problem I was stuck on.

scottx611x: I think that block in the for loop should look something like this
```
node = NodeFactory(
    study=study
)
AttributeFactory(
    node=node
)
AnnotatedNodeFactory(
    study=study,
    assay=assay,
    node=node,
    node_name='AnnotatedNode-{}'.format(i),
    node_type=node.type
)
```

PR #1887, opened 2017-07-10
Annotated node explosion with tests

scottx611x:
- Add transaction mgmt. to ensure that when  the "Annotated Node Explosion" is hit that the current db transaction is nullified
- Add tests to ensure this behavior


refinery/data_set_manager/tests.py
```diff
@@ -1719,3 +1720,36 @@ def test_prepare(self):
                           'type': u'',
                           'uuid': self.node_uuid,
                           'workflow_output': None})
+
+
+class AnnotatedNodeExplosionTestCase(TestCase):
+    def setUp(self):
+        self.username = 'coffee_lover'
```
flekschas: ☕️

flekschas:
I ☕️ this pull request. Looks all fine to me

PR #1897, opened 2017-07-11
Scottx611x/analysis custom name bugfix

### LINKED ISSUE #1893: Naming analyses doesn't seem to do anything?
Giving a non-default names to a new analysis  doesn't seem to do anything, as far as I can tell: Were would I expect to see it be used?

scottx611x:
Fixes #1893
- Bugfix for user-specified Analysis names not being assigned
- Add test coverage to ensure that user-specified names are properly set

### DIFF
--- a/refinery/analysis_manager/utils.py
+++ b/refinery/analysis_manager/utils.py
@@ -440,10 +440,9 @@ def _fetch_node_set(node_set_uuid):
 
 def _fetch_solr_uuids(nodeset_instance):
     """
-    Fetches a NodeSet instance from a given UUID
-    :param node_set_uuid: UUID String
-    :return: <NodeSet>
-    :raises: RuntimeError
+    Fetches solr_uuids from a given NodeSet instance
+    :param nodeset_instance: <NodeSet> instance
+    :return: list of UUIDs corresponding to Nodes indexed in Solr
     """
     curr_node_dict = json.loads(nodeset_instance.solr_query_components)
     return get_solr_results(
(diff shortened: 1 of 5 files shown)


refinery/analysis_manager/utils.py
```diff
@@ -194,19 +163,17 @@ def create_noderelationship_analysis(validated_analysis_config):
 
     # ANALYSIS MODEL
     # How to create a simple analysis object
-    if not custom_name:
-        temp_name = "{} {}".format(
+    if not name:
+        name = "{} {}".format(
```
mccalluc: Is there logic that should be shared between here and line 81? (This feels like copy-and-paste which preceded this commit, so not a high priority to fix.)

refinery/analysis_manager/utils.py
```diff
@@ -413,3 +380,75 @@ def validate_analysis_config(analysis_config):
...
+        )
+
+
+def _fetch_solr_uuids(nodeset_instance):
+    """
+    Fetches a NodeSet instance from a given UUID
+    :param node_set_uuid: UUID String
+    :return: <NodeSet>
+    :raises: RuntimeError
+    """
```
mccalluc: Docs seem to be copy and paste, but don't match this method's signature?

scottx611x: Nice find, thanks :)

refinery/analysis_manager/utils.py
scottx611x: Yes definitely, I can refactor

PR #1919, opened 2017-07-13
Add svgxuse dependency to allow for external svgs to properly be loaded

### LINKED ISSUE #1905: Satori icon loading could be enhanced
* Specific code commit: 2623df259c0eeed61a4e861ca1c1ac5e74302d3f
* Version of the web browser and OS: MacOS Chrome
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): Test Production

None of the icons for Satori are loading: Neither the hints in the header, nor elements in the real gui. In the js console:
```
d3.min.js:4 Unsafe attempt to load URL 
https://refinery-static.s3.amazonaws.com/vendor/d3-list-graph/dist/icons.svg 
from frame with URL https://test.stemcellcommons.org/#/exploration?context=OWL:Thing&branchId=0. 
Domains, protocols and ports must match.
```

scottx611x:
Addresses issue here: #1905 but is probably not a great solution. See: https://github.com/refinery-platform/refinery-platform/pull/1919#issuecomment-315373543

Will leave #1905 open but merge this as a workaround until a better solution can be found.

### DIFF
--- a/refinery/templates/base.html
+++ b/refinery/templates/base.html
@@ -323,7 +323,7 @@
   <script src="{% static "js/data-set-import.js" %}"></script>
 
   <!-- Fetch external SVGs referenced in <use> elements when the browser fails to do so -->
-  <script defer src="{% static "vendor/svgxuse/svgxuse.js" %}"></script>
+  <script defer src="{% static "vendor/svgxuse/svgxuse.min.js" %}"></script>
 
   {% block script %}
   {% endblock %}
--- a/refinery/templates/base.html
+++ b/refinery/templates/base.html
@@ -322,6 +322,9 @@
   <script src="{% static "vendor/jquery-file-upload/js/jquery.fileupload-angular.js" %}"></script>
   <script src="{% static "js/data-set-import.js" %}"></script>
 
+  <!-- Fetch external SVGs referenced in <use> elements when the browser fails to do so -->
+  <script defer src="{% static "vendor/svgxuse/svgxuse.js" %}"></script>
+
   {% block script %}
   {% endblock %}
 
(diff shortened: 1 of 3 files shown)


refinery/templates/base.html
```diff
@@ -322,6 +322,9 @@
   <script src="{% static "vendor/jquery-file-upload/js/jquery.fileupload-angular.js" %}"></script>
   <script src="{% static "js/data-set-import.js" %}"></script>
 
+  <!-- Fetch external SVGs referenced in <use> elements when the browser fails to do so -->
+  <script defer src="{% static "vendor/svgxuse/svgxuse.js" %}"></script>
```
flekschas: Does this work? It looks like you want to load the lib async (i.e., `defer`) after other JS stuff. What happens if the icons are suppose to be loaded before the lib is loaded?

refinery/ui/bower.json
```diff
@@ -46,6 +46,7 @@
     "ng-webworker": "0.2.3",
     "spark-md5": "2.0.2",
     "spectrum": "1.7.0",
+    "svgxuse": "1.2.4",
     "tipsy": "0.1.7",
```
flekschas: 🍾🥂

refinery/templates/base.html
scottx611x: It works (locally at least) the `defer` was in their usage docs so I utilized as well.

flekschas: Yeah locally it might not be an issue because everything loads fast :) The question is wether svgxuse needs to be loaded before `<use>` are parsed.

Also, is "vendor/svgxuse/svgxuse.js" the minified version?

scottx611x:
FWIW this works on AWS deployed instances as well, and allows for the svgs to load in Chrome, FF and Safari on macOS. 

The console errors from #1905 still persist. 
Although, these errors have been around for a while (they exist on the beta site currently) so I'm not sure how to proceed.

PR #1934, opened 2017-07-20
"++" is not an operator in Python. 

mccalluc:
@flekschas : Does this look like the intended behavior? Have we actually noticed any problems because of this? If it doesn't actually make a difference, maybe we should just set it to a constant?

### DIFF
--- a/refinery/data_set_manager/models.py
+++ b/refinery/data_set_manager/models.py
@@ -859,7 +859,6 @@ def initialize_attribute_order(study, assay):
     results = _query_solr(study=study, assay=assay)
 
     attribute_order_objects = []
-    rank = 0
     for key in results['response']['docs'][0]:
         is_facet = _is_facet_attribute(key, study, assay)
         is_exposed = _is_exposed_attribute(key)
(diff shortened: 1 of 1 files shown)


flekschas:
@mccalluc No idea. I am not the author of this piece of code. Check https://github.com/refinery-platform/refinery-platform/commit/502b5d52a7183b0b4ada09f4ff299dab52ecc343

mccalluc:
Originally written in 2012. I feel like we should identify some behavior this actually fixes, or maybe even leave rank unassigned?

mccalluc:
I should write a test for this before merge: Behavior at this level has changed, though maybe something at a higher level is also cleaning things up.

mccalluc:
@scottx611x : Sorry to bother, but would you be ok with just not initializing rank? The schema allows it to be null, and if that's a problem it would be good to address that directly.

scottx611x:
@mccalluc I would feel more comfortable just initializing `rank = 0` and not removing it entirely. That is, unless you've pinned down its usages and understand that not initializing it here won't hurt anything downstream

PR #1939, opened 2017-07-24
Scottx611x/workflow launching

### LINKED ISSUE #878: Provide status updates for each Galaxy file import
There is no progress reported for Galaxy file import which can take a long time. We should refactor `import_analysis_in_galaxy()` and `start_galaxy_analysis()` to create separate galaxy_import tasks for each file like in `get_galaxy_download_tasks()`.

scottx611x:
This PR allows for Tool-based Workflow launches file's to be imported into a galaxy history with a progress update for each imported file.

- Fixes #878 for new Tool-based Analysis launches

### DIFF
--- a/refinery/analysis_manager/migrations/0002_analysisstatus_galaxy_import_progress.py
+++ b/refinery/analysis_manager/migrations/0002_analysisstatus_galaxy_import_progress.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('analysis_manager', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='analysisstatus',
+            name='galaxy_import_progress',
+            field=models.PositiveSmallIntegerField(default=0),
+            preserve_default=True,
+        ),
+    ]
--- a/refinery/analysis_manager/models.py
+++ b/refinery/analysis_manager/models.py
@@ -34,6 +34,8 @@ class AnalysisStatus(models.Model):
     galaxy_import_progress = PositiveSmallIntegerField(default=0)
 
     #: percentage of successfully processed datasets in Galaxy history
+    # TODO: refactor `galaxy_history_progress` to take advantage of a
+    # default value of 0, and
     galaxy_history_progress = PositiveSmallIntegerField(blank=True, null=True)
 
     def __unicode__(self):
--- a/refinery/analysis_manager/models.py
+++ b/refinery/analysis_manager/models.py
@@ -50,6 +50,9 @@ class AnalysisStatus(models.Model):
         return get_task_group_state(self.refinery_import_task_group_id)
 
     def galaxy_file_import_state(self):
+        return get_task_group_state(self.galaxy_import_task_group_id)
+
+    def tool_based_galaxy_file_import_state(self):
         if self.galaxy_history_state and self.galaxy_import_progress != 0:
             galaxy_file_import_state = [{
                 'state': self.galaxy_history_state,
--- a/refinery/analysis_manager/models.py
+++ b/refinery/analysis_manager/models.py
@@ -29,6 +29,10 @@ class AnalysisStatus(models.Model):
     #: state of Galaxy history
     galaxy_history_state = CharField(max_length=10, blank=True,
                                      choices=GALAXY_HISTORY_STATES)
+
+    #: percentage of successfully imported datasets in Galaxy history
+    galaxy_import_progress = PositiveSmallIntegerField(default=0)
+
     #: percentage of successfully processed datasets in Galaxy history
     galaxy_history_progress = PositiveSmallIntegerField(blank=True, null=True)
 
@@ -46,7 +50,14 @@ class AnalysisStatus(models.Model):
         return get_task_group_state(self.refinery_import_task_group_id)
 
     def galaxy_import_state(self):
-        return get_task_group_state(self.galaxy_import_task_group_id)
+        if self.galaxy_history_state and self.galaxy_import_progress != 0:
+            galaxy_import_state = [{
+                'state': self.galaxy_history_state,
+                'percent_done': self.galaxy_import_progress
+            }]
+        else:
+            galaxy_import_state = []
+        return galaxy_import_state
 
     def galaxy_analysis_state(self):
         if self.galaxy_history_state and self.galaxy_history_progress:
--- a/refinery/analysis_manager/models.py
+++ b/refinery/analysis_manager/models.py
@@ -49,15 +49,15 @@ class AnalysisStatus(models.Model):
     def refinery_import_state(self):
         return get_task_group_state(self.refinery_import_task_group_id)
 
-    def galaxy_import_state(self):
+    def galaxy_file_import_state(self):
         if self.galaxy_history_state and self.galaxy_import_progress != 0:
-            galaxy_import_state = [{
+            galaxy_file_import_state = [{
                 'state': self.galaxy_history_state,
                 'percent_done': self.galaxy_import_progress
             }]
         else:
-            galaxy_import_state = []
-        return galaxy_import_state
+            galaxy_file_import_state = []
+        return galaxy_file_import_state
 
     def galaxy_analysis_state(self):
         if self.galaxy_history_state and self.galaxy_history_progress:
--- a/refinery/analysis_manager/tasks.py
+++ b/refinery/analysis_manager/tasks.py
@@ -229,6 +229,7 @@ def _refinery_file_import(analysis_uuid):
     """
     analysis = _get_analysis(analysis_uuid)
     analysis_status = _get_analysis_status(analysis_uuid)
+    tool = _get_tool(analysis_uuid)
 
     if not analysis_status.refinery_import_task_group_id:
         logger.info("Starting analysis '%s'", analysis)
@@ -237,7 +238,11 @@ def _refinery_file_import(analysis_uuid):
                     analysis)
         refinery_import_tasks = []
 
-        input_file_uuid_list = analysis.get_input_file_uuid_list()
+        if tool:
+            input_file_uuid_list = tool.get_input_file_uuid_list()
+        else:
+            input_file_uuid_list = analysis.get_input_file_uuid_list()
+
         for input_file_uuid in input_file_uuid_list:
             refinery_import_task = import_file.subtask((input_file_uuid,))
             refinery_import_tasks.append(refinery_import_task)
--- a/refinery/analysis_manager/tasks.py
+++ b/refinery/analysis_manager/tasks.py
@@ -18,6 +18,7 @@ from core.utils import get_full_url
 from data_set_manager.models import Node
 from file_store.models import FileStoreItem
 from file_store.tasks import create, import_file
+import tool_manager
 
 from .models import AnalysisStatus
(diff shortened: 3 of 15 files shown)


refinery/analysis_manager/migrations/0002_analysisstatus_galaxy_import_progress.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('analysis_manager', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='analysisstatus',
+            name='galaxy_import_progress',
+            field=models.PositiveSmallIntegerField(default=0),
```
mccalluc: What values does Galaxy return? 0-100?

refinery/analysis_manager/models.py
```diff
@@ -29,6 +29,10 @@ class AnalysisStatus(models.Model):
     #: state of Galaxy history
     galaxy_history_state = CharField(max_length=10, blank=True,
                                      choices=GALAXY_HISTORY_STATES)
+
+    #: percentage of successfully imported datasets in Galaxy history
+    galaxy_import_progress = PositiveSmallIntegerField(default=0)
```
mccalluc: Is there are reason that this is `default=0` and just below it's `blank=True, null=True`?

refinery/analysis_manager/models.py
```diff
@@ -45,9 +49,19 @@ def set_galaxy_history_state(self, state):
...
         return get_task_group_state(self.galaxy_import_task_group_id)
 
+    def tool_based_galaxy_file_import_state(self):
+        if self.galaxy_history_state and self.galaxy_import_progress != 0:
+            galaxy_file_import_state = [{
+                'state': self.galaxy_history_state,
+                'percent_done': self.galaxy_import_progress
+            }]
+        else:
+            galaxy_file_import_state = []
```
mccalluc: Is the if-then necessary? It feels like a case where we're introducing a variation in the structure and all the code downstream will need to take both possibilities into account.

refinery/analysis_manager/tasks.py
```diff
@@ -86,6 +87,17 @@ def get_taskset_result(task_group_id):
     return TaskSetResult.restore(task_group_id)
 
 
+def _get_tool(analysis_uuid):
+    try:
+        return tool_manager.models.Tool.objects.get(
+            analysis__uuid=analysis_uuid
+        )
+    except (tool_manager.models.Tool.DoesNotExist,
+            tool_manager.models.Tool.MultipleObjectsReturned) as e:
```
mccalluc: Are both errors of the same class? I'm wondering if `DoesNotExist` is actually ok for some situations, but `MultipleObjectsReturned` indicates a deeper problem with the database?

refinery/analysis_manager/tasks.py
```diff
@@ -336,6 +358,67 @@ def _run_galaxy_workflow(analysis_uuid):
...
+
+        galaxy_import_tasks = [
+            _tool_based_galaxy_file_import.subtask(
+                (
+                    analysis_uuid,
+                    file_store_item_uuid,
+                    history_dict,
+                    library_dict,
+                )
+            ) for file_store_item_uuid in tool.get_input_file_uuid_list()
```
mccalluc: Do you need to handle the case of tool being `None`?

refinery/analysis_manager/tasks.py
```diff
@@ -336,6 +358,67 @@ def _run_galaxy_workflow(analysis_uuid):
...
+        galaxy_file_import_taskset.save()
+
+        analysis_status.galaxy_import_task_group_id = (
+            galaxy_file_import_taskset.taskset_id
+        )
+        analysis_status.set_galaxy_history_state(AnalysisStatus.PROGRESS)
+        run_analysis.retry(countdown=RETRY_INTERVAL)
+
+    # Check if data files were successfully imported into Galaxy
+    galaxy_file_import_taskset = get_taskset_result(
```
mccalluc: maybe the variable should be taskset_result?

refinery/analysis_manager/tasks.py
```diff
@@ -336,6 +358,67 @@ def _run_galaxy_workflow(analysis_uuid):
...
+        analysis_status.galaxy_import_task_group_id
+    )
+    if not galaxy_file_import_taskset.ready():
+        logger.debug("Analysis '%s' pending in Galaxy", analysis)
+        run_analysis.retry(countdown=RETRY_INTERVAL)
+    elif not galaxy_file_import_taskset.successful():
```
mccalluc: Are you sure this is reachable? I could imagine if something is not `successful` it might not be `ready` either.

refinery/analysis_manager/tasks.py
```diff
@@ -336,6 +358,67 @@ def _run_galaxy_workflow(analysis_uuid):
...
+        error_msg = "Analysis '{}' failed in Galaxy".format(analysis)
+        logger.error(error_msg)
+        analysis.set_status(Analysis.FAILURE_STATUS, error_msg)
+        analysis_status.set_galaxy_history_state(AnalysisStatus.ERROR)
+        analysis.send_email()
```
mccalluc: Can send_email fail? Would you like to do the cleanup first if so? Not sure how important cleaning up is.

refinery/analysis_manager/tasks.py
```diff
@@ -448,6 +531,53 @@ def _start_galaxy_analysis(analysis_uuid):
...
+        get_full_url(file_store_item.get_datafile_url())
+    )
+    history_dataset_dict = (
+        connection.histories.upload_dataset_from_library(
+            history_dict["id"],
+            library_dataset_dict[0]["id"]
+        )
+    )
+
+    tool.update_file_relationships_with_galaxy_history_data(
```
mccalluc: same concerns about tool being None.

refinery/analysis_manager/tasks.py
```diff
@@ -448,6 +531,53 @@ def _start_galaxy_analysis(analysis_uuid):
...
+        {
+            "refinery_file_uuid": file_store_item_uuid,
```
mccalluc: define a const for repeated string?

refinery/analysis_manager/tasks.py
```diff
@@ -448,6 +531,53 @@ def _start_galaxy_analysis(analysis_uuid):
...
+    )
+
+    number_of_files = len(tool.get_input_file_uuid_list())
+    single_file_percentage = (100 / number_of_files)
+    analysis_status.galaxy_import_progress = (
+        analysis_status.galaxy_import_progress + single_file_percentage
+    )
+    analysis_status.save()
+
+    if (analysis_status.galaxy_import_progress ==
```
mccalluc: Testing floats for equality is never a good idea. Can you find some way to do this based on the number of files processed?

refinery/analysis_manager/tasks.py
```diff
@@ -448,6 +531,53 @@ def _start_galaxy_analysis(analysis_uuid):
...
+            single_file_percentage * number_of_files):
```
mccalluc: This will be 100?

refinery/analysis_manager/tasks.py
```diff
@@ -448,6 +531,53 @@ def _start_galaxy_analysis(analysis_uuid):
...
+        analysis_status.set_galaxy_history_state(AnalysisStatus.OK)
+
+    # - Create DatasetCollection
+    # - Invoke galaxy workflow (in a new history so that fetching outputs
+    #   is easy!)
+    # - Attach results of workflow (Analysis.attach_outputs_dataset is
+    # really nasty :/ )
```
mccalluc: What does this comment belong to?

refinery/analysis_manager/views.py
```diff
@@ -72,11 +73,24 @@ def analysis_status(request, uuid):
     if request.is_ajax():
         ret_json = {
             'refineryImport': status.refinery_import_state(),
-            'galaxyImport': status.galaxy_import_state(),
             'galaxyAnalysis': status.galaxy_analysis_state(),
             'galaxyExport': status.galaxy_export_state(),
             'overall': analysis.get_status(),
         }
+        try:
+            Tool.objects.get(analysis=analysis)
```
mccalluc: Since you don't actually need the object, perhaps just use `count`?

refinery/tool_manager/models.py
```diff
@@ -280,14 +281,33 @@ def __str__(self):
         )
 
     def get_input_file_uuid_list(self):
-        # Tools can't be created without the `node_uuid_list` existing so no
+        # Tools can't be created without the `file_uuid_list` existing so no
         # KeyError is being caught
-        return self.get_tool_launch_config()["node_uuid_list"]
+        return self.get_tool_launch_config()["file_uuid_list"]
```
mccalluc: What motivates the name change?

refinery/tool_manager/models.py
```diff
@@ -280,14 +281,33 @@ def __str__(self):
...
+
         return ast.literal_eval(
-            self.get_tool_launch_config()["file_relationships"]
+            self.get_tool_launch_config()["file_relationships_galaxy"]
+        )
+
+    def get_node_uuids(self):
+        tool_launch_config = self.get_tool_launch_config()
+        node_uuids = re.findall(
+            r'[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}',
```
mccalluc: We've copied and pasted this RE too many times: It's another constant string.

refinery/tool_manager/models.py
```diff
@@ -280,14 +281,33 @@ def __str__(self):
...
+            tool_launch_config["file_relationships"]
```
mccalluc: Is this a blob of JSON or something? I feel uncomfortable just throwing an RE at it if there really is a structure there.

refinery/tool_manager/models.py
```diff
@@ -354,24 +388,24 @@ def _launch_workflow(self):
...
 
-        analysis_url = "/data_sets2/{}/#/analyses/".format(self.dataset.uuid)
-        return JsonResponse({"tool_url": analysis_url})
+        return JsonResponse(
+            {
+                "tool_url": "/data_sets2/{}/#/analyses/".format(
+                    self.dataset.uuid
+                )
+            }
+        )
```
mccalluc: hmm... scratch variables aren't a bad thing. Tradeoffs.

refinery/tool_manager/models.py
```diff
@@ -390,32 +424,65 @@ def set_analysis(self, analysis_uuid):
...
 
-        for uuid in node_uuids:
-            file_url = get_file_url_from_node_uuid(uuid)
-            tool_launch_config["file_relationships"] = (
-                tool_launch_config["file_relationships"].replace(
-                    uuid, "'{}'".format(file_url)
+            file_url = get_file_url_from_node_uuid(node_uuid)
+            tool_launch_config["file_relationships_urls"] = (
+                tool_launch_config["file_relationships_urls"].replace(
+                    node_uuid, "'{}'".format(file_url)
```
mccalluc: I'd like to understand this better.

refinery/analysis_manager/models.py
scottx611x: Look into a refactor if possible, or a TODO/comment if not

refinery/analysis_manager/tasks.py
scottx611x: probably, but I'll just make a note since its out of the scope of this PR

PR #1941, opened 2017-07-24
Mccalluc/public datasets regression

### LINKED ISSUE #1940: Publicly shared files are not visible when not logged in
* Specific code commit: b5ce7da220a32515e8656d0e706c6305168bf7dc (We think this may be a recent regression)
* Version of the web browser and OS: Mas Chrome
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): local development

### Steps to reproduce

0. Login
1. Upload a file, perhaps hg19-metadata-s3 from the testing protocol
2. Click on it to get the option to "Share", and click on it
3. Click the radio button for "Public" "Read-only"
4. It does get the "Globe" icon
5. Logout

### Observed behavior
* The data set is not listed on the public dashboard

### Expected behavior
* It should be listed. We think this may be a recent regression, so I'll try reverting to recent versions and will see if the behavior changes.

mccalluc:
Fix #1940, and do some refactoring to make it available for `user/files`

### DIFF
--- a/refinery/core/tests.py
+++ b/refinery/core/tests.py
@@ -3,7 +3,7 @@ import random
 import string
 from urlparse import urljoin
 
-from django.contrib.auth.models import Group, User
+from django.contrib.auth.models import AnonymousUser, Group, User
 from django.core.files.uploadedfile import SimpleUploadedFile
 from django.test import TestCase
 from django.utils import timezone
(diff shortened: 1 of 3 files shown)


mccalluc:
@scottx611x : I think I should add a test here, but is the direction ok?

mccalluc:
@scottx611x : Do these tests look ok?

refinery/core/tests.py
```diff
@@ -1492,6 +1492,52 @@ def setUp(self):
             "910117c5-fda2-4700-ae87-dc897f3a5d85"
             ]
 
+    def test_get_resources_for_user(self):
+        anon_user = AnonymousUser()
```
scottx611x: I would use `get_anaoymous_user()` here to be consistent, but other than that tests look good to me.

mccalluc: Thanks: In the next commit I add tests for the Guardian user. Will merge, assuming tests pass.

PR #1951, opened 2017-07-31
More JS cleaning

mccalluc:


refinery/ui/source/js/dashboard/controllers/tutorial-data-upload-second-page.js
```diff
@@ -5,7 +5,7 @@
 function dataUploadTutorialCtrl2 (
   $scope, tutorialPageNavigation, updateUserTutorials) {
   var stepText = $scope.refineryTutorialSteps.DATA_UPLOAD_TUTORIAL;
-  var lastEventId = null;
+  // var lastEventId = null;
```
scottx611x: Comment here?

refinery/ui/source/js/dashboard/controllers/tutorial-data-upload-second-page.js
```diff
@@ -20,12 +20,10 @@ function dataUploadTutorialCtrl2 (
   $scope.dataUploadExitEvent = function () {
   };
 
-  $scope.dataUploadChangeEvent = function (event) {
-    lastEventId = event.id;
```
scottx611x: This looks like a No-op, but I just wanted to be sure

refinery/ui/.eslintignore
```diff
@@ -4,5 +4,3 @@ development/**
 docs/**
 node_modules/**
 production/**
-
-source/js/provvis/**
```
scottx611x: Could we just keep this instead of adding the `/* eslint-disable */`s below?

refinery/ui/source/js/dashboard/controllers/tutorial-data-upload-second-page.js
mccalluc: yeah: The variable is never used, so seemed good to get rid of it... but there's another no-op just above, so it doesn't seem terrible to add another: Always possible that something looks for some function by that name.

refinery/ui/.eslintignore
mccalluc: I'll revert this, but...
- If we can get down to zero warnings, it makes it more likely that we'll keep it clean going forward
- it's probably more obvious to have a TODO at the top of the file that needs fixing, than to ignore a whole directory

refinery/ui/source/js/dashboard/controllers/tutorial-data-upload-second-page.js
mccalluc: I'll delete

mccalluc:
Added the global ignore back in, but also leaving the file-by-file. I would need to jump through hoops to get pre-commit to ignore them, now that they've been touched.

Issue #942, opened 2016-03-12
Replace jshint with eslint


flekschas:
ESLint is heavily developed and supported by one of the smartes guys in the JS scene. It's pluggable (offering loooots of good rules for formatting code) and is used by major style guides like Airbnb.

flekschas:
I've exchanged jshint with eslint. You can check it out in branch flekschas/eslint. I've also installed Airbnb's style guide and fixed all issues in `Gruntfile.js` but we might want to hold of with merging it into develop because **we have a lot of issues to fix**:

```bash
$ grunt eslint
Running "eslint:sourceCode" (eslint) task
...
✖ 9675 problems (9574 errors, 101 warnings)

Warning: Task "eslint:sourceCode" failed. Use --force to continue.

Aborted due to warnings.
```

This issues are no show stoppers but if we want to have really clean, good-looking code we should address them at some point.

flekschas:
This might be a very useful preset as well. It comes with tons of Angular best practices such as using `controllerAs`:
https://github.com/Gillespie59/eslint-plugin-angular

flekschas:
Aftersetting up esformatter and going through _commons_, _dashboard_, _globals_, _node-mapping_, and _node-relationships_ we're down to: **6657 problems** (6604 errors, 53 warnings)

flekschas:
After applying the auto-formatting to all JS files we are down to **1846 problems** (1805 errors, 41 warnings). (Commit: dc136e38488fa4ae8907fc937e28de557ef15ff8)

It is not the most exiting thing to do but it's definitely possible to fix all those issues in less then one day. (I went through ~2000 today with configuring _esformatter_.)

Lesson learn: Apart from many style related fixes, I was able to fix a couple of more serious bugs and removed lots of unused / old code.

flekschas:
Formatted and cleaned Antons code.

Without the provvis* we're down to: 697 problems (668 errors, 29 warnings)

*) The provvis directory needs a more thorough refactoring, see here #1000.

In order to be able to merge the code we could also ignore other directories as well.

PR #1953, opened 2017-08-01
Not using jshint since #942

mccalluc:


mccalluc:
@scottx611x : PR fail is on `test_visualization_container_launch_IGV`: Is that surprising?

scottx611x:
@mccalluc Yes definitely is, I'll take note of it.

PR #1961, opened 2017-08-01
Mccalluc/remove esformatter

### LINKED ISSUE #1952: How to run esformatter? Does it have role in build?
ESFormatter was introduced in #982.

- How is it run? It seems to be missing a dependency:
```
$ grunt esformatter **/*
Warning: Task "bower_components/angular" not found. Use --force to continue.
Aborted due to warnings.
```

- Does it, or should it have any role in the build? I don't think it does right now, but maybe it should, or maybe I'm missing something?

mccalluc:
Fixes  #1952


mccalluc:
Ack: Forgot to switch branches. Will delete this and push the right one.

PR #1964, opened 2017-08-02
Mccalluc/user files sort by filename

### LINKED ISSUE #1960: user/files filename sort doesn't work

mccalluc:
fix #1960


refinery/ui/source/js/user-file-browser/services/user-file.js
```diff
@@ -40,8 +40,14 @@
             },
             sort: function () {
               var sort = userFileSortsService.fields.map(function (field) {
-                return field.name + '_Characteristics_generic_s ' + field.direction
-                    + ', ' + field.name + '_Factor_Value_s ' + field.direction;
+                var name = field.name;
+                var dir = field.direction;
+                if (name === 'filename') {
```
mccalluc: Would like to have this reference `FILENAME`, but not sure about the best idiom for that.

hackdna: What is `dir`? I'd avoid abbreviations for clarity or simply not use those extra intermediate variables.

mccalluc:
Looking at this more closely: It's sorting by the download url, not just the filename. Options:
- Remove sort on Filename
- Add a new field to solr index
- See if solr can manipulate strings before sorting.

I'd still like to get this PR in, and then put the fixes on top of that.

PR #1969, opened 2017-08-02
Mccalluc/make filename normal field

mccalluc:
Remove the special handling for filename and instead index it with the others


mccalluc:
igv failure again.

PR #1970, opened 2017-08-02
1000 rows: see if there is a performance cost

mccalluc:
Let's try this out locally and see what the performance is like.


scottx611x:
I'm not at the 1000 file mark, but I have a couple hundred here at least and it doesn't seem to affect anything too differently than beforehand.
aug-03-2017 15-08-21

PR #1974, opened 2017-08-03
Mccalluc/where is technology

### LINKED ISSUE #1972: /user/files: where is technology?

mccalluc:
This fixes #1972... but the other attributes of assay seem equally plausible, so I think they should all be indexed. (Otherwise, someone will tweak the site config at some point, and the data they're looking for just won't be there.)

```python
>>> pprint({attr:set([getattr(assay,attr,None) for assay in Assay.objects.all()]) for attr in ['measurement','measurement_accession','measurement_source','technology_accession','technology_source','platform']})
{'measurement': set([None,
                     u'transcription factor binding site identification',
                     u'transcription profiling']),
 'measurement_accession': set([None,
                               u'http://purl.obolibrary.org/obo/OBI_0000291',
                               u'http://purl.obolibrary.org/obo/OBI_0000424']),
 'measurement_source': set([None, u'OBI']),
 'platform': set([None,
                  u'Affymetrix Human Genome U133A Array; Affymetrix Human Genome U133B Array',
                  u'Affymetrix Mouse Expression 430A Array; Affymetrix Mouse Expression 430B Array',
                  u'Affymetrix Mouse Genome 430 2.0 Array',
                  u'Affymetrix Mouse Genome 430A 2.0 Array',
                  u'Affymetrix Mouse Genome U74A v2',
                  u'Helicos HeliScope Single Molecule Sequencer',
                  u'Illumina Genome Analyzer',
                  u'Illumina Genome Analyzer II',
                  u'Illumina HiSeq 2000']),
 'technology_accession': set([None]),
 'technology_source': set([None, u'', u'OBI'])}
```


refinery/data_set_manager/search_indexes.py
```diff
@@ -66,13 +66,17 @@ def prepare(self, object):
...
-        suffix = "_" + suffix + "_s"
+        id_suffix = "_" + id_suffix + "_s"
         generic_suffix = "_generic_s"
 
         data['filename_Characteristics' + generic_suffix] = \
             re.sub(r'.*/', '', data['name'])
+        data['technology_Characteristics' + generic_suffix] = \
+            set() if (object.assay is None or
+                      object.assay.technology is None) \
+            else {object.assay.technology}
```
scottx611x: Could this be cleaned up?

Its not immediately apparent why we need an empty `set()`

maybe something like:
```
technology_key = 'technology_Characteristics' + generic_suffix
if (object.assay is None orobject.assay.technology is None):
    # We need an empty set here because: ...
    data[technology_key] = set()
else:
    data[technology_key] = object.assay.technology
```
and an additional test to assert the empty set if there is no technology or Assay

mccalluc:
@scottx611x : Want to give it another look? Big change in behavior is to get all likely fields from Assay. (Looked at Study and Investigation, but there's really nothing there.) Also, pulled it into private method, which is hopefully more clear?

PR #1975, opened 2017-08-04
Mccalluc/max height

mccalluc:
This can be argued:
- Is this an ok height?
- Should there be some on-screen indication that a facet list is truncated? Maybe just some kind of fuzzy border on the bottom? But we wouldn't want it on lists that weren't truncated.
- The `:hover` doesn't feel like a distraction, at least to me, but it does mean the `overflow: scroll` is moot, and to a lesser degree, the whole fold / unfold machinery.

screen shot 2017-08-04 at 1 15 16 pm

### DIFF
--- a/refinery/ui/source/styles/file-browser.less
+++ b/refinery/ui/source/styles/file-browser.less
@@ -302,9 +302,11 @@
 
 #table-filter {
   height: 81vh;
+  max-height: 81vh;
   margin: 0 0 0 10px;
   padding: 0;
   border: 1px solid #d4d4d4;
+  overflow: auto;
   h3 {
     font-weight: bold;
   }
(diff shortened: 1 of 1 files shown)


ngehlenborg:
Is there any visual indicator that div content is scrollable?

hackdna:
One option is to show all entries for each of the facets to avoid scrolling within each facet when it is open. This may or may not require displaying facets closed initially.

mccalluc:
@ngehlenborg :
> Is there any visual indicator that div content is scrollable?

There isn't a good way to get native scrollbars to show reliably on overflow content on recent Mac OS versions. Could imagine some treatment at the bottom of the div (like fading to transparent), but I'd suggest getting this much out, and then filing a new bug if unsatisfactory.

PR #1976, opened 2017-08-05
Mccalluc/search facet labels

mccalluc:
@ngehlenborg : Does this look like what you want? (Note that it's just a show-hide, so any previously selected checkboxes still apply... which might be confusing?)
screen shot 2017-08-05 at 6 32 11 pm
screen shot 2017-08-05 at 6 32 27 pm

(Is alphabetical sorting of facets really important? We started down this track because we wanted to make the long, concatenated facets more accessible, but sorting doesn't make a difference if they've been assembled in an arbitrary order.)


mccalluc:
IGV Failure on PR build

PR #1978, opened 2017-08-06
Mccalluc/sorting facets

mccalluc:
screen shot 2017-08-05 at 8 07 02 pm

@ngehlenborg : Did you have something like this in mind?
- Not super-confident that this is more useful than distracting, but if it is:
- maybe sort all the facets either by name or by count, instead of allowing each section to be different?
- Make the buttons lighter, particularly if we're keeping them on each section? (Do we still need the arrows to collapse the sections if the hover part of #1975 is approved? If we could get rid of those arrows there, we might use them here instead to indicate sorting.)
- Instead of disappearing completely, maybe just a gray-out?

### DIFF
--- a/refinery/ui/source/js/user-file-browser/partials/user-file-browser-filters.html
+++ b/refinery/ui/source/js/user-file-browser/partials/user-file-browser-filters.html
@@ -9,7 +9,7 @@
     </div>
     <div
       ng-repeat="(attribute, attributeObj) in $ctrl.attributeFilters track by $index"
-      ng-init="attributeIndex = $index">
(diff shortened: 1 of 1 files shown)


mccalluc:
IGV failure on push build

ngehlenborg:
The functionality is helpful but we need to change the buttons to icons and ideally place them next to the facet name ("section header").

scottx611x:
@ngehlenborg Please checkout #1984 When you have a chance

mccalluc:
Close in favor of #2008

PR #1981, opened 2017-08-07
Mccalluc/preserve facets in fb2

mccalluc:
Perhaps it's best to take this as a proof of concept? The first problem is that filebrowser2 is still coming up blank for me on first load. filebrowser1 can't be targeted here, because it does not update the url when facets are applied. Is that's resolved, then there's the problem that the sets of facets just don't match between the pages. Do facets need to be added to one, or removed from the other?


mccalluc:
(IGV test failure)

PR #1982, opened 2017-08-07
Towards CSV filtering

mccalluc:
I think this is in the right direction towards CSV filtering.  I haven't figured out the right way to template the URL in the HTML, but it might make more sense if I were more awake.


scottx611x:
@mccalluc I poked around for a bit and couldn't get `csvFilterQuery()` to evaluate in the template either. (Take that with a grain of salt though, because my Angular knowledge is limited)

mccalluc:
Close this in favor of #2039 which merges to develop instead.

PR #1984, opened 2017-08-07
More user-friendly sorting buttons

scottx611x:
**Before:** 
aug-07-2017 17-24-47

**After:**
aug-07-2017 17-23-08

### DIFF
--- a/refinery/ui/source/js/user-file-browser/partials/user-file-browser-filters.html
+++ b/refinery/ui/source/js/user-file-browser/partials/user-file-browser-filters.html
@@ -24,14 +24,24 @@
         role="tabpanel"
         aria-labelledby="{{ attribute }}">
         <div class="panel-body">
-          <div class="checkbox container">
+          <div ng-hide="$ctrl.hidden[attribute]" class="checkbox container">
             <div class="row">
               <div class="col-xs-1"></div>
               <div class="col-xs-8">
-                <button ng-click="order = 'name'" ng-show="order === '-count'" class="btn btn-sm">sort</button>
+                <button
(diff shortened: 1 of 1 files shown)


ngehlenborg:
After discussion with @scottx611x: the current solution with buttons between facet name and facet values takes up too much space. 

Let's have a single blue icon sitting above the facet value count (not "button background" fill because that draw too much attention) that can be in one of four states (each represented by the corresponding icon): 

* default: sort by count descending image
* sort by count ascending image
* sort by name A - Z image
*  sort by name Z - A image

Clicking the icon switches to the next state. A tooltip on the icon says: "Sorted by <current state>. Click to sort by <next state>".

sjhosui:
Yes, I like that solution better - a single icon to facilitate all 4 modes for sorting that sits above the counts.

mccalluc:
@ngehlenborg : 
- Are you sure you really want a button that cycles through 4 states? Even cycles through 3 states make me a little uncomfortable. Are the reverse alphabetic and the increasing numeric really that useful?
- I think we're all ok with each section having its own control, but if we're concerned with over-all clutter, could imagine just one button to sort all facets? Not sure that's a good idea, but wanted to bring it up.
- Mockup: Something like this, ideally?
screen shot 2017-08-15 at 10 29 03 am
```
<i style="float: right;" class="fa fa-sort-alpha-asc"></i>
```
Getting it to align with the counts might be tricky, just because that positioning depends on a ratio of the width, and the top line has a greater width.


@scottx611x : Probably close this PR, and make a new one reflecting this discussion?

mccalluc:
Close in favor of #2008

PR #1997, opened 2017-08-16
Hackdna/s3 upload dataset integration

### LINKED ISSUE #1711: Modify tabular metadata file upload form to use S3
jQuery-File-Upload
pre-signed URLs?

hackdna:
Closes #1711


refinery/core/utils.py
```diff
@@ -983,8 +983,7 @@ def move_obj_to_front(obj_arr, match_key, match_value):
 
 
 def api_error_response(error_message, http_status_code):
-    """Return and log error for Django Rest Framework API calls"""
-    logger.error(error_message)
+    """Return a standardized error for Django Rest Framework API calls"""
     return Response({'Error': error_message}, status=http_status_code)
```
mccalluc: Would we like to keep the logging, in addition to returning the error response? Or maybe its handled elsewhere, or maybe it would be log clutter?

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.js
```diff
@@ -5,25 +5,133 @@
     .module('refineryDataSetImport')
     .controller('RefineryFileUploadS3Ctrl', RefineryFileUploadS3Ctrl);
 
-  RefineryFileUploadS3Ctrl.$inject = ['s3UploadService'];
```
jkmarx: Ordering Dependencies (alphabetize with groups)
3rd parties
Custom dependencies (our services)

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.js
```diff
@@ -5,25 +5,133 @@
...
+      });
+    };
+
+    vm.uploadFiles = function () {
+      if (vm.files) {
+        vm.multifileUploadInProgress = true;
+        for (var i = 0; i < vm.files.length; i++) {
+          if (vm.isFileNew(vm.files[i])) {
+            vm.uploadFile(vm.files[i]);
+            return;
```
mccalluc: And then this function is re-entered to upload the next file, and the next file after that? In isolation, it seems to only be uploading one file.

refinery/core/utils.py
hackdna: I've removed logging from here because the messages it dumps do not indicate the actual place in code where the error occurred. Also, it may not be necessary to log all API errors, and definitely not at the same log level.

requirements-aws.txt
```diff
@@ -1,3 +1 @@
-boto3==1.4.4
 django-smtp-ssl==1.0
```
mccalluc: This leaves just one line here -- Is there a reason it should remain?

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.js
```diff
@@ -5,25 +5,133 @@
...
+          $log.error('Error uploading file ' + file.name + ': ' + file.$error);
+          if (vm.multifileUploadInProgress) {
+            vm.uploadFiles();
+          }
+        });
+      });
+    };
+
+    vm.uploadFiles = function () {
+      if (vm.files) {
```
jkmarx: https://developer.mozilla.org/en-US/docs/Glossary/Truthy

refinery/file_store/tasks.py
```diff
@@ -154,8 +171,8 @@ def import_file(uuid, refresh=False, file_size=0):
 
                     # check if we have a sane value for file size
                     if remote_file_size > 0:
-                        percent_done = local_file_size * 100. / \
-                                       remote_file_size
+                        percent_done = \
+                            local_file_size * 100. / remote_file_size
```
scottx611x: Instead of appending the `.` could we do `from __future__ import division`?

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.js
```diff
@@ -5,25 +5,133 @@
...
+      file.managedUpload.promise().then(function () {
+        $scope.$apply(function () {
+          file.success = true;
+          if (vm.multifileUploadInProgress) {
+            vm.uploadFiles();
+          }
         });
+      }, function (error) {
+        $scope.$apply(function () {
+          file.progress = 100;
```
scottx611x: `file.progress = 100` for an erring upload?

refinery/data_set_manager/views.py
```diff
@@ -446,6 +449,20 @@ def post(self, request, *args, **kwargs):
             column.replace('string:', '') for column in source_column_index
         ]
 
+        if settings.REFINERY_DEPLOYMENT_PLATFORM == 'aws':
+            try:
+                identity_id = request.POST.get('identity_id')
+            except (KeyError, ValueError):
+                error = {'error_message': 'identity_id is missing'}
```
hackdna: For some reason, in other places just above it's actually `error_message` for non-AJAX response and `error` for AJAX response which is verbose and confusing, so I thought I'd standardize on `error_message`. I can change it to `error` instead.

refinery/data_set_manager/views.py
```diff
@@ -497,29 +515,55 @@ def post(self, request, *args, **kwargs):
...
         try:
-            base_path = file_data["base_path"]
+            file_data = json.loads(request.body)
+        except ValueError:
+            return HttpResponseBadRequest()
+        try:
+            input_file_list = file_data['list']
         except KeyError:
-            base_path = ""
+            return HttpResponseBadRequest()
```
hackdna: `base_path` is optional, so no response necessary. The file list would not be present only if the are no files in the meta data file. I'm not sure how could that happen. If this is of concern, it should be a separate issue.

refinery/data_set_manager/views.py
```diff
@@ -497,29 +515,55 @@ def post(self, request, *args, **kwargs):
...
+            uploaded_s3_key_list = []
+            s3 = boto3.resource('s3')
+            s3_bucket = s3.Bucket(settings.MEDIA_BUCKET)
+            for s3_object in s3_bucket.objects.filter(
+                    Prefix='uploads/{}'.format(identity_id)
+            ):
+                uploaded_s3_key_list.append(s3_object.key)
+
+        for input_file_path in input_file_list:
+            if not isinstance(input_file_path, unicode):
```
hackdna: Actually, I'm simply re-using existing code here but I was curious about this too. May be @scottx611x could comment since he wrote it originally :)

refinery/file_store/models.py
```diff
@@ -120,23 +120,31 @@ def map_source(source):
...
+
         # process relative path
-        if base_path:
+        if identity_id:
+            source = "s3://{}/uploads/{}/{}".format(
+                settings.MEDIA_BUCKET, identity_id, source
+            )
+        elif base_path:
             source = os.path.join(base_path, source)
         elif username:
```
hackdna: `base_path` and `username` are two alternative options for the Vagrant mode

refinery/file_store/models.py
```diff
@@ -519,18 +527,18 @@ def get_datafile_url(self):
...
             # data file doesn't exist on disk
             if os.path.isabs(self.source):
                 # source is a file system path
-                logger.error("File not found at '%s'",
-                             self.datafile.name)
+                logger.error("File not found at '%s'", self.datafile.name)
                 return None
             else:
                 # source is a URL
+                if self.source.startswith('s3://'):
```
hackdna: It is not really an error but simply an indicator that a file is still being imported from S3. Also, I'm trying to reduce the amount of logging: there a lot of spurious messages that make logs hard to read.

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.js
```diff
@@ -5,25 +5,133 @@
...
+      }
+      return vm.files.some(vm.isFileNew);
+    };
+
+    vm.uploadFile = function (file) {
+      try {
+        file.managedUpload = s3UploadService.upload(file);
+      } catch (e) {
+        file.progress = 100;
+        file.$error = 'Data upload configuration error';
```
hackdna: It is already logged to console in `s3UploadService` and I'll another log statement there too.

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.js
hackdna: Exactly: `uploadFiles()` is called from `uploadFile()`. It is a bit too clever, so I'll add a comment.

refinery/data_set_manager/views.py
scottx611x: I remember plain strings weren't good enough here. Before we were check `isinstance` on `str`, and I believe that behavior changed after upgrading to Django 1.7.

refinery/templates/data_set_manager/import.html
```diff
@@ -18,7 +18,7 @@
 <link
   rel="stylesheet"
   type="text/css"
-  href="{% static "vendor/intro.js/minified/introjs.min.css" %}">
```
scottx611x: Just wondering if all tutorials have been tested after making this change?

requirements-aws.txt
hackdna: Yes, we might add more in the future. Also, we should split up `requirements.txt` more.

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.js
hackdna: Right, we don't want to upload files if `vm.files` is `undefined`

refinery/file_store/tasks.py
hackdna: Probably but I don't know enough about this to comment. I simply re-arranged existing code for readability.

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.js
hackdna: Yes, to show a red stripe across the progress bar area.

refinery/templates/data_set_manager/import.html
hackdna: That's what release testing is for :) Actually, `GET vendor/intro.js/minified/introjs.min.css` was giving a 404, so this change can only be for the better :)

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.js
```diff
@@ -5,25 +5,134 @@
...
+      if (typeof vm.files === 'undefined') {
+        vm.files = files;
+      } else {
+        vm.files = vm.files.concat(files);
+      }
+    };
+
+    vm.isFileNew = function (file) {
+      // check if file upload has been attempted
+      return typeof file.progress === 'undefined';
```
jkmarx: http://underscorejs.org/#has

Better for checking object properties

jkmarx: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined

"However, this kind of technique should be avoided. JavaScript is a statically scoped language, so knowing if a variable is declared can be read by seeing whether it is declared in an enclosing context. The only exception is the global scope, but the global scope is bound to the global object, so checking the existence of a variable in the global context can be done by checking the existence of a property on the global object (using the in operator, for instance)."

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.js
```diff
@@ -5,25 +5,134 @@
...
+
+    vm.isUploadConfigured = function () {
+      if (s3UploadService.isConfigReady) {
+        return s3UploadService.isConfigValid();
+      }
+      return false;
+    };
+
+    vm.addFiles = function (files) {
+      if (typeof vm.files === 'undefined') {
```
jkmarx: Was there a reason to not initialize vm.files = []?

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.spec.js
```diff
@@ -0,0 +1,269 @@
...
+  describe('RefineryFileUploadS3Ctrl', function () {
+    var $controller;
+    var $rootScope;
+    var s3UploadService;
+    var ctrl;
+    var scope;
+
+    beforeEach(module('refineryApp'));
+    beforeEach(module('refineryDataSetImport'));
+    beforeEach(inject(function (_$controller_, _$rootScope_, _s3UploadService_) {
```
jkmarx: At some point we agreed to update this syntax, since it wasn't necessary.  No need to wrap dependencies.

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.js
hackdna: OK, I'll use `!('progress' in file)`.

refinery/ui/source/js/data-set-import/controllers/file-upload-s3.spec.js
hackdna: OK but I think I should still wrap `s3UploadService` because I would [have to use a different name for it locally in tests](https://stackoverflow.com/a/15318137) otherwise.

PR #2007, opened 2017-08-16
user/files -> files; Not sure about csv

### LINKED ISSUE #2001: simpler url: "user/files" should just be "files"

mccalluc:
fix #2001

### DIFF
--- a/refinery/ui/source/js/dashboard/views/launch-pad.html
+++ b/refinery/ui/source/js/dashboard/views/launch-pad.html
@@ -89,7 +89,7 @@
             <div>
               <span>
                 <a
-                  href="/user/files/"
+                  href="/files/"
                   class="no-underline btn">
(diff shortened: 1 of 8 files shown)


mccalluc:
@scottx611x : Could you review this?
@ngehlenborg : Is `/files_download` ok for the CSV? Or suggest something you prefer: I don't think the whole team needs to be looped into this discussion.

ngehlenborg:
What about `.../files?download=csv`? Could easily be extended to `.../files?download=xls` etc.

Note our naming convention for API endpoints: https://github.com/refinery-platform/refinery-platform/wiki/coding-style-guide#rest-api

mccalluc:
Sure: since that's more than a routing change, I don't want to let that block this bit, but I've filed #2013.

PR #2010, opened 2017-08-17
Try to make this a little easier to read by replacing RE with constant

mccalluc:


refinery/core/urls.py
```diff
@@ -11,66 +11,53 @@
...
-        r'[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})/$',
+    url(r'^group_invite/(?P<token>' + UUID_RE + r')/$',
         'group_invite', name='group_invite'),
     url(r'^users/(?P<query>[\@\.\-\+a-z0-9]+)/$', 'user'),
     # "name" is required for use with the url tag in templates
-    url(r'^users/(?P<query>'
-        r'[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})/$',
+    url(r'^users/(?P<query>' + UUID_RE + r')/$',
         'user', name="user"),
-    url(r'^users/(?P<uuid>'
```
scottx611x: Could we just do a `format()` for these? I feel like it might be easier on the eyes than having the extra `r')/$'` hanging around.

hackdna: I would rather keep it as is to make it easier for anyone who has read Django docs understand what's going on.

refinery/core/urls.py
```diff
@@ -11,66 +11,53 @@
 from .views import (AnalysesViewSet, DataSetsViewSet, NodeViewSet, OpenIDToken,
                     WorkflowViewSet)
 
+UUID_RE = r'[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}'
```
hackdna: Good idea but I'd move it somewhere like `refinery/constants.py` (see Two Scoops of Django, p. 38)

PR #2011, opened 2017-08-17
data_sets2 slugs

mccalluc:
@ngehlenborg : I'm assuming this bit of functionality also needs to be carried over... If not, great, I'd be happy to close this.


ngehlenborg:
Yes, should be carried over. We use that to give public data sets pretty names to reference them, e.g. in publications.

PR #2020, opened 2017-08-18
Mccalluc/delete data sets edit

### LINKED ISSUE #2019: Is data_sets_old/<uuid>/edit still needed?
There is code to handle it, but:
- I don't see that we have any links to it
- and when I visit it, locally, I get a NoReverseMatch error.

My guess is that this is vestigial code which could be deleted, but correct me if I'm wrong?

Scott confirms that
- this can be done a different way with the new browser
- and that the old code was already vestigial and unused.

mccalluc:
Fix #2019


mccalluc:
Push failure is IGV

scottx611x:
#2021 builds off of this

PR #2035, opened 2017-08-22
Mccalluc/isatab test

mccalluc:


mccalluc:
`data_set_manager.tests.AnnotatedNodeExplosionTestCase` fails, but I think that's in `add_error_messages_to_isatab_parser`, not my branch

PR #2036, opened 2017-08-22
Add error messages to isatab parser

### LINKED ISSUE #2026: ISA-Tab parsing: report error messages
(Might choose to no longer parse directly to DB, because that makes it harder to provide feedback to user.)

scottx611x:
Fixes #2026

- [x] Add tests
- [x] Make sure coverage is up to par


refinery/data_set_manager/isa_tab_parser.py
```diff
@@ -565,11 +571,14 @@ def _parse_term_information(self, headers, row):
                 row.popleft()
```
mccalluc: I'm stepping back and wondering if it'd be more helpful if the error message described the pair of elements it expected to see, rather than just the one. "Unexpected element" doesn't really tell the user what *was* expected.

(The code is also redundant in many ways: `row.popleft()` returns the value popped.)

refinery/data_set_manager/isa_tab_parser.py
```diff
@@ -930,9 +947,10 @@ def run(self, path, isa_archive=None, preisa_archive=None):
                     # the extract path
                     for name in zip.namelist():
                         if name.startswith("..") or name.startswith("/"):
-                            logger.exception(
-                                "Unable to extract assumed ISArchive file \"" +
-                                path + "\" due to illegal file path: " + name
+                            raise ParserException(
+                                "Unable to extract assumed ISArchive "
+                                "file {!r} due to illegal file path: {}"
```
mccalluc: I don't remember why repr is necessary here?

refinery/data_set_manager/isa_tab_parser.py
```diff
@@ -983,10 +1005,10 @@ def run(self, path, isa_archive=None, preisa_archive=None):
                             self._parse_assay_file(study, assay,
                                                    assay_file_name)
         else:
-            logger.exception(
+            error_message = (
```
mccalluc: Or just put it inside the constructor like everywhere else?

refinery/data_set_manager/tests.py
```diff
@@ -1777,34 +1782,176 @@ def test_prepare(self):
...
+
+        assays = studies[0].assay_set.all()
+        self.assertEqual(len(assays), 1)
+
+    def test_mising_investigation(self):
+        with self.assertRaises(ParserException):
+            self.parse('missing-investigation')
+
+    def test_mising_study(self):
+        with self.assertRaises(IOError):
```
mccalluc: We report all errors to the user, and not just ParserExceptions, right? Even so, might check that the information returned by this is enough to act on.

refinery/data_set_manager/views.py
```diff
@@ -197,8 +201,32 @@ def get(self, request, *args, **kwargs):
...
+                PARSER_UNEXPECTED_ERROR_MESSAGE,
+                traceback.format_exc(e)
+            )
+            logger.error(error_message)
+            return HttpResponseBadRequest(
+                PARSER_UNEXPECTED_ERROR_MESSAGE +
+                e.message
+            )
+        else:
+            dataset_uuid = parse_isatab_invocation
```
mccalluc: Why not set this variable in the first place?

refinery/data_set_manager/isa_tab_parser.py
scottx611x: It would probably be more helpful, but I have no clue as to what should be expected. I'd need some outside input. I can mange fixing some of the redundancies though

refinery/data_set_manager/isa_tab_parser.py
scottx611x: @mccalluc Not sure if its needed either? (You're talking about the `{!r}` ?

refinery/data_set_manager/tests.py
scottx611x: @mccalluc Yep we'll catch anything unexpected. This case will show in the ui like so:
screen shot 2017-10-10 at 11 37 04 am

scottx611x:
@mccalluc I think those travis error have been there for awhile, not related to this PR. Pretty sure they are caused by the Selenium tests mostly. Might be easier to figure out if #2217 is merged

PR #2039, opened 2017-08-23
apply facets to csv download

mccalluc:
Checkpoint: I think the solution is something like this, but I have not gotten csvFilterQuery to evaluate in the template


mccalluc:
Close in favor of #2041

PR #2044, opened 2017-08-24
Mccalluc/update solr from celery

mccalluc:
ISA-Tab import failing locally, but not seeing the connection to these changes right now: will revert to develop and see if it's still there.

Would something like `REFINERY_DOWNLOAD_URL` be a good name for the solr field?


mccalluc:
UI looks good. Will merge when tests pass.
screen shot 2017-08-24 at 3 35 30 pm

PR #2046, opened 2017-08-24
Scottx611x/analysis names in fb2 workaround

scottx611x:
- Replaces the Analysis UUIDs in FB2 Table and Facets with Analysis Names instead.
- This is a quick implementation that works. While attempting to cleanup REFINERY_ANALYSIS_* strings and references I ended up going down a few rabbit holes that seemed to lead to more harm than good, and caused the ui-grid and facets to misbehave.

- [x] Create a new issue regarding the cleanup that should be done


refinery/data_set_manager/tests.py
```diff
@@ -1502,6 +1507,23 @@ def test_get_relative_file_store_item_url(self):
         self.assertEqual(relative_url, self.file_node.get_file_store_item(
         ).get_datafile_url())
 
+    def test_get_analysis(self):
+        make_analyses_with_single_dataset(1, self.user)
+        self.analysis = Analysis.objects.all()[0]
```
mccalluc: inside a test, I'm not sure there's a good reason to use fields instead of plain variables...

refinery/tool_manager/utils.py
```diff
@@ -283,19 +280,17 @@ def create_tool_analysis(validated_analysis_config):
...
         raise RuntimeError("Couldn't fetch Tool from UUID: {}".format(e))
 
     analysis = AnalysisFactory(
         uuid=str(uuid.uuid4()),
         summary="Galaxy workflow execution for: {}".format(tool.name),
-        name="{} - {}".format(
-            name,
-            get_aware_local_time().strftime("%Y-%m-%d @ %H:%M:%S")
-        ),
+        name="{} - {}".format(name, tool.uuid),
```
mccalluc: The point of this is to get a human-friendly name? I'm not sure we want to include the whole uuid still. Imagining what the facet will look like. Sorry if I'm thinking about this wrong.

scottx611x: @mccalluc Yeah, but more so to maintain uniqueness of names for the faceting. I could imagine a case before where users launch Tools in the same second and then there would be a single facet value that corresponds to both launches. 

Maybe updating the naming to include a username/timestamp combination would helpful?

PR #2054, opened 2017-08-25
Maybe this handles visualization warm up errors?

mccalluc:
I have not been able to test this in context:

- When I turn on `DEBUG`, my server doesn't start, so I must be doing it the wrong way: I have not seen the pretty error page locally.
- Other constants defined in the same file are referenced in the base.html template, but I'm not confident they can also be referenced in 500.html

I'm going to swap to something else, and then come back if I think I might have more clarity.

### DIFF
--- a/refinery/templates/500.html
+++ b/refinery/templates/500.html
@@ -13,7 +13,20 @@
 {% block content %}
 <div class="refinery-panel refinery-panel-content scrollable">
 	<div class="row">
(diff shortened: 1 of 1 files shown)


mccalluc:
Talking with Scott, we want something finer grained, at the very least. The `view` for docker visualizations isn't in this project: I think we want to look at `Proxy._proxy_view`. Closing.

Issue #2008, opened 2017-08-17
Facet sorts done right


mccalluc:
Confirmed that we just want to toggle between an alphabetic and a numeric sort.

![screen shot 2017-08-15 at 10 29 03 am](https://user-images.githubusercontent.com/730388/29320127-4139ec48-81a4-11e7-8fe5-af1f67f679aa.png)
```
<i style="float: right;" class="fa fa-sort-alpha-asc"></i>
```
Getting it to align with the counts might be tricky, just because that positioning depends on a ratio of the width, and the top line has a greater width.

ngehlenborg:
@sjhosui: as per discussion with me. We can extend if we need all 4 possible sorting options.

mccalluc:
Fixed by #2055

PR #2055, opened 2017-08-25
Mccalluc/better fold on facet search

mccalluc:
I think this is reasonable behavior for the facet search, but there are certainly other ways it could be done.

There are two modes
- With no search, the facet folding is controlled only by the user.
- With a search, facets can be unfolded by the user, or by having a matching term; only matching terms are shown. So, it's possible to unfold a facet, but not have anything listed under it.

On load state:
screen shot 2017-08-25 at 5 14 50 pm

Fold down "technology":
screen shot 2017-08-25 at 5 15 04 pm

Type "n":
screen shot 2017-08-25 at 5 15 21 pm

Type "na":
screen shot 2017-08-25 at 5 15 41 pm

Type "naX":
screen shot 2017-08-25 at 5 16 00 pm

You can still unfold the facets, but with no matches there's nothing to see:
screen shot 2017-08-25 at 5 16 22 pm

### DIFF
--- a/refinery/ui/source/js/user-file-browser/ctrls/user-file-browser-filters-ctrl.js
+++ b/refinery/ui/source/js/user-file-browser/ctrls/user-file-browser-filters-ctrl.js
@@ -20,10 +20,10 @@
     var vm = this;
 
     vm.togglePanel = function (attribute) {
-      vm.visible[attribute] = ! vm.visible[attribute];
+      vm.folded_down[attribute] = ! vm.folded_down[attribute];
     };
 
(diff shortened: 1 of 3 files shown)


mccalluc:
Bring back the facet sorting. Fix #2008:
screen shot 2017-08-25 at 6 13 16 pm

mccalluc:
highlight the matching substring:
screen shot 2017-08-28 at 10 42 25 am

mccalluc:
got scott's approval.

Issue #2043, opened 2017-08-24
data_sets (new) making unnecessary API calls


mccalluc:
jmarx:
> Are we aware the current data_set (now data_set2) is making all the old api calls in addition to the correct api calls?

mccalluc:
> No, I hadn’t looked at it, but I think I’ve seen something similar on `user/files` aka `files`: more API requests than I need. I know there are the checks in the javascript for whether we’re on data_sets2 or not, and I had though I had replaced them correctly, but maybe not, or do you think something else is to blame? You’re pretty confident this is a regression since the URL swap?

jmarx:
> Yes, because I resolved it for the data_set2 address.

mccalluc:
@jkmarx : I believe your PR #2056 fixed this? Reopen if I'm mistaken

PR #2056, opened 2017-08-26
Prevent old_file_browser api calls and grid generation in current fil…

jkmarx:
Resolves issue #2043

### DIFF
--- a/refinery/static/source/js/refinery/solr/solr_document_table.js
+++ b/refinery/static/source/js/refinery/solr/solr_document_table.js
@@ -107,12 +107,10 @@ SolrDocumentTable.prototype.render = function(solrResponse) {
     .trigger('refinery/solrTable/destroy')
     .html("");
    // Required while data set 2 lives in parallel.
-  if (window.location.href.indexOf('data_sets') === -1) {
-    self._renderTable(solrResponse);
+  self._renderTable(solrResponse);
     //$( "#" + self._parentElementId ).html( code );
     // attach event listeners
     // ..
-  }
 };
 
 SolrDocumentTable.prototype._renderTable = function(solrResponse) {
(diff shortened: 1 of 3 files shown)


refinery/static/source/js/refinery/solr/solr_document_table.js
```diff
@@ -107,12 +107,10 @@ SolrDocumentTable.prototype.render = function(solrResponse) {
     .trigger('refinery/solrTable/destroy')
     .html("");
    // Required while data set 2 lives in parallel.
-  if (window.location.href.indexOf('data_sets') === -1) {
-    self._renderTable(solrResponse);
+  self._renderTable(solrResponse);
     //$( "#" + self._parentElementId ).html( code );
     // attach event listeners
     // ..
-  }
```
mccalluc: [Originally](https://github.com/refinery-platform/refinery-platform/blame/46619c1a75bf765ca874e1797e3ea35db3aa0e24/refinery/static/source/js/refinery/solr/solr_document_table.js#L109), this had been:
```
  if (window.location.href.indexOf('data_sets2') === -1) {
    self._renderTable(solrResponse);
    //$( "#" + self._parentElementId ).html( code );
    // attach event listeners
    // ..
}
```

and the intention seemed to be *not* to run it in FB2, but I'm really not sure what it should be at this point.

jkmarx: @mccalluc The old table generation should only be called when in the data_set old which is provenance. It's hacky, but will work while we are in the transition phase. Open to other suggestions, but note this is temporary.

jkmarx: I relocated the conditional in the document.ready, so it's no longer needed here.

mccalluc: ok, I see the connection now. Thanks!

PR #2017, opened 2017-08-18
Remove unused templates

mccalluc:
Fix #1993?

@jkmarx : Looking at this, I think each of these files is still referenced... Do you remember what you meant in saying that the imports could be removed?

PR #2059, opened 2017-08-28
keep "import.html" this time; remove the others

mccalluc:
Replaces #2017: keep import.html this time.


mccalluc:
scott approves: chuck with confirm that it still renders.

Issue #1890, opened 2017-07-10
Discrepancies between "Details" tab: file browser 1 vs. 2


scottx611x:
* Specific code commit: bc425b3403264b8a0fb1d28cd412a7360e05dc6b

**Changes to be made in File Browser 2:**
- [x] Extra `:` in title
- [ ] Missing `None:None` under `STUDIES`
- [x] Cleaner display of `DATA FILES`
- [x] Prettified dates under `HISTORY`
- [x] Missing `Not Available` under `META DATA`
- [x] Missing `Import into Own Space` button (add "?" info popover to this and have nicer button

## File Browser 1
![screen shot 2017-07-10 at 5 14 02 pm](https://user-images.githubusercontent.com/5629547/28040294-58110fe6-6593-11e7-8817-c82f3fb41992.png)

## File Browser 2
![screen shot 2017-07-10 at 5 13 51 pm](https://user-images.githubusercontent.com/5629547/28040295-58128150-6593-11e7-9f0e-6fd43e793a39.png)

mccalluc:
> Missing `None:None` under `STUDIES`

This does not come through because in the source code we have
```
  <span ng-if="study.identifier">
    <strong>{{study.identifier}}</strong>
  </span>
  <span ng-if="study.title">
    : {{study.title}}
  </span>
```
... which seems more like a feature than a bug?

Also: New one is missing creation date of the current version: Should that also be ported from the old version?

mccalluc:
TODO: deploy this, confirm that all the features are complete.

scottx611x:
@mccalluc Everything looks good to me.

PR #2060, opened 2017-08-28
Mccalluc/details tab

mccalluc:
Towards #1890. 

Two questions about the old one, copied from the thread there:
> Missing `None:None` under `STUDIES`

This does not come through because in the source code we have
```
  <span ng-if="study.identifier">
    <strong>{{study.identifier}}</strong>
  </span>
  <span ng-if="study.title">
    : {{study.title}}
  </span>
```
... which seems more like a feature than a bug?

Also: New one is missing creation date of the current version: Should that also be ported from the old version?

`Import into Own Space` might be more complicated, but this much could be merged, I think.

original:
screen shot 2017-07-10 at 5 14 02 pm

with this pr:
screen shot 2017-08-28 at 3 05 09 pm

### DIFF
--- a/refinery/ui/source/js/data-set-about/partials/details.html
+++ b/refinery/ui/source/js/data-set-about/partials/details.html
@@ -1,6 +1,6 @@
 <div class="refinery-header">
   <div class="refinery-header-left">
-    <h3>{{ ADCtrl.dataSet.accession }}: {{ ADCtrl.dataSet.title }}</h3>
+    <h3>{{ ADCtrl.dataSet.title }}</h3>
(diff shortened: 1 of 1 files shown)


mccalluc:
@scottx611x : Thanks: If you have thoughts on the two questions at the top, let me know?

scottx611x:
@mccalluc regarding the `None:None` i'm not sure if we want to remove the `ng-if`s to keep consistent with the functionality that we had before? It seems odd to me to explicitly state the `None`s, though

As for the creation date of the newest version I think that would be great to include.

PR #2062, opened 2017-08-29
undo copy and paste

mccalluc:
While I was looking at this wanted to fix it.

### DIFF
--- a/refinery/data_set_manager/views.py
+++ b/refinery/data_set_manager/views.py
@@ -231,8 +231,10 @@ class ProcessISATabView(View):
             except KeyError:
                 url = None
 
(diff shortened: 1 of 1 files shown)


refinery/data_set_manager/views.py
```diff
@@ -231,48 +231,27 @@ def post(self, request, *args, **kwargs):
...
-                    return render_to_response(
-                        self.template_name,
-                        context_instance=RequestContext(
-                            request,
-                            {
-                                'form': form,
-                                'error': response.message
-                            }
-                        )
+            response = self.import_by_url(url) if url \
```
scottx611x: multi lline if then

PR #2072, opened 2017-08-31
download url in csv + map to ascii in csv

### LINKED ISSUE #1968: user/files: csv export does not include data urls

### LINKED ISSUE #2064: all files download CSV fails on test if unfiltered (1000+ rows)
* Specific code commit: a679b9904cc8d9c199cd4122f15b502f931aaab9
* Version of the web browser and OS: mac ff
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance):  test production

### Steps to reproduce

1. https://test.stemcellcommons.org/files_download?fq=&sort=
2. 500

@scottx611x : I look like a large number of rows are necessary to reproduce this. Before I start trying to upload all those files locally, could we get the logs from the server? Is that some I should already know how to do?

mccalluc:
fix #1968 and fix #2064.

Could argue whether this is the best approach for handling characters like n-dash in the data. I'm using [unidecode](https://pypi.python.org/pypi/Unidecode) to get best approximate translation of unicode character. I'm working with the assumption that the end-user is not super-technical, and just wants the data in Excel. Alternatives:
- UTF-8 CSV: Possible, but not Excel's default. The user wouldn't notice during the import that the encoding is bad, only latter, and even then they wouldn't know how to fix it.
- Map no-ascii to `#` or ` ` or empty string: reasonable, but less helpful.
- Map individual problem cases: Why reinvent the wheel?
- Limit to ascii back at ingest. No.

### DIFF
--- a/refinery/user_files_manager/tests.py
+++ b/refinery/user_files_manager/tests.py
@@ -41,7 +41,7 @@ class UserFilesAPITests(APITestCase):
 
 
 class UserFilesUITests(StaticLiveServerTestCase):
-    def test_get(self):
+    def test_ui(self):
         response = requests.get(
             urljoin(
                 self.live_server_url,
@@ -49,6 +49,19 @@ class UserFilesUITests(StaticLiveServerTestCase):
             )
         )
         self.assertIn("All Files", response.content)
+
+    def test_csv(self):
+        response = requests.get(
+            urljoin(
+                self.live_server_url,
+                'files_download'
+            )
+        )
+        self.assertEqual(
+            response.content,
+            'url,filename,organism,technology,'
+            'antibody,date,genotype,experimenter\r\n'
+        )
 
 
 class UserFilesUtilsTests(TestCase):
--- a/refinery/user_files_manager/tests.py
+++ b/refinery/user_files_manager/tests.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from urlparse import urljoin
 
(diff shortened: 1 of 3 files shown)


refinery/user_files_manager/tests.py
```diff
@@ -41,7 +44,7 @@ def test_get(self):
 
 
 class UserFilesUITests(StaticLiveServerTestCase):
```
scottx611x: Could inherit from `RefinerySeleniumTestBase` here

refinery/user_files_manager/views.py
```diff
@@ -31,14 +33,19 @@ def user_files_csv(request):
...
+    # DOWNLOAD_URL's internal solr name not good for end-user.
 
     docs = loads(solr_response)['response']['docs']
     for doc in docs:
-        row = []
+        row = [doc.get(NodeIndex.DOWNLOAD_URL) or '']
         for col in cols:
-            row.append(doc.get(col + '_Characteristics_generic_s') or
-                       doc.get(col + '_Factor_Value_generic_s') or '')
+            possibly_unicode = (
```
scottx611x: Might be good to have a clearer name than `possibly_unicode`? Though, it makes the `unidecode` call fairly clear

refinery/user_files_manager/tests.py
mccalluc: If we're just doing HTTP and don't need a mock browser, is there are reason to use `RefinerySeleniumTestBase`?

PR #2093, opened 2017-09-05
Store full urls in solr index. Requires reindex.

### LINKED ISSUE #2086: cross-dataset browser: Download urls are sometimes still paths
Not sure about reproducer: talk to @scottx611x

mccalluc:
Fix #2086. More discussion there.

### DIFF
--- a/refinery/analysis_manager/tasks.py
+++ b/refinery/analysis_manager/tasks.py
@@ -14,7 +14,6 @@ from celery.task.sets import TaskSet
 import requests
 
 from core.models import Analysis, AnalysisResult, Workflow
-from core.utils import get_full_url
 from data_set_manager.models import Node
 from file_store.models import FileStoreItem
 from file_store.tasks import create, import_file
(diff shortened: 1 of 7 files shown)


refinery/data_set_manager/search_indexes.py
```diff
@@ -140,7 +141,7 @@ def prepare(self, object):
         data.update({
             NodeIndex.DOWNLOAD_URL:
                 '' if file_store_item is None
-                else file_store_item.get_datafile_url(),
+                else get_full_url(file_store_item.get_datafile_url()),
```
hackdna: It's probably better to modify `get_datafile_url()` to return an absolute URL for local files. This is how it appears to be used throughout the codebase anyway (wrapped in `get_full_url()`).

refinery/analysis_manager/tasks.py
```diff
@@ -14,7 +14,6 @@
 import requests
```
scottx611x: `dataset_manager.views:90`
`  full_isa_tab_url = get_full_url(request.POST['isa_tab_url'])` -> `  full_isa_tab_url = request.POST['isa_tab_url']`

mccalluc: I'm going to put a TODO on this line: No harm done by having an extra get_full_url call, but if it did confuse something, it could be hard to track down.

PR #2098, opened 2017-09-07
Hackdna/file store settings update

hackdna:
One place to configure `file_store` app. Use `get_setting()` to ensure necessary variables exist.


refinery/file_store/models.py
```diff
@@ -54,34 +54,11 @@ def _mkdir(path):
...
-    settings.FILE_UPLOAD_MAX_MEMORY_SIZE = 0
-
-# http://stackoverflow.com/q/4832626
-FILE_STORE_BASE_URL = \
-    urlparse.urljoin(settings.MEDIA_URL, settings.FILE_STORE_DIR) + '/'
+# create data storage directories
+if not os.path.isdir(settings.FILE_STORE_BASE_DIR):
+    _mkdir(settings.FILE_STORE_BASE_DIR)
+if not os.path.isdir(settings.FILE_STORE_TEMP_DIR):
+    _mkdir(settings.FILE_STORE_TEMP_DIR)
```
mccalluc: If there were a problem creating a directory, `_mkdir` just returns `False` instead of throwing the exception... I'm not sure that that behavior is useful here? Is there a reason not to use `mkdirs`?

hackdna: Return value of `_mkdir()` is never checked so I'll remove corresponding statements. `os.makedirs()` is problematic because it [raises an exception if the leaf directory already exists](https://docs.python.org/2/library/os.html#os.makedirs) which is often the case.

PR #2099, opened 2017-09-07
Scottx611x/fix same name analysis results

scottx611x:
When a DataSetCollection is used as input to a workflow which only accepts a `data_input`, said workflow will be run (N x the number of elements in the collection) with the potential to create N derived data nodes with the same name if the workflow doesn't have a dynamically named `label` for it's `outputs`. 

`Analysis.attach_outputs_dataset()` had logic in place to not attach derived data nodes to their AnalysisNodeConnections if this scenario occured, but this is behavior that we would like now.

Reviewed side by side w/ @mccalluc

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -1655,7 +1655,8 @@ class Analysis(OwnableResource):
             analysis=self, direction=INPUT_CONNECTION)[0].node.assay
         # 1. read workflow into graph
         graph = create_expanded_workflow_graph(
-            ast.literal_eval(self.workflow_copy))
+            ast.literal_eval(self.workflow_copy)
+        )
         # 2. create data transformation nodes for all tool nodes
         data_transformation_nodes = [graph.node[node_id]
                                      for node_id in graph.nodes()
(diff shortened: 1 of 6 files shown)


refinery/core/models.py
```diff
@@ -1740,23 +1730,30 @@ def attach_outputs_dataset(self):
...
+                        )
                         derived_data_file_node.add_child(
-                            data_transformation_input_node)
+                            data_transformation_input_node
+                        )
                         # TODO: here we could add a (Refinery internal)
-                        # attribute to the derived data file node to indicate
-                        # which output of the tool it corresponds to
-            # connect outputs that are not inputs for any data transformation
+                        # attribute to the derived data file node to
```
scottx611x: fix comment line

refinery/tool_manager/models.py
```diff
@@ -1116,6 +1124,7 @@ def update_file_relationships_with_galaxy_history_data(self):
 
         for galaxy_to_refinery_dict in galaxy_to_refinery_mapping_list:
             node = Node.objects.get(
+                uuid__in=self.get_input_node_uuids(),
```
scottx611x: test this

scottx611x:
default dict usage, tuples for mapping

PR #2100, opened 2017-09-07
Mccalluc/use new django docker

mccalluc:
This depends on https://github.com/refinery-platform/django_docker_engine/pull/75 being released and tagged, but even apart from that, I suspect things could be done differently here.

### DIFF
--- a/refinery/tool_manager/urls.py
+++ b/refinery/tool_manager/urls.py
@@ -1,3 +1,5 @@
+from os.path import dirname, join
+
 from django.conf import settings
 from django.conf.urls import include, url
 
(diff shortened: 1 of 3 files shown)


refinery/tool_manager/urls.py
```diff
@@ -11,7 +13,14 @@
 tool_manager_router.register(r'tools', ToolsViewSet)
 tool_manager_router.register(r'tool_definitions', ToolDefinitionsViewSet)
 
-url_patterns = Proxy(FileLogger(settings.PROXY_LOG)).url_patterns()
```
scottx611x: Could we just pass in a template and have the `PleaseWaitView` render it? Not sure if thats the best approach, but I think it would be more straightforward than this approach. 

This could be shortsighted of me, but it would be easier to extend a template in the future, than a simple string.

Could utilize the nifty:
`django.shortcuts.render_to_response(template)`

mccalluc: We talked, and there's really not that much information that would go in the template. If we did want the template to be filled out for each request (perhaps giving information coming from Docker? Or depending on the type of underlying error we hit?) that would need to be done inside of django_docker_engine.

mccalluc:
Given Scott's earlier approval, I'll go ahead and merge.

PR #2103, opened 2017-09-08
Purge docker containers after inactivity

mccalluc:
There is a lot about how the wiring works that I don't understand, and I don't know how to test this.

### DIFF
--- a/refinery/tool_manager/tasks.py
+++ b/refinery/tool_manager/tasks.py
@@ -5,7 +5,7 @@ from django_docker_engine.docker_utils import DockerClientWrapper
 
 
 @task()
-def docker_garbage_collection():
+def django_docker_cleanup():
     # TODO: Specify manager, if not default
     client = DockerClientWrapper()
     client.purge_inactive(settings.DJANGO_DOCKER_ENGINE_SECONDS_INACTIVE)
--- a/refinery/tool_manager/tasks.py
+++ b/refinery/tool_manager/tasks.py
@@ -1,22 +1,11 @@
-import logging
-
 from django.conf import settings
 
-from celery.task import on_after_configure, task
+from celery.task import task
 from django_docker_engine.docker_utils import DockerClientWrapper
-
-logger = logging.getLogger(__name__)
-
-
-# TODO?
-@on_after_configure.connect
-def setup_periodic_tasks(sender, **kwargs):
-    sender.add_periodic_task(docker_garbage_collection.s(
-            settings.DJANGO_DOCKER_ENGINE_SECONDS_INACTIVE
-    ))
 
 
 @task()
-def docker_garbage_collection(seconds_inactive):
-    client = DockerClientWrapper()  # TODO: manager= ?
-    client.purge_inactive(seconds_inactive)
+def docker_garbage_collection():
+    # TODO: Specify manager, if not default
+    client = DockerClientWrapper()
+    client.purge_inactive(settings.DJANGO_DOCKER_ENGINE_SECONDS_INACTIVE)
--- a/refinery/tool_manager/tasks.py
+++ b/refinery/tool_manager/tasks.py
@@ -1,0 +1,22 @@
+import logging
+
+from django.conf import settings
+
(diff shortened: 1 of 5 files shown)


refinery/tool_manager/tasks.py
```diff
@@ -0,0 +1,22 @@
...
+from celery.task import on_after_configure, task
+from django_docker_engine.docker_utils import DockerClientWrapper
+
+logger = logging.getLogger(__name__)
+
+
+# TODO?
+@on_after_configure.connect
+def setup_periodic_tasks(sender, **kwargs):
+    sender.add_periodic_task(docker_garbage_collection.s(
```
scottx611x: s -> subtask

mccalluc:
@scottx611x : I was reading the wrong version of the docs. I am pretty sure the event is being called repeatedly... less sure if the cleanup is actually happening.

```
[2017-09-08 19:14:29,632: INFO/MainProcess] beat: Starting...
[2017-09-08 19:14:37,859: INFO/MainProcess] Scheduler: Sending due task gc-docker-containers (tool_manager.tasks.docker_garbage_collection)
[2017-09-08 19:15:07,878: INFO/MainProcess] Scheduler: Sending due task gc-docker-containers (tool_manager.tasks.docker_garbage_collection)
[2017-09-08 19:15:37,905: INFO/MainProcess] Scheduler: Sending due task gc-docker-containers (tool_manager.tasks.docker_garbage_collection)
```

mccalluc:
And I have seen it kill a docker instance. Not sure that I'm putting the new configuration stuff in the right place.

mccalluc:
Passed for Scott: On Travis, fails at:
```
tool = VisualizationTool.objects.all()[0]
```

mccalluc:
@scottx611x : Assuming tests pass now, ok to merge?

refinery/tool_manager/tasks.py
```diff
@@ -0,0 +1,11 @@
+from django.conf import settings
+
+from celery.task import task
+from django_docker_engine.docker_utils import DockerClientWrapper
+
+
+@task()
+def docker_garbage_collection():
```
scottx611x: Maybe rename this per Ilya's suggestion, but looks good to me otherwise

refinery/tool_manager/utils.py
```diff
@@ -468,14 +468,18 @@ def parse_file_relationship_nesting(nested_structure, nesting_dict=None,
         raise RuntimeError(
             "LIST/PAIR structure is not balanced {}".format(nesting_contents)
         )
+    # TODO: If it's a unicode instead of str, it breaks. Too fragile?
```
scottx611x: I'm fine with adding `unicode` to this check

refinery/tool_manager/tests.py
```diff
@@ -2665,8 +2660,7 @@ def test_invalid_TLC_non_LIST_PAIR_file_relationships(self):
         with self.assertRaises(RuntimeError) as context:
             create_tool(tool_launch_configuration, self.user)
         self.assertIn(
-            "The `file_relationships` defined didn't yield a valid "
```
scottx611x: Could we formulate this string to be more in line with 
```
"The 'file_relationships' {} type {} is not a valid LIST/PAIR nesting: {}".format(nesting_contents, nesting_types, valid_types)
```

refinery/tool_manager/utils.py
mccalluc: Filed #2108: Hopefully not hard, but I could imagine it blowing up.

refinery/tool_manager/tasks.py
mccalluc: renamed.

scottx611x: @mccalluc  Renamed to what?

refinery/tool_manager/tests.py
mccalluc: I'd prefer not to get into testing the precise wording of exceptions: too fragile.

PR #2110, opened 2017-09-11
Mccalluc/limit docker instances

mccalluc:
I think at this point the concern is only with imposing a limit on the container count. Now is not the time to figure out how errors from the API should be reported, and how the front end should handle them.

### DIFF
--- a/refinery/tool_manager/models.py
+++ b/refinery/tool_manager/models.py
@@ -393,7 +393,7 @@ class Tool(OwnableResource):
         self.set_tool_launch_config(tool_launch_config)
 
 
-class VisualizationToolError(Exception):
+class VisualizationToolError(StandardError):
     pass
 
 
@@ -418,8 +418,8 @@ class VisualizationTool(Tool):
             - <HttpResponseBadRequest>, <HttpServerError>
         """
         client = DockerClientWrapper()
-        max = settings.DJANGO_DOCKER_ENGINE_MAX_CONTAINERS
-        if len(client.list()) >= max:
+        max_containers = settings.DJANGO_DOCKER_ENGINE_MAX_CONTAINERS
+        if len(client.list()) >= max_containers:
             raise VisualizationToolError('Max containers')
         container = DockerContainerSpec(
             image_name=self.tool_definition.image_name,
--- a/refinery/tool_manager/models.py
+++ b/refinery/tool_manager/models.py
@@ -413,6 +413,11 @@ class VisualizationTool(Tool):
         launched container's url
             - <HttpResponseBadRequest>, <HttpServerError>
         """
+        client = DockerClientWrapper()
+        max = settings.DJANGO_DOCKER_ENGINE_MAX_CONTAINERS
+        if max and len(client.list()) >= max:
+            logger.warn('Already have max containers; will not start another')
+            return JsonResponse({})  # TODO: How to report error?
         container = DockerContainerSpec(
             image_name=self.tool_definition.image_name,
             container_name=self.container_name,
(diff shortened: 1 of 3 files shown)


refinery/tool_manager/models.py
```diff
@@ -413,6 +417,10 @@ def launch(self):
         launched container's url
             - <HttpResponseBadRequest>, <HttpServerError>
         """
+        client = DockerClientWrapper()
+        max = settings.DJANGO_DOCKER_ENGINE_MAX_CONTAINERS
```
scottx611x: could be confusing with Python's `max()` builtin?

refinery/tool_manager/tests.py
```diff
@@ -2518,19 +2519,22 @@ def _start_visualization(
...
             self.assertEqual(post_response.status_code, 200)
 
-            tool = VisualizationTool.objects.get(
+            tools = VisualizationTool.objects.filter(
                 tool_definition__uuid=self.td.uuid
             )
-            self.assertEqual(tool.get_owner(), self.user)
+            if count:
+                self.assertEqual(len(tools), count)
+            last_tool = tools[len(tools)-1]
```
scottx611x: Shorthand for this would be: `tools[:-1]`

refinery/tool_manager/models.py
```diff
@@ -393,6 +393,10 @@ def update_file_relationships_with_urls(self):
         self.set_tool_launch_config(tool_launch_config)
 
 
+class VisualizationToolError(Exception):
```
scottx611x: If we are to have a custom exception, might want to inherit from a less-broad exception class

refinery/tool_manager/tests.py
mccalluc: not quite:
```
>>> a = [1,2,3]
>>> a[:-1]
[1, 2]
>>> a[-1:]
[3]
>>> a[-1:][0]
3
>>> a[-1]
3
```
But [alas](https://docs.djangoproject.com/en/1.11/topics/db/queries/#limiting-querysets):
> Negative indexing (i.e. Entry.objects.all()[-1]) is not supported

refinery/tool_manager/models.py
mccalluc: good thought

refinery/tool_manager/tests.py
hackdna: [last()](https://docs.djangoproject.com/en/1.7/ref/models/querysets/#last) returns the last object in the queryset

scottx611x: @mccalluc Oops, my mistake w/ the shorthand comment.

Regarding the `alas` though, one could do: `Entry.objects.last()`

mccalluc:
Push failed with:
```
FAIL: test_invalid_TLC_non_LIST_PAIR_file_relationships (tool_manager.tests.ToolLaunchConfigurationTests)

  File "/home/travis/build/refinery-platform/refinery-platform/refinery/tool_manager/tests.py", line 2684, in test_invalid_TLC_non_LIST_PAIR_file_relationships

    context.exception.message

AssertionError: "The `file_relationships` defined didn't yield a valid LIST/PAIR nesting." not found in 'The \'file_relationships\' [{"Dicts aren\'t": \'LIST/PAIR-like\'}, {"Dicts aren\'t": \'LIST/PAIR-like\'}] type set([<type \'dict\'>]) is not a valid LIST/PAIR nesting: [set([<type \'list\'>]), set([<type \'tuple\'>])]'
```

Merge issue? Will look at right now.

PR #2111, opened 2017-09-11
Make _mkdir() more robust

hackdna:


refinery/file_store/models.py
```diff
@@ -42,22 +42,19 @@ def _mkdir(path):
...
-    :returns: bool -- True if directory was created, False if it wasn't.
     """
-    logger.debug("Creating directory '%s'", path)
     try:
         os.mkdir(path)
-    except OSError as e:
-        logger.error("Error creating directory '%s': %s", path, e)
+    except OSError as exc:
+        if exc.errno != errno.EEXIST or not os.path.isdir(path):
+            logger.error("Error creating directory '%s': %s", path, exc)
```
mccalluc: Maybe keep just `os.path.isdir(path)`? What behavior do you want if there's a file at that path, but it's not a directory?

And this is not a change from the current state, but if it does fail, it would probably be more useful to throw the exception and exit rather than just log, continue, and fail in a weirder way down the road?

hackdna: The goal is to create the directory if it doesn't exist. Report an error if there was a problem (including a file at that path). I'll raise an exception on failure.

PR #2112, opened 2017-09-12
Hit this limit in testing, just bumping `max_length` value

scottx611x:


scottx611x:
@mccalluc this is something that is variable based on the Galaxy workflow construction, and I figured that if I was able to make a workflow that hit this limit, one of our users could too.

PR #2115, opened 2017-09-12
Mccalluc/sort filter facets on file browser

### LINKED ISSUE #336: allow searching in facet values
Either search on client with type ahead or use solr to do a more thorough search.

Pivotal Tracker story 40911635 (Nils Gehlenborg - Dec 9, 2012)

### LINKED ISSUE #337: allow lexicographical or alphanumeric sorting of facet attribute values
K9 before K36 if possible. Sorting could be done on client or server.

Pivotal Tracker story 40911623 (Nils Gehlenborg - Dec 9, 2012)

mccalluc:
Fix #337. Fix #336.

- alphabetic sorting is normal out of the box sorting: "A9" is after "A10". Doing it the right way (zero padding any numbers in strings) would get a lot harder: We'd have to find a place to keep those padded strings, and I'm not sure right now how that would fit in the DOM.
- With fold/unfold - check/uncheck - match/no-match - search/no-search there are a lot of possible states, and it's not unfolding for searches the way I'd like... but I'm also not so sure that searching facet values is even as important when we're just looking at one dataset?
- I'm also noticing that the behavior to show selected facets, even when folded up, is not what I've done in the cross-dataset browser.

### DIFF
--- a/refinery/ui/source/js/file-browser/directives/assay-filters.js
+++ b/refinery/ui/source/js/file-browser/directives/assay-filters.js
@@ -28,6 +28,14 @@
       controller: 'AssayFiltersCtrl',
       controllerAs: 'ASCtrl',
       link: function (scope, element, attrs, ctrl) {
+        scope.isDropped = function (attributeName) {
+          var escapeAttributeName = attributeName.replace(/ /g, '-');
+          var attributeTitle = angular.element(
+            document.querySelector('#attribute-panel-' + escapeAttributeName)
+          );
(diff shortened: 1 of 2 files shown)


mccalluc:
@scottx611x: Thinking about it, there's enough going on in the UI that I'd like the approval to be contingent on someone (other than me) seeing it in the UI, and being sure it does the right thing.

mccalluc:
Going to go ahead and merge: At this point it seems reasonably low risk.

PR #2116, opened 2017-09-12
Still arbitrary, but probably a better default?

mccalluc:


mccalluc:
I'm going to close this in favor of #2169 which does the same thing on the release branch, and at some point in the future the release branch will be merged back to develop.

PR #2118, opened 2017-09-12
Clean up display of FB buttons

mccalluc:
This is subjective, but... before:
screen shot 2017-09-12 at 1 54 56 pm

after:
screen shot 2017-09-12 at 1 52 10 pm

- arrow points in the direction the panel will expand
- button size consistency
- Remove "Config" text to avoid wrap

Other thoughts:
- Maybe just "Tool Panel"?
- The popover is really not that useful. Maybe just get rid of it, and think about whether the modal itself is as clear as it can be?
> **Select File Display Mode**
> The collection of files in this data set can be displayed using different representations.

### DIFF
--- a/refinery/ui/source/js/file-browser/views/files-tab.html
+++ b/refinery/ui/source/js/file-browser/views/files-tab.html
@@ -10,23 +10,25 @@
       <div class="row"
         ng-if="FBCtrl.isOwner || FBCtrl.canEdit"
         id="tool-panel-show-button">
+
         <button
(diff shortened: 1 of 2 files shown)


mccalluc:
@jkmarx : Do you feel one way or the other about this? And do you have thoughts on the other notes?:
> - Maybe just "Tool Panel"?
> - The popover is really not that useful. Maybe just get rid of it, and think about whether the modal itself is as clear as it can be?

Issue #2126, opened 2017-09-12
Satori never loads if data sets loads first?


mccalluc:
* Specific code commit: 2a3c825183c423d388c391f52a18e13d5caa2925
* Version of the web browser and OS: ff macos + chome (but more often in Chrome?)
* Environment where the error occurred: test production

### Steps to reproduce

I do not have a reproducer, but if I repeatedly reload https://test.stemcellcommons.org/#/exploration?context=OWL:Thing&branchId=0 or https://test.stemcellcommons.org/#/exploration there are times when the half of the page for Satori never loads. I have never seen Satori load if the data sets list renders first. Right now, this seems to happen more often for Chrome than FF, but I have seen it on both.

This seems like a race condition: I could imagine that the right and left parts of the screen may have different characteristics under load, because they are hitting different backends? ... So it might be easy to reproduce at one time, but more difficult at another (ie, if Neo4j is faster, and SQL is slower).

flekschas:
Yeah this is most likely a startup issue. I had to give more time to the initial loading for FF to get satori.cloud..refinery-platform.org loaded consistently. The entire page startup time is pretty slow and unfortunately not super robust.

I've basically increase it to 10 instead of 5 trials on the following file:
https://github.com/refinery-platform/refinery-platform/blob/develop/refinery/ui/source/js/dashboard/services/width-fixer.js#L53

This alleviates but does not fix the problem. To fix this one would need to re-structure the frontend more drastically to make sure that all the data is in place before rendering.

Hope it helps.

mccalluc:
My sense is that 
- this isn't a regression
- anyone can make a PR to bump it to 10
- it doesn't need to be in 1.6?
- and we should probably keep an issue open for finding the "right" way to do it?

Are there any downsides to increasing it to 10? Does it slow down other parts of the page load while we retry?

PR #2127, opened 2017-09-13
Increase wait to avoid Satori load race condition

mccalluc:
Towards #2126, but not the best long-term solution.


mccalluc:
@flekschas : I should have cc'd you earlier. I'll go ahead and merge, but let us know if this isn't what you wanted, or isn't sufficient.

PR #2131, opened 2017-09-13
Reduce max memory for Solr on Vagrant

hackdna:


deployment/refinery-modules/refinery/manifests/solr.pp
```diff
@@ -39,10 +39,19 @@
...
-    file_line { "solr_config_home":
-      path  => "/var/solr/solr.in.sh",
-      line  => "SOLR_HOME=${django_root}/solr",
-      match => "^SOLR_HOME",
+    file_line {
+      "solr_config_home":
+        path  => "/var/solr/solr.in.sh",
+        line  => "SOLR_HOME=${django_root}/solr",
+        match => "^SOLR_HOME";
+      'solr_config_heap':
```
scottx611x: Double quotes for consistency?

deployment/refinery-modules/refinery/manifests/solr.pp
```diff
@@ -39,10 +39,19 @@
...
+        path    => '/var/solr/solr.in.sh',
+        line    => 'SOLR_HEAP="32m"',
+        match   => 'SOLR_HEAP=',
+        replace => $::deployment_platform ? {
```
scottx611x: This is cool, had no clue we could do that

deployment/refinery-modules/refinery/manifests/solr.pp
hackdna: Actually, they should all be single quotes unless variable interpolation required :) https://docs.puppet.com/puppet/5.1/style_guide.html#quoting

scottx611x: Okay cool. Could you update the other double quotes w/o interpolated vars?

hackdna: In all the manifests? :)

scottx611x: If you're feeling up to it I guess? But I was just thinking about ones from this diff.

PR #2172, opened 2017-09-19
Use the latest django-docker

mccalluc:
Use https://github.com/refinery-platform/django_docker_engine/pull/79: django-docker temp dirs should be cleaned up, and we can configure them to be stored under `/data`.

### DIFF
--- a/refinery/config/settings/base.py
+++ b/refinery/config/settings/base.py
@@ -658,7 +658,8 @@ DJANGO_DOCKER_ENGINE_BASE_URL = "visualizations"
 # Time in seconds to wait before killing unused visualization
 DJANGO_DOCKER_ENGINE_SECONDS_INACTIVE = 60 * 60
 # Location of DjangoDockerEngine proxy logging
-PROXY_LOG = '/tmp/django_docker_engine.log'
+DJANGO_DOCKER_ENGINE_DATA_DIR = '/data/django-docker-engine-data'
+PROXY_LOG = '/data/django-docker-engine.log'
 
 REFINERY_DEPLOYMENT_PLATFORM = "vagrant"
 
--- a/refinery/tool_manager/urls.py
+++ b/refinery/tool_manager/urls.py
@@ -18,6 +18,7 @@ with open(please_wait_path, 'r') as f:
     please_wait_content = f.read()
 
 url_patterns = Proxy(
+    settings.DJANGO_DOCKER_ENGINE_DATA_DIR,
     logger=FileLogger(settings.PROXY_LOG),
     please_wait_content=please_wait_content
 ).url_patterns()
--- a/requirements.txt
+++ b/requirements.txt
@@ -79,4 +79,4 @@ watchdog==0.6.0
 xmltodict==0.9.2
 yolk==0.4.1
 
-git+https://github.com/refinery-platform/django_docker_engine.git@v0.0.20
+git+https://github.com/refinery-platform/django_docker_engine.git@v0.0.21


refinery/config/settings/base.py
```diff
@@ -658,7 +658,8 @@ def get_setting(name, settings=local_settings, default=None):
 # Time in seconds to wait before killing unused visualization
 DJANGO_DOCKER_ENGINE_SECONDS_INACTIVE = 60 * 60
 # Location of DjangoDockerEngine proxy logging
-PROXY_LOG = '/tmp/django_docker_engine.log'
+DJANGO_DOCKER_ENGINE_DATA_DIR = '/data/django-docker-engine-data'
```
hackdna: `/data` doesn't exist in Vagrant VM and will go away on AWS once Refinery switches to storing data files in S3

refinery/config/settings/base.py
```diff
@@ -658,7 +658,8 @@ def get_setting(name, settings=local_settings, default=None):
...
+PROXY_LOG = '/data/django-docker-engine.log'
```
hackdna: It's probably best to follow [standard Django logging practices](https://docs.djangoproject.com/en/1.7/topics/logging/) instead of restricting logging to just a file.

refinery/config/settings/base.py
mccalluc: This is where docker instances will have mounted volumes for their inputs and ephemeral storage needs. The path is accessed by docker-engine: If docker-engine is running on a separate instance, it will be a directory on that instance. If we can use ECS, then this may be moot. But it needs to be somewhere for now, and this is better than having it hard-coded to a path on the root volume.

Using `/data` also seemed better than mounting an entirely new EBS volume.

On vagrant, do you have a strong preference about whether we create a `/data` directory as part of setup, or have this setting depend on environment, or something else entirely?

refinery/config/settings/base.py
mccalluc: Good point: I think there was some reason initially why it didn't fit, but I can't recall it now, and I'm not sure the reason would be valid even if I could. https://github.com/refinery-platform/django_docker_engine/issues/80

refinery/config/settings/base.py
hackdna: OK then it's fine to use '/data' for Docker files for now on a condition that only django_docker_engine would have access to it to avoid tight coupling with the rest of the application. Also, django_docker_engine should consider this storage ephemeral because this is going to be the case when it runs on a separate instance.

It's probably a good idea to add `/data` to Vagrant for consistency and move all dirs from $project_root (`media`, `import`, `isa-tab` and Solr index dirs) there.

Finally, it'd be great to consolidate all storage related Puppet resources into one module (similar to [python.pp](https://github.com/refinery-platform/refinery-platform/blob/mccalluc/django-docker-data-dir/deployment/refinery-modules/refinery/manifests/python.pp) for example).

mccalluc:
At least locally, `/data` does seem to be created, and the first test that fails on travis because of the missing directory passes:
```
(refinery-platform)vagrant@refinery:/vagrant/refinery$ ./manage.py test tool_manager.tests.ToolAPITests.test_bad_extra_directories_path_with_rollback
2017-09-20 11:20:05 DEBUG    pyvirtualdisplay:10 <module>() - version=0.2.1
Creating test database for alias 'default'...
2017-09-20 11:20:58 DEBUG    tool_manager.utils:158 create_tool_definition() - Pulling Docker image: nginx:1.10.3-alpine
2017-09-20 11:21:00 DEBUG    tool_manager.views:73 create() - Successfully created Tool: Test LIST Visualization bad extra_directories-launch
2017-09-20 11:21:00 ERROR    tool_manager.views:76 create() - Specified path: `not_an_absolute_path` is not absolute
.
Ran 1 test in 3.110s

OK
```
vs [travis](https://travis-ci.org/refinery-platform/refinery-platform/builds/277594763#L7309):
```
FAIL: test_bad_extra_directories_path_with_rollback (tool_manager.tests.ToolAPITests)

  File "/home/travis/build/refinery-platform/refinery-platform/refinery/tool_manager/tests.py", line 2182, in test_bad_extra_directories_path_with_rollback

    'Specified path: `not_an_absolute_path` is not absolute'

AssertionError: "could not create '/data': Permission denied" != 'Specified path: `not_an_absolute_path` is not absolute'
```
- Check if something in the larger test is causing problems?
- Check whether provisioning on Travis creates `/data` or not

deployment/manifests/default.pp
```diff
@@ -31,6 +31,14 @@
...
 }
 
+# Django-docker-engine needs a place for ephemeral data.
+# In production, this is a separate EBS mount, so we don't need to create it locally.
+file { '/data':
+  ensure => 'directory',
+  owner  => $app_user,
+  group  => $app_group
+}
+
```
hackdna: This should be combined into a separate module with storage related items in [aws.pp](https://github.com/refinery-platform/refinery-platform/blob/mccalluc/django-docker-data-dir/deployment/refinery-modules/refinery/manifests/aws.pp).

mccalluc:
On [travis](https://travis-ci.org/refinery-platform/refinery-platform/builds/277594763#L7309):
```
FAIL: test_bad_extra_directories_path_with_rollback (tool_manager.tests.ToolAPITests)
  File "/home/travis/build/refinery-platform/refinery-platform/refinery/tool_manager/tests.py", line 2182, in test_bad_extra_directories_path_with_rollback
    'Specified path: `not_an_absolute_path` is not absolute'
AssertionError: "could not create '/data': Permission denied" != 'Specified path: `not_an_absolute_path` is not absolute'
```

Locally, `/data` is created, and the test passes:
```
(refinery-platform)vagrant@refinery:/vagrant/refinery$ ./manage.py test tool_manager.tests.ToolAPITests.test_bad_extra_directories_path_with_rollback
2017-09-20 11:20:05 DEBUG    pyvirtualdisplay:10 <module>() - version=0.2.1
Creating test database for alias 'default'...
2017-09-20 11:20:58 DEBUG    tool_manager.utils:158 create_tool_definition() - Pulling Docker image: nginx:1.10.3-alpine
2017-09-20 11:21:00 DEBUG    tool_manager.views:73 create() - Successfully created Tool: Test LIST Visualization bad extra_directories-launch
2017-09-20 11:21:00 ERROR    tool_manager.views:76 create() - Specified path: `not_an_absolute_path` is not absolute
.
Ran 1 test in 3.110s

OK
```
- Why does travis seem not to have `/data`?
- or does the larger test suite make a difference?

hackdna:
I don't think Travis allows creating dir in `/`, so perhaps Docker ephemeral files should use [system temp storage](https://docs.python.org/2/library/tempfile.html) instead for simplicity. This means increasing the root volume size on AWS temporarily (to be reduced when Docker runs on a separate instance).

hackdna:
Actually, my suggestions should be quick to implement. I could help with the Puppet code if needed.

deployment/manifests/default.pp
scottx611x: @hackdna Are you saying create a `storage.pp`? 
Would this be better suited for an issue outside of this pr?

mccalluc:
The proposed refactoring in #2175 isn't working right now. Talked with Scott and decided the best thing for where we are now is to merge this as is.

PR #2179, opened 2017-09-21
With this, I can get to the old file browser

### LINKED ISSUE #2083: Old file browser is broken
- https://test.stemcellcommons.org/provenance/dc0bdaa9-eb00-4162-9567-309c7f3b3de9/#/files/browse brings up an empty pane
- Details tab can not even be clicked on

This page needs to hang around for the provenance visualization, and other functionality will be removed, gradually, but it seems broken right now.

mccalluc:
TODO:
- add a link from new file browser
- add tests
- strip it down

screen shot 2017-09-21 at 1 09 32 pm

Fixes #2083
Fixes #2093


mccalluc:
@ngehlenborg :

screen shot 2017-09-21 at 4 08 54 pm

If we can, I would prefer *not* to put this in 1.6.0, because it's one more set of changes than could break things in unexpected ways.

I hope this might be sufficient for 1.6.1, though I know that ideally it would be a tab on the same view, and ideally more vestigial code would be removed, but I wanted to keep this PR as close to the minimal set of changes necessary to get an unencumbered provvis as possible.

If we do want more in this direction, I think the next steps should be to pull out provvis, rather that to worry first about stripping the extraneous code.

mccalluc:
I think this is good enough, either for 1.6.0, or, preferably, 1.6.1. There's a whole lot of clean up that could be done, but that's probably not the priority.

PR #2175, opened 2017-09-20
Simplify data dir management

hackdna:

Issue #1275, opened 2016-07-26
DataSet2: Bug with Filters and Complement Nodes


jkmarx:
Commit:
11e2117941dcb2631123fab378716610e0f1a6fb
### Steps to reproduce
1. Locally, in the dataset 2 select a filter
2. Click the select all box
3. Deselect a few rows

### Observed behavior
The API registers nodes from all the selected nodes, not just the filters.

### Expected behavior
Nodes should be only selected from the filtered set.

jkmarx:
- [ ] Update Node Group API Post and Put to accept attribute_filters
- [ ] Update front end to send attribute_filters ```{ attribute_solr_name: [field_name]}```

PR #2187, opened 2017-09-22
Fix 2184

### LINKED ISSUE #2184: Launching Visualization tools fails
* Specific code commit: 6003a4a7bb2f4591ee7042d606865837bd44499a
* Version of the web browser and OS: Chrome/FF macOS Sierra
* Environment where the error occurred: `test.stemcellcommons.org` 

### Steps to reproduce

1. Launch a `VisualizationTool`

### Observed behavior
* 400 Error stating: `"could not create '/data/django-docker-engine-data': Permission denied"`

### Expected behavior
screen shot 2017-09-22 at 2 14 02 pm

scottx611x:
- Fixes #2184

### DIFF
--- a/deployment/manifests/default.pp
+++ b/deployment/manifests/default.pp
@@ -16,6 +16,8 @@ $django_settings_module = "config.settings.${conf_mode}"
 $ui_app_root = "${django_root}/ui"
 $site_name = "Refinery"
 $site_url = "192.168.50.50:8000"
+$data_dir = "/data"
+$django_docker_engine_data_dir = "${data_dir}/django-docker-engine-data"
 
 # to make logs easier to read
 class { 'timezone':
@@ -33,7 +35,7 @@ exec { "activate_user":
 
 # Django-docker-engine needs a place for ephemeral data.
 # In production, this is a separate EBS mount, so we don't need to create it locally.
-file { '/data':
+file { $::data_dir:
   ensure => 'directory',
   owner  => $app_user,
   group  => $app_group
--- a/deployment/manifests/default.pp
+++ b/deployment/manifests/default.pp
@@ -16,6 +16,7 @@ $django_settings_module = "config.settings.${conf_mode}"
 $ui_app_root = "${django_root}/ui"
 $site_name = "Refinery"
 $site_url = "192.168.50.50:8000"
+$django_docker_engine_data_dir = "/data/django-docker-engine-data"
 
 # to make logs easier to read
 class { 'timezone':
--- a/deployment/manifests/default.pp
+++ b/deployment/manifests/default.pp
@@ -16,7 +16,6 @@ $django_settings_module = "config.settings.${conf_mode}"
 $ui_app_root = "${django_root}/ui"
 $site_name = "Refinery"
 $site_url = "192.168.50.50:8000"
-$django_docker_engine_data_dir = "/data/django-docker-engine-data"
 
 # to make logs easier to read
 class { 'timezone':
--- a/deployment/manifests/default.pp
+++ b/deployment/manifests/default.pp
@@ -36,9 +36,16 @@ exec { "activate_user":
 # Django-docker-engine needs a place for ephemeral data.
 # In production, this is a separate EBS mount, so we don't need to create it locally.
 file { $::data_dir:
-  ensure => 'directory',
+  ensure => directory,
   owner  => $app_user,
   group  => $app_group
+}
+->
+file { $::django_docker_engine_data_dir:
+    ensure => directory,
+    owner => "$app_user",
+    group => "$app_user",
+    mode => "0755",
 }
 
 # See code in refinery-modules/refinery/...
--- a/deployment/refinery-modules/refinery/manifests/aws.pp
+++ b/deployment/refinery-modules/refinery/manifests/aws.pp
@@ -69,7 +69,13 @@ file { "$solr_core_data":
   group => "$app_user",
   mode => "0755",
 }
-
+->
+file { $::django_docker_engine_data_dir:
+    ensure => directory,
+    owner => "$app_user",
+    group => "$app_user",
+    mode => "0755",
+}
 
 exec { "generate_superuser_json":
   command     => "${virtualenv}/bin/python /srv/refinery-platform/deployment/bin/generate-superuser > /srv/refinery-platform/refinery/core/fixtures/superuser.json.new",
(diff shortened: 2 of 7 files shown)


deployment/refinery-modules/refinery/manifests/aws.pp
```diff
@@ -56,6 +56,13 @@
   before => Exec["solr_install"],
 }
 ->
+file { "$django_docker_engine_data_dir":
```
hackdna: `$django_docker_engine_data_dir` is a global, so should be referenced as `$::django_docker_engine_data_dir`. Also, how will this directory be created on Vagrant?

deployment/manifests/default.pp
```diff
@@ -16,6 +16,7 @@
 $ui_app_root = "${django_root}/ui"
 $site_name = "Refinery"
 $site_url = "192.168.50.50:8000"
+$django_docker_engine_data_dir = "/data/django-docker-engine-data"
```
hackdna: How will this value be set on AWS?

refinery/config/config.json.erb
```diff
@@ -31,6 +31,7 @@
     }
   },
   "DEFAULT_FROM_EMAIL": "<%= @default_from_email || "webmaster@localhost" %>",
+  "DJANGO_DOCKER_ENGINE_DATA_DIR": "<%= @django_docker_engine_data_dir || "/data" %>",
```
hackdna: Should the default value be `/data/django-docker-engine-data`? Is it even necessary when it's already set in `deployment/manifests/default.pp`?

refinery/tool_manager/tests.py
```diff
@@ -2778,6 +2778,7 @@ def _start_visualization(
             )
             force_authenticate(self.post_request, self.user)
             post_response = self.tools_view(self.post_request)
+            logger.debug("RESPONSE_DATA: %s", post_response.content)
```
hackdna: Any reason `RESPONSE_DATA` is all caps?

scottx611x: @hackdna Not particularly, can change if you'd like

refinery/config/config.json.erb
scottx611x: We had to add the default for Travis builds since the `erb` command doesn't populate this setting We're just creating `/data` for Travis builds, but we should probably create the same dir structure across all envs.

hackdna: Yes, there is a configuration drift between environments already. There are now two different ways of setting `DJANGO_DOCKER_ENGINE_DATA_DIR`: explicitly in Puppet for Vagrant and AWS, and implicitly in this template for Travis. This is confusing and could lead to difficult to troubleshoot bugs later. If Puppet were used to configure Travis then there would be no need for workarounds like this.

refinery/tool_manager/tests.py
hackdna: Yes, I think it's a good idea to keep logging uniform.

refinery/config/config.json.erb
scottx611x: @hackdna Sure, If you wanted to spearhead this that would be great since you are the most knowledgeable abut puppet on the team.

deployment/refinery-modules/refinery/manifests/docker.pp
```diff
@@ -2,4 +2,11 @@
   class { '::docker':
     docker_users => [$::app_user]
   }
+  ->
+  file { $::django_docker_engine_data_dir:
```
hackdna: This could fail because it could be executed before `/data` dir is available.

deployment/refinery-modules/refinery/manifests/aws.pp
```diff
@@ -69,7 +69,13 @@
   group => "$app_user",
   mode => "0755",
 }
-
+->
+file { $::django_docker_engine_data_dir:
```
hackdna: This creates `$django_docker_engine_data_dir` on AWS only. What about on Vagrant?

scottx611x: I don't think we care so much on Vagrant because a temporary directory will be created... @mccalluc Is this true?

deployment/manifests/default.pp
```diff
@@ -33,11 +35,18 @@
...
 # In production, this is a separate EBS mount, so we don't need to create it locally.
-file { '/data':
-  ensure => 'directory',
+file { $::data_dir:
+  ensure => directory,
   owner  => $app_user,
   group  => $app_group
 }
+->
+file { $::django_docker_engine_data_dir:
```
hackdna: This violates DRY because this exact same resource is declared for both AWS and Vagrant manifests.

scottx611x: @hackdna Okay, do you have any ideas as to how I should factor out the similarities? Its great that you are pointing out my faults, but without research on my end I may not be able to make these changes as quickly as you could, and our current release is highly dependent on this fix.

hackdna: Sure, classes are a great way to keep the code DRY and re-usable, for example: https://github.com/refinery-platform/refinery-platform/pull/2175/files#diff-cef6f90f8edfbbb045c18b4f98d5e4a6

scottx611x:
I'm able to launch a containerized HiGlass with these changes on a clean AWS instance and in Vagrant:
screen shot 2017-09-25 at 1 31 39 pm
screen shot 2017-09-25 at 1 47 23 pm

deployment/manifests/default.pp
scottx611x: @hackdna Cool this is great.

I'm just thinking that in the context of the `1.6.0` release we should only be addressing bugs and I'd like to introduce the minimum amount of changes possible to fix #2184.

If I introduced changes like you've done in #2175 wouldn't we theoretically have to re-test everything that we've already done?

Would you considering pinning down working functionality with this PR since #2175 seems like a great base to resolve the DRY-ness?

hackdna: OK. I've already merged changes into PR #1275 from develop that happened after creating the release-1.6.0 branch, so at this point PR #1275 should only be merged back into develop.

Going forward, it would be great to figure out a way to fix bugs without introducing even more technical debt.

scottx611x: @hackdna Thats fine, I was more so wondering if we could address the DRYness here in #1257 since that seems like a perfect place to do so , and it wouldn't require us to test everything for `1.6.0` over again.

hackdna: Sure, as long as you don't mind the increase in technical debt.

deployment/refinery-modules/refinery/manifests/aws.pp
mccalluc: see the discussion on default.pp below.

PR #2195, opened 2017-09-25
Hackdna/static asset delivery fix

hackdna:
Fix for #2192


hackdna:
Yes but we need different configurations for static and media files.

PR #2196, opened 2017-09-26
Updating neo4j is slow and should only be done async

### LINKED ISSUE #2084: Details tab, metadata edit: save button doesn't seem to work
* Specific code commit: 24dd3c679983467e6f75de8a9858b2bfbc33d853

### Steps to reproduce

1. For a data set, bring up the details tab
2. Edit a field, like "description"
3. Try to save

### Observed behavior

Nothing seems to happen on the screen... but if you go to the dashboard it seems to have updated?

### Expected behavior

The input box should go away and the new value should be displayed.

mccalluc:
Fix #2084

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -63,12 +63,12 @@ from galaxy_connector.models import Instance
 import tool_manager
 
 from .utils import (add_or_update_user_to_neo4j, add_read_access_in_neo4j,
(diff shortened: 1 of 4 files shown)


mccalluc:
Scott has made a note to look into neo4j slowdown, but we're happy with this this fix for this bug.

PR #2200, opened 2017-09-26
Update config.yaml.template

hackdna:


deployment/aws-config/config.yaml.template
```diff
@@ -1,99 +1,82 @@
...
 STACK_NAME: "RefineryStack"
-# Prefix for all S3 bucket names
+
+# Prefix for all S3 bucket names (must not be shared among stacks)
 # Constraint: DNS-compliant without periods (".")
 S3_BUCKET_NAME_BASE: "refinery"
 
-# the AWS ARN for the TLS certifcate that you wish to present.
+# Constraint: [\w]+ (must be unique within AWS account)
+COGNITO_IDENTITY_POOL_NAME: "Refinery Platform"
```
mccalluc: `\w` stands on its own, and shouldn't be inside `[]`, but apart from that, the regex would not allow the space in the example -- Which is correct?

Would a default value like `Refinery Platform - CHANGE ME` be ok? Or maybe `XXX` as in `RDS_SNAPSHOT`, just so the thing that requires action stands out.

hackdna: Here's the actual description of the field: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-identitypool.html#cfn-cognito-identitypool-identitypoolname
I suspect [these restrictions](http://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-name.html) apply as well. How would you write a regex for this?

The sample values for `RDS_SNAPSHOT` and `EBS_SNAPSHOT` actually show what those values must look like. On the other hand, `COGNITO_IDENTITY_POOL_NAME` can actually be anything (it doesn't have to start with "Refinery Platform"). Galaxy has a [pretty good example of a config template](https://github.com/galaxyproject/galaxy/blob/dev/config/galaxy.ini.sample) which is what I've tried to follow.

mccalluc: `.{1,128}`, though the second page doesn't give space as a valid character? Not worth worrying about. My sense was that this setting is more likely to need changing than others, and it might be good for it to stand out, but it's fine.

PR #2206, opened 2017-09-28
Please-wait template has been moved into django-docker-engine

### LINKED ISSUE #2105: Move please-wait js to django_docker, and refinery only supplies text and style.
The javascript would work for any application, I think, so it should be moved to the library.

### LINKED ISSUE #2154: Visualization launch fails ~10% of the time?
* Specific code commit: 2a3c825183c423d388c391f52a18e13d5caa2925
* Version of the web browser and OS: ff macos
* Environment where the error occurred: test production

### Observed behavior

I started HiGlass 10 times. Most of the time, I got the please wait, and then HiGlass came up, but for 2, I got 
>Internal Server Error
>The server encountered an internal error or misconfiguration and was unable to complete your request.

Not seeing anything useful in the logs at that point: Try to reproduce locally?

... Did 10 more an hour latter and didn't see any errors

### LINKED ISSUE #2199: Just use pypi instead of git tags for django_docker_proxy versioning

mccalluc:
fix #2199 (push to pypi from travis works wonderfully!), fix #2105 (please-wait template is now in django-docker), and, hopefully, fix #2154 (capture one more exception and reply with please-wait instead of error)


mccalluc:
Close in favor of #2209

Issue #822, opened 2015-12-11
Replace before_script in Travis-CI config with Puppet manifest


hackdna:
To reduce duplication of deployment code

mccalluc:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;jy1C6AEK&#x2F;31-822"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> #822</a>

mccalluc:
(We hope that upgrading the npm used on travis will also fix #1768.)

hackdna:
![](https://github.trello.services/images/mini-trello-icon.png) [Update Travis CI build process to use Puppet manifests](https://trello.com/c/TnUOL8vl/149-update-travis-ci-build-process-to-use-puppet-manifests)

PR #2217, opened 2017-10-09
Scottx611x/separate travis test builds

scottx611x:
Only run selenium tests in PRs in their own parallelized build
- [x] Document how to run [entire test suite locally](https://github.com/refinery-platform/refinery-platform/wiki/Server-side-testing#running-all-tests-selenium--unit)
- [ ] Make sure test coverage is up to par
- [ ] Make sure the selenium test are actually run


.travis.yml
```diff
@@ -13,74 +17,111 @@ python:
...
+matrix:
+  include:
+    - env: GRUNT_TESTS=True
+    - env: DJANGO_SELENIUM_TESTS=True
+    - env: DJANGO_UNIT_TESTS=True
 
-  - pip install -r requirements.txt --quiet
+install:
+  - |
+    if ! test -n "$DJANGO_UNIT_TESTS"; then
```
mccalluc: That works, but I prefer the `[]` syntax which defaults to `-n`:
```
$ THIS_IS_SET=True
$ if [ "$THIS_IS_SET" ]; then echo 'here'; fi
here
$ if [ "$THIS_IS_NOT" ]; then echo 'here'; fi
$ if ! [ "$THIS_IS_SET" ]; then echo 'here'; fi
$ if ! [ "$THIS_IS_NOT" ]; then echo 'here'; fi
here

File: .travis.yml [hunk 2]
```diff
@@ -13,74 +17,111 @@ python:
...
-  - gunzip $TAR.gz
-  - tar -xvf $TAR
-  - chmod a+x geckodriver
-  - export PATH=$PATH:`pwd`
+  - |
+    if test -n "$DJANGO_SELENIUM_TESTS"; then
+      GECKODRIVER_VERSION=v0.17.0
+      TAR=geckodriver-$GECKODRIVER_VERSION-linux64.tar
+      wget https://github.com/mozilla/geckodriver/releases/download/$GECKODRIVER_VERSION/$TAR.gz
+      gunzip $TAR.gz
```
mccalluc: While we're here, do you think we could get Travis to cache geckodriver?

.travis.yml
```diff
@@ -13,74 +17,111 @@ python:
...
+      # Vagrant is not used on travis: This user is just for the database.
+      createuser --createdb --no-superuser --no-createrole vagrant
+      createdb -O vagrant refinery
+      sudo mkdir /data
+      sudo chown $USER /data
+      cd refinery
+      # See http://www.stuartellis.eu/articles/erb/#running-erb-from-the-command-line
+      erb config/config.json.erb > config/config.json
+      python manage.py migrate --noinput
+      cd ../
```
mccalluc: Slight preference for `popd`: If at the top we use `pushd`, and then at some point change the directory structure, we don't need to track down and modify these corresponding lines.

.travis.yml
```diff
@@ -13,74 +17,111 @@ python:
...
-  - npm install --progress false --quiet || ( cat npm-debug.log && false )
-  - bower install --config.interactive=false --quiet
-  - cd ../
 script:
-  - set -e # Any error will cause travis to exit early and report a failure.
-
-  # TODO: Not the same behavior as pre-commit hook.
-  # https://github.com/refinery-platform/refinery-platform/issues/1851
-  - flake8 --exclude=migrations,ui ..
+  - "set -e" # Any error will cause travis to exit early and report a failure.
```
mccalluc: extra quotes?

.travis.yml
```diff
@@ -13,74 +17,111 @@ python:
...
-  - echo 'travis_fold:start:django-tests'
-  - coverage run manage.py test
-  - echo 'travis_fold:end:django-tests'
+  - echo 'travis_fold:start:django-unit-tests'
+  - if test -n "$DJANGO_UNIT_TESTS"; then coverage run manage.py test; fi
+  - echo 'travis_fold:end:django-unit-tests'
+
+  - echo 'travis_fold:start:django-selenium-tests'
+  - if test -n "$DJANGO_SELENIUM_TESTS"; then coverage run manage.py test --pattern="selenium_tests.py"; fi
+  - echo 'travis_fold:end:django-selenium-tests'
```
mccalluc: With the build matrix, only one of these sections is being run, so the `travis_fold` stuff should be inside the `if-then`, if it is retained at all.

.travis.yml
```diff
@@ -13,74 +17,111 @@ python:
...
-  - cd ../
 script:
-  - set -e # Any error will cause travis to exit early and report a failure.
-
-  # TODO: Not the same behavior as pre-commit hook.
-  # https://github.com/refinery-platform/refinery-platform/issues/1851
-  - flake8 --exclude=migrations,ui ..
+  - "set -e" # Any error will cause travis to exit early and report a failure.
+  - |
+    if ! test -n "$GRUNT_TESTS"; then
```
mccalluc: probably just run this under one django test instead of both? Or add a new element to the build matrix?

.travis.yml
```diff
@@ -13,74 +17,111 @@ python:
...
-  - codecov
-  - npm run codecov
+  - |
+    if ! test -n "$GRUNT_TESTS"; then
+      codecov
+    fi
+  - |
+    if test -n "$GRUNT_TESTS"; then
+       npm run codecov
+    fi
```
mccalluc: Bash has `else`: Add that and put the plain `codecov` in it.

.travis.yml
scottx611x: @mccalluc Are you saying just run flake8 under one Django build?

.travis.yml
```diff
@@ -13,74 +17,111 @@ python:
...
+      sudo chown $USER /data
+      cd refinery
+      # See http://www.stuartellis.eu/articles/erb/#running-erb-from-the-command-line
+      erb config/config.json.erb > config/config.json
+      python manage.py migrate --noinput
+      cd ../
+    fi
+  - |
+    if ! test -n "$DJANGO_UNIT_TESTS"; then
+      cd refinery
```
mccalluc: `travis_fold` around the install stuff? Also, I like these util functions:
```
start() { echo travis_fold':'start:$1; echo $1; }
end() { echo travis_fold':'end:$1; }
```

refinery/tool_manager/selenium_tests.py
```diff
@@ -0,0 +1,186 @@
...
+from selenium_testing.utils import (MAX_WAIT, SeleniumTestBaseGeneric,
+                                    wait_until_class_visible)
+from tool_manager.models import Tool, ToolDefinition, VisualizationTool
+from tool_manager.tasks import django_docker_cleanup
+from tool_manager.tests import TEST_DATA_PATH, ToolManagerTestBase
+
+logger = logging.getLogger(__name__)
+
+
+class VisualizationToolLaunchTests(ToolManagerTestBase,
```
mccalluc: This looks like the whole block is being moved from below? Is there anything particular I should look more closely at?

scottx611x: @mccalluc Nope! just moved into its own file

hackdna:
What is the problem that this PR is trying to solve?

scottx611x:
@hackdna 
Speeding up our overall build times, isolating different tests into their own builds for easier debugging, and stopping our single build from hitting its 10,000 line log limit.

.travis.yml
```diff
@@ -13,74 +18,118 @@ python:
...
+  - if ! [ "$DJANGO_UNIT_TESTS" ]; then grunt make; fi
+  - if [ "$GRUNT_TESTS" ]; then grunt test; fi
+  - popd
+  - end $GRUNT
 
-  - echo 'travis_fold:start:grunt'
-  - pushd ui && grunt && popd
-  - echo 'travis_fold:end:grunt'
+  - start $DJANGO_UNIT_TESTS
+  - if [ "$DJANGO_UNIT_TESTS" ]; then coverage run manage.py test; fi
```
hackdna: `manage.py test` runs all tests, including Selenium: https://travis-ci.org/refinery-platform/refinery-platform/jobs/286079823

It would be great to move **all** Selenium tests into a separate folder (`selenium_tests`?) and [name them such that they wouldn't be picked up](https://docs.djangoproject.com/en/1.7/topics/testing/overview/#running-tests) by `manage.py test` command. Then they can be easily run separately by something like `manage.py test -p="selenium_*.py`.

.travis.yml
```diff
@@ -13,74 +18,118 @@ python:
...
   global:
     # These env vars are available to every build
     - PYTHONPATH=$PYTHONPATH:../refinery:../refinery/config
+    - CODECOV=codecov
     - DJANGO_SETTINGS_MODULE=config.settings.prod
+    - FLAKE8=flake8
+    - GRUNT=grunt
+    - SELENIUM_TESTS=selenium-tests
+
+matrix:
```
hackdna: Why not just use a single variable for running different test suites as described here: https://docs.travis-ci.com/user/speeding-up-the-build/
This way there is no need for all the conditionals.

.travis.yml
scottx611x: @hackdna `manage.py test` by default only runs python files matching the pattern `test*.py`, so the selenium tests are not executed by that command now. Only our unit tests are.

In this PR Selenium-based tests have been moved to files named `selenium_tests.py` in their respective Django apps, and we are running them in separate builds ussing the`--pattern` already.

The `selenium_testing` app has tests specific to Refinery UI code, but I felt like it would be odd to include the `tool_manager` ones there as well since we are utilizing Selenium in those to ensure that Refinery plays nicely with containerized visualizations, not testing Refinery UI code

.travis.yml
scottx611x: @hackdna Wouldn't we still need conditionals even if we used a single variable w/ different values? I'm not seeing how we could benefit from that here

.travis.yml
hackdna: I see. This should be documented because it is not obvious why `selenium_testing/tests.py` are not discovered by default.

The purpose of unit testing is to test the smallest testable parts of an application, so they should not have external dependencies and should run as quickly as possible to enable quick development cycle. On the other hand, integration/functional tests are not limited to just testing of the UI code. They test how application components (and their dependencies) work together.

Existing tool manager tests have external dependencies and test several components at once. So, they are actually integration/functional tests and thus should be grouped together with other such tests.

scottx611x: `selenium_testing/tests.py` doesn't exist now. It is now `selenium_testing/selenium_tests.py`

And sure that sounds good, but we have integration tests mixed in with our unit tests throughout our codebase so I'm not sure if that type of change is in the scope of this PR.

.travis.yml
hackdna: This is the whole point of the build matrix.

scottx611x: If you could elaborate and alert me as to how we could utilize the build matrix and not need conditional statements in our `travis.yaml` that would be very helpful

Otherwise I'm still seeing a need for conditional statements

.travis.yml
hackdna: Still, it is not immediately obvious that the name of that file is significant. I would just put a note at the top of it and mention this on the wiki page about testing.

There are not too many of those tests and moving them is straightforward. I think this PR provides a great context for making this change.

.travis.yml
hackdna: Some more examples:
https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix
https://docs.travis-ci.com/user/build-stages/matrix-expansion/

Something like:
```yaml
env:
  - TEST_SUITE="grunt make"
  - TEST_SUITE="manage.py test"
script: $TEST_SUITE
```
"[Travis CI will determine the build matrix based on the environment variables and schedule two builds to run](https://docs.travis-ci.com/user/speeding-up-the-build/#Parallelizing-your-builds-across-virtual-machines)"

scottx611x: Yes, but for example Selenium tests are dependent on grunt operations while Django tests are not. 

So I don't believe this approach fits our use case unless I am mistaken.

hackdna: Does it matter?

scottx611x: I mean, the whole idea behind the conditionals is that we don't run unnecessary code for builds that don't need it. No downloading of `geckodriver` if we don't need to, no compilation of static assets if we don't need to etc.

So I would say that it does matter. Otherwise splitting up the builds doesn't make much sense if we are going to run the majority of the same code in each and just have a different test command to execute.

I'm not convinced that the approach you're suggesting gains us anything, but I'm looking to be swayed.

hackdna: TBH, I'm not even sure if parallelizing the build gains us much while it certainly makes `.travis.yml` a lot more complicated. How much time is actually saved compared to a regular build? If you look at the [build history](https://travis-ci.org/refinery-platform/refinery-platform/builds) the gains are not clear at all.

scottx611x: So [this parallelized build](https://travis-ci.org/refinery-platform/refinery-platform/builds/286079607) took about 12 minutes

While [this non-parallelized build](https://travis-ci.org/refinery-platform/refinery-platform/builds/286116497) took about 19 minutes

The `total time` that Travis reports back for the parallelized builds seems to just be an addition of all the builds time not the actual wall clock time

hackdna: I see. Still, there is a lot of variation in run times: this [regular build](https://travis-ci.org/refinery-platform/refinery-platform/builds/286137728) took 13 mins 39 secs while this [parallelized build](https://travis-ci.org/refinery-platform/refinery-platform/builds/286048105) ran for 13 mins 37 secs.

Also, the time waiting to run a build is not taken into account and that can vary a lot, so benefits are certainly not clear, especially given the additional code complexity. If a quick test run is required, it is very fast to do that locally and can be done at different levels of granularity.

I don't think installing dependencies for each build job would impact the runtime significantly since a lot of them are now cached but it would make `.travis.ci` a lot simpler.

To make a long story short, I think it's worth to parallelize the builds only if it doesn't lead to a significant increase in complexity of the build code (i.e., use a standard build matrix expansion and avoid conditionals).

scottx611x: The parallelized build you referenced was before I split the selenium builds into their own respective builds so that won't be an accurate time estimate.

I don't think the time a build is pending before its run is relevant in this conversation.

I still see benefit in isolating code specific to each build regardless of how its done. For example, if there is a failed migration, and we run our migrations in every parallelized build we see 4 failed builds and then have to dive deeper to figure out what went wrong. I was going to add shell scripts to make `.travis.yml` a bit more tidy looking but even then we have ~5 more things to maintain.

As for code complexity, we are adding some if statements to get general gains of ~40% speed per build. I find that to be an okay tradeoff.

hackdna: The point is that there is large inherent variation in run times that we can't control. Also, the wait time is very relevant because we are interested in the total run time: it often takes the same time (or longer) to wait for a build to start running as it does to run. All this makes the build time savings of a few minutes marginal (see also: http://wiki.c2.com/?ProfileBeforeOptimizing).

Moreover, the build times on Travis should not even matter. It is not a tool for rapid iteration - your local environment is.

Btw, if the build is not parallelized then any error like a failed migration would result in a single failed build :)

Finally, if you really want to make Travis code more robust, it would be more useful to [make it use Puppet manifests](https://github.com/refinery-platform/refinery-platform/issues/822). No need for custom scripts that duplicate existing functionality.

scottx611x:
@hackdna @mccalluc 

So I've done some research and it seems that Travis only allows 5 concurrent builds for open source projects. With that being said, this approach will probably not be beneficial for our team.

Since our Selenium based tests are the main cause of the slowness of our builds it may be a good idea to focus on how to improve those.

We could:
- Run them only on PR builds
- Have two separate builds instead of 4 with one dedicated to Selenium-based tests
- Use something speedier than Selenium
- ... ?

Any ideas are welcome!

hackdna:
I don't see a problem with the current build times. Integration tests are expected to run longer. This is part of the reason why unit tests are useful. However, there are integration tests mixed with unit tests which makes it impossible to run the unit test suite quickly. This is easy to solve by moving all integration tests in one place.

Now there is also a problem with the test output. It is way too verbose. This can be solved by: eliminating tracebacks (tests should not be crashing) and reducing output from grunt uglify task, PhantomJS, etc. (#2224)

Finally, .travis.yml should be using Puppet to avoid custom scripts that duplicate existing functionality but that could be tackled separately. (#822)

.travis.yml
```diff
@@ -32,8 +32,8 @@ matrix:
   include:
```
hackdna: I don't think the matrix section is needed. The TEST_SUITE var can simply be defined in the env section above. https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

mccalluc:
@scottx611x : My sense is that this PR is not the direction we're going to go. Is there anything you want to salvage from it, or can it just be closed?

scottx611x:
@mccalluc Can definitely be closed!

Issue #2218, opened 2017-10-10
Remove config.json before `vagrant up`?


mccalluc:
Something was wrong with my environment, so I just wanted to start over from the top, but rather than getting a new checkout, I started from my existing. ... but that meant I had an old `config.json` in place that was missing a necessary key.

Could `vagrant up` overwrite `config.json` if it already exists, rather than just taking the existing one? Or is this better done by hand? 

@scottx611x : Do you have an opinion?

hackdna:
Historic reference: https://github.com/refinery-platform/refinery-platform/issues/909 and https://github.com/refinery-platform/refinery-platform/issues/785.

mccalluc:
Closing in favor of #909, which is pretty much the same problem.

Issue #909, opened 2016-02-25
config.json is confusing.


drj11:
If a refinery admin has local changes to `config.json`, how do they tell what they need to change when updating refinery?

In normal provisioning, via vagrant, the `config.json` will not be overwritten if it already exists.

drj11:
Suggestion: during provisioning puppet should always render `config.json.erb` to `config.json.sample`. Additionally, if `config.json` does not already exist, it should copy `config.json.sample` to `config.json`.

That way, if there are local changes to `config.json` the admin can compare that file against `config.json.sample`.

drj11:
Maybe it's worse than we/I think. If you continually provision a machine, `config.json` never changes after the first provision.

hackdna:
Should we backup `config.json` and overwrite it or leave it as is and create `config.json.new` ([rpmsave vs rpmnew](http://serverfault.com/a/48819))?

drj11:
I think that would be better, yes. Ideally we'd do something clever, like: if the existing `config.json.sample` and `config.json` match, then just overwrite, otherwise backup and overwrite.

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;dNYokUZJ&#x2F;71-switch-config-json-to-a-format-that-support-comments"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> switch config.json to a format that support comments</a>

hackdna:
![](https://github.trello.services/images/mini-trello-icon.png) [update config.json when deploying code on AWS](https://trello.com/c/qytgHZZk/300-update-config-json-when-deploying-code-on-aws)

mccalluc:
#2219 tried an adhoc approach to this, but Ilya pointed out that [backup attribute of the file resource](https://docs.puppet.com/puppet/3/type.html#file-attribute-backup) would be cleaner.

PR #2219, opened 2017-10-10
Feedback if config.json already exists

mccalluc:
Not sure if the logic is quite right, and would it be possible for it to fail fast? Towards #909 or #2218?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ codekit-config.json
 /deployment/aws-config/tags.yaml
 /refinery/celerybeat-schedule*
 /refinery/config/config.json
+/refinery/config/config.json.prev
 /refinery/static/production/
 /refinery/static/development/
 /refinery/ui/production/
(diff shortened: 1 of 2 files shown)


hackdna:
Do we all agree that we want to change the current behavior to backup and replace existing`config.json` with a new version? If yes, then it is a lot easier to do using the [backup attribute of the file resource](https://docs.puppet.com/puppet/3/type.html#file-attribute-backup).

drj11:
replacing is fine for AWS (preferred, strongly).

Would vagrant users be upset? I thought they might have existing modifications in their `config.json` that they don't want overwritten.

hackdna:
We do immutable deployments on AWS, so config.json is always created from scratch there.

I never actually customize my local config.json. I did run into problems when new settings were introduced, so I think overwriting it on updates would be a good idea (if it's backed up first).

mccalluc:
Ok, so I'll close this, but added a comment to #909 referencing `file-attribute-backup`.

PR #2223, opened 2017-10-10
Mccalluc/provvis regression

mccalluc:
I went a lot farther down this path, but then realized that my edits weren't actually being deployed. In any event, I'm pretty much flying blind here: There is so much functionality behind this, and my only test is whether a rectangle shows up on the screen, and even that is not automated.


scottx611x:
Also you may want to hold off on merging since our outdated version of bower is causing builds to fail. I have a branch right now that will fix it though

PR #2225, opened 2017-10-11
Scottx611x/fix bower

scottx611x:
Bower changed their registry url today w/o a redirect causing our builds to fail. 

We could have manually specified the new registry url in `.bowerrc`, but I figured this was a call for an updated bower version.

### DIFF
--- a/refinery/ui/bower.json
+++ b/refinery/ui/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "refinery-platform-ui",
-  "version": "1.8.2",
+  "version": "1.1.0",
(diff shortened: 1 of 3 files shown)


refinery/ui/bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "refinery-platform-ui",
-  "version": "1.1.0",
```
hackdna: What's the reason?

scottx611x: This change isn't needed, you're right. I'll fix now

PR #2227, opened 2017-10-12
Mccalluc/try cypress

mccalluc:
Definitely a work in progress, but this much passes. One concern would be that it removes selenium coverage without replacing all of it right now. There should also be some documentation for developers. I would also like to turn on the dashboard so we can get a view on the runs inside travis.

I've branched from this branch to continue: https://github.com/refinery-platform/refinery-platform/compare/mccalluc/cypress-continued?expand=1


mccalluc:
This branch was extended in #2239, and some of the Selenium tests did end up coming back. Closing this PR in favor of that one.

Issue #1886, opened 2017-07-10
Links in pop-overs are un-clickable


mccalluc:
* Specific code commit: bc425b3403264b8a0fb1d28cd412a7360e05dc6b
* Version of the web browser and OS: both ff and chrome, macos
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): production, test instance

### Steps to reproduce
Please list all the actions and the input data used:

1. On the dashboard, click on dataset you own to get a preview to fold out
1. Click on "share" above the right pane: This brings up "Permission Editor"
1. Click on the question mark to bring up the "Share a Data Set" popover
1. Click on a link in the popover

or, in file-upload

1. click on the "?" for " Column containing species names or identifiers."
### Observed behavior
* The popover closes, rather than taking you to the appropriate resource

### Expected behavior
* Links take you to the appropriate pages... but maybe it'd be easier just not to have links here?

(Should also review other pop-overs to see if they contain links, and if they are clickable.)

scottx611x:
Unable to reproduce. @mccalluc will check it out again with other browsers.

jkmarx:
I encountered this issue when updating the global analysis popover. The append-to-body option was required, but note, it will move the popover scope out of the parent's scope. I'll double check the permission editor and either fix it or close this issue.

jkmarx:
I checked the permission editor and any popover-html references. I could not reproduce this issue in chrome, firefox, or safari. @mccalluc If you reproduce, please make a giphy and reopen the issue.

jkmarx:
Reference #2231

PR #2231, opened 2017-10-13
 Fix visual bug with tool tip.

### LINKED ISSUE #1884: Same page link on the Collaboration page
Commit: 

While on the Collaboration page click on the "?" next to `Groups` in the first column header.
There is a link in the popover text to the collaboration page (which we are currently on). 
Seems a bit unnecessary.
screen shot 2017-07-07 at 4 41 18 pm

jkmarx:
Ref #1886 
Resolves #1884 
When on the data set share -- ? tooltip, clicking the link 'data_set_preview' doesn't visually do anything since the user is already on the data set preview page.
Update html by adding a component and template for the new logic.

to-collab-page
to-dashboard

### DIFF
--- a/refinery/ui/source/js/commons/ctrls/sharing-help-popover-details-ctrl.js
+++ b/refinery/ui/source/js/commons/ctrls/sharing-help-popover-details-ctrl.js
@@ -20,6 +20,7 @@
     $location
   ) {
     var vm = this;
+    vm.isCollaborationPage = false;
 
    /*
    * ---------------------------------------------------------
@@ -27,8 +28,8 @@
    * ---------------------------------------------------------
    */
     vm.$onInit = function () {
+      // check if the user is on the collaboration page
       vm.isCollaborationPage = $location.search('collaboration').$$search.collaboration;
-      console.log(vm.isCollaborationPage);
     };
   }
 })();
--- a/refinery/ui/source/js/commons/ctrls/sharing-help-popover-details-ctrl.js
+++ b/refinery/ui/source/js/commons/ctrls/sharing-help-popover-details-ctrl.js
@@ -29,7 +29,7 @@
    */
     vm.$onInit = function () {
       // check if the user is on the collaboration page
-      vm.isCollaborationPage = $location.absUrl().indexOf('collaboration') > -1;
+      vm.isCollaborationPage = $location.absUrl().indexOf('/collaboration/') > -1;
     };
   }
 })();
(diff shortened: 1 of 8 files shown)


refinery/ui/source/js/commons/ctrls/sharing-help-popover-details-ctrl.js
```diff
@@ -0,0 +1,35 @@
...
+    vm.isCollaborationPage = false;
+
+   /*
+   * ---------------------------------------------------------
+   * LifeCycle Hooks
+   * ---------------------------------------------------------
+   */
+    vm.$onInit = function () {
+      // check if the user is on the collaboration page
+      vm.isCollaborationPage = $location.absUrl().indexOf('collaboration') > -1;
```
mccalluc: Would it be worth while to be more precise here? `/collaboration/` should begin the path?

jkmarx: I wouldn't imagine so, since the popover is just use on the home and collaboration page. But, I can add it.

refinery/ui/source/js/commons/partials/sharing-help-popover-details.html
```diff
@@ -0,0 +1,15 @@
...
+  with an authenticated user, create a new group or invite the user to an
+  existing group
+  <span ng-if="!$ctrl.isCollaborationPage">
+    <a href='/collaboration'> on the collaboration page. </a> Owners can
+    update data set permissions (including making it public) on the data set
+    preview by clicking the “sharing” button.
+  </span>
+  <span ng-if="$ctrl.isCollaborationPage">
+    on the collaboration page. Owners can update data set permissions
+    (including making it public) on the <a href='/#'> data set preview </a> by
```
mccalluc: Could this be more confusing than useful? We can't actually take them to the data set preview, but if we just take them to the dashboard, there isn't going to be a "sharing" button for them to click either.

jkmarx: @mccalluc Good point. How about this revised text: 
"Only data set owners can change sharing options. To share
  with an authenticated user, create a new group or invite the user to an
  existing group on the collaboration page. Owners can update data set permissions (including making it public) on the <a href=/#> dashboard </a> by opening a data set preview (click a data set link) and then clicking the “sharing” button."

mccalluc: That sounds good.

PR #2234, opened 2017-10-16
Jkmarx/metadata edit enhancement

### LINKED ISSUE #2085: Details tab, metadata edit: input boxes not pre-populated
* Specific code commit: 24dd3c679983467e6f75de8a9858b2bfbc33d853

### Steps to reproduce

1. For a data set, bring up the details tab
2. Edit a field, like "description"
3. Try to save

(Screen should update here, but if it doesn't, that's #2084.)

4. Reload the page
5. Try to edit the field again

### Observed behavior

The input box should be pre-populated with the current value, but it's not.

jkmarx:
- Resolves #2085 
- Adjusts CSS for readability

### DIFF
--- a/refinery/ui/source/js/data-set-about/partials/details.html
+++ b/refinery/ui/source/js/data-set-about/partials/details.html
@@ -12,23 +12,29 @@
   <div
     id="dataSetAccession"
     uib-collapse="ADCtrl.isCollapsed.title">
-    <div class="vertical-align">
+    <div class="vertical-align p-t-1-2">
       <form
         ng-submit="ADCtrl.updateDataSet(
         'accession', ADCtrl.updatedField.accession
         )"
         name="dataSetAccessionForm"
-        class="navbar-form text-align-center float-left">
-      <input
-        autofocus
-        type="text"
-        name="dataSetAccession"
-        class="form-control"
-        placeholder="update accession"
-        ng-model="ADCtrl.updatedField.accession"
-        ng-minlength="3"
-        ng-maxlength="32"
-        required>
+        class="p-r-1-2">
+        <div class="form-group" ng-class='{"has-warning":!dataSetAccessionForm.$valid}'>
+          <label class="control-label" for="dataSetAccessionInput">
+            Edit Accession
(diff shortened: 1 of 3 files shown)


jkmarx:
## New
screen shot 2017-10-16 at 12 37 59 pm


screen shot 2017-10-16 at 12 38 28 pm

## Old
screen shot 2017-10-16 at 12 46 57 pm

refinery/ui/source/js/data-set-about/partials/details.html
```diff
@@ -1,6 +1,10 @@
 <div class="refinery-header">
   <div class="refinery-header-left">
-    <h3>{{ ADCtrl.dataSet.title }}</h3>
+
+    <h3>
+      {{ ADCtrl.dataSet.accession }}<span ng-if="ADCtrl.dataSet.title &&
```
mccalluc: formatting: Just put the span on a line by itself?

refinery/ui/source/js/data-set-about/partials/details.html
```diff
@@ -41,16 +50,18 @@
         type="submit"
         class="btn btn-xs btn-primary"
         ng-click="ADCtrl.updateDataSet(
-        'accession', ADCtrl.updatedField.accession)"
-        ng-disabled="!dataSetAccessionForm.$valid">
+        'accession', ADCtrl.editedDataSet.accession)"
+        ng-disabled="!dataSetAccessionForm.$valid ||
+        ADCtrl.editedDataSet.accession == ADCtrl.dataSet.accession">
```
mccalluc: Shouldn't make a difference here, but should `===` be the default?

jkmarx: @mccalluc https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

Typically for HTML comparisons, I use the abstract == because I want the same value. In forms, if I want validation I will add them for types (ie booleans or integers). In JS code, yes.
Thoughts?

PR #2239, opened 2017-10-16
Mccalluc/cypress continued

mccalluc:
Here's a first look at what testing with cypress would look like. 

Plus:
- The test runner UI for local development
- Their dashboard service, so we can see the results of runs on Travis
- Tests are very easy to write, and run behavior is predictable.

Minus:
- Vendor lock-in
- This will be easiest to use for writing end-to-end integration tests, where we install solr on Travis; Alternatively, API calls can be mocked, but that would make it much harder to write each test.
- It is impossible to mock any of the internals in python: Either we hit an actual server, or we don't.
- Pollution of development environment: Since real objects are created by the tests, they would need to be explicitly cleaned up, as well, if we care about that.

I think this could be useful for us, but I think having more eyes on it before going forward would be good.

---
PS: As of [today](https://www.cypress.io/blog/2017/10/10/cypress-is-now-public-beta/) it's open source.


README.md
```diff
@@ -52,6 +52,12 @@ To pull the latest code and update Refinery installation:
 $ fab vm update
 ```
 
+To run integration tests, hit "Run All Tests" in the Cypress UI after this command:
+```bash
+$ ./node_modules/.bin/cypress run --config baseUrl=http://192.168.50.50:8000
+```
+The results of integration tests can be viewed on the [Cypress Dashboard](https://dashboard.cypress.io/#/projects/odo3mu/runs).
+
```
hackdna: The purpose of README is just to provide quick start instructions on how to run the project. This should be moved to the testing wiki pages. Also, info on how to install Cypress on the host is missing.

refinery/config/settings/prod.py
```diff
@@ -1,11 +1,10 @@
 # custom settings for production environment
 
 from .base import *  # NOQA
+from .base import get_setting
```
hackdna: Already done by the line above.

.travis.yml
```diff
@@ -74,6 +66,18 @@ script:
   - coverage run manage.py test
   - echo 'travis_fold:end:django-tests'
 
+  # TODO: Start up solr (and neo4j?)
+  - echo 'travis_fold:start:cypress'
+  - python manage.py loaddata guest.json
+  - python manage.py activate_user guest
+  - python manage.py add_users_to_public_group
+  - python manage.py collectstatic --noinput -v 0
+  - python manage.py runserver --insecure &
```
hackdna: These five lines should probably be in the install and/or before_script sections. Ideally, Puppet should be used for provisioning to avoid code duplication.

refinery/ui/cypress/fixtures/api-v1-data_sets.json
```diff
@@ -0,0 +1,67 @@
...
+      "creation_date": "2017-10-10T15:13:49.303358",
+      "date": "2017-10-10T15:13:49.726331",
+      "description": "",
+      "file_count": 8,
+      "file_size": 0,
+      "id": 2,
+      "is_owner": false,
+      "is_shared": true,
+      "isa_archive": "2e9b6457-177a-45f7-af1d-0b606d3df488",
+      "isa_archive_url": "http://192.168.50.50:8000/media/file_store/d1/c4/isa_9909_558276.zip",
```
hackdna: Seems brittle. Would this file actually exist on disk?

refinery/ui/cypress/fixtures/api-v1-data_sets.json
```diff
@@ -0,0 +1,67 @@
...
+      "creation_date": "2017-10-10T15:13:49.303358",
+      "date": "2017-10-10T15:13:49.726331",
+      "description": "",
+      "file_count": 8,
+      "file_size": 0,
+      "id": 3,
+      "is_owner": false,
+      "is_shared": true,
+      "isa_archive": "2e9b6457-177a-45f7-af1d-0b606d3df489",
+      "isa_archive_url": "http://192.168.50.50:8000/media/file_store/d1/c4/isa_9909_558277.zip",
```
hackdna: Same as above.

hackdna:
Tests should not leave any artifacts behind (#2242). Also, Cypress is still in beta and it is unclear how much traction it would get which is a risk.

mccalluc:
Still need to revert the Selenium deletion, but after that I think it's ready to merge.

refinery/config/settings/prod.py
```diff
@@ -6,6 +6,4 @@
 TEMPLATE_DEBUG = DEBUG
 
 # Required when DEBUG = False
-ALLOWED_HOSTS = get_setting("ALLOWED_HOSTS")
-
-EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"
```
mccalluc: @scottx611x : Looking at this again, I remember that we did need to add `console.EmailBackend` for testing, but in production I'd think we'd still want this?

hackdna: Yes, both are needed in production.

scottx611x: @hackdna Doesn't the `EMAIL_BACKEND` in `aws.py` override this anyways?

We were wondering if this was even needed in `prod.py` if theres no accompanying SMTP server running by default.

hackdna: Yes. The purpose of a prod config is to run the site with real users which means you need a real SMTP backend to send email.

deployment/manifests/default.pp
```diff
@@ -55,5 +55,4 @@
 include refinery::neo4j
 include refinery::postgresql
 include refinery::python
-include refinery::selenium
```
scottx611x: I think this should be brought back if we're going to keep selenium around and ween ourselves off of it.

refinery/factory_boy/django_model_factories.py
```diff
@@ -16,6 +16,7 @@ class Meta:
     name = "Test DataSet - {}".format(uuid)
     creation_date = datetime.now()
     modification_date = datetime.now()
+    slug = None
```
scottx611x: I think you could already use the factory here to create a Dataset with a `slug` without adding this line.
I could also very well be wrong and would like to know if I am!

refinery/config/settings/prod.py
mccalluc: Talking with Scott, realized that we were probably too fixated on running these tests in production mode. If we we ran it in development mode, the only real difference is that if something did go wrong we'd get the extra debug information, and that's a good thing.

.travis.yml
```diff
@@ -40,7 +44,9 @@ env:
   global:
     # These env vars are available to every build
     - PYTHONPATH=$PYTHONPATH:../refinery:../refinery/config
-    - DJANGO_SETTINGS_MODULE=config.settings.prod
+    - DJANGO_SETTINGS_MODULE=config.settings.dev
```
scottx611x: I was thinking of  something more along the lines of on line 76: 
`python manage.py runserver --insecure --settings=config.settings.dev`
so that the cypress testing is the one off thing and we can still run our units against prod settings

hackdna: FYI, unit tests should be unaffected by Django settings because they should only test small pieces of code in isolation.
Also, `--insecure` is [only needed when DEBUG is False](https://docs.djangoproject.com/en/1.7/ref/contrib/staticfiles/#django-admin-option---insecure).

mccalluc: [core runserver](https://docs.djangoproject.com/en/1.11/ref/django-admin/#django-admin-runserver) only has `--noreload`,  `--nothreading`, `--ipv6` and the option to pick a port. 

[staticfiles runserver](https://docs.djangoproject.com/en/1.11/ref/contrib/staticfiles/#cmdoption-runserver-insecure) adds `--insecure`.

But I could just reset that envar before runserver?

hackdna: There is only one Django dev server. It is invoked by the `runserver` subcommand. If DEBUG is set to True in the settings (e.g., in dev.py) you simply don't need to use `--insecure` (that mode is insecure by default).
Btw, the docs you've linked to are for Django 1.11 while we are running 1.7 - there often big differences between versions.

mccalluc:
@ilya: Thanks for the feedback! I think your concerns have been addressed.

Talked with Jennifer, and she would defer to Scott and me, if we thought this looked reasonable, and I think my last commit manages settings in the way Scott suggested. Unless there are other problems, or tests fail, I'll merge tomorrow morning.

.travis.yml
```diff
@@ -59,6 +65,15 @@ before_script:
...
   - cd ../
+
+  # Required for cypress tests; TODO: Move to puppet.
+  # TODO: Start up solr (and neo4j?)
+  - python manage.py loaddata guest.json
+  - python manage.py activate_user guest
+  - python manage.py add_users_to_public_group
+  - pushd ui && grunt make && popd
+  - python manage.py collectstatic --noinput -v 0
+  - python manage.py runserver --insecure --settings config.settings.prod_except_email &
```
hackdna: This configuration doesn't test any actual scenario. In dev mode Django devserver is used with `DEBUG=True`. In prod mode Apache is used with `DEBUG=False`.

Also, there are very few bugs that could appear in prod mode but not in dev and they would be highly dependent on the deployment environment anyway, so I would simply run Cypress tests with dev settings.

mccalluc:
Thanks: That makes sense, @hackdna. In the last commit:
- The python tests will continue to run with prod settings, as has been the case for a while, but runserver will use dev.
- In base.py we set console email as the default, and that is overridden if running prod.

PR #2240, opened 2017-10-17
Jkmarx/fix analyses padding

jkmarx:
- Makes padding around files/analysis/details tab CONTENT the same.
 *Note the provenance still needs to be fixed

### DIFF
--- a/refinery/templates/core/data_set.html
+++ b/refinery/templates/core/data_set.html
@@ -62,8 +62,6 @@
         </div>
       </div>
     </div>
-
-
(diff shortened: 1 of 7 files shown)


scottx611x:
In the same boat as Chuck. The changes look fine, but a before and after would be great

jkmarx:
## Current Test Site
screen shot 2017-10-18 at 10 16 08 am
screen shot 2017-10-18 at 10 16 12 am

jkmarx:
The padding is fixed to be similar to the details page. @scottx611x @mccalluc

PR #2243, opened 2017-10-18
In AWS deployment, check for None tags

drj11:
This is useful, because the default template for tags has
(the YAML for) `None` as the values.

### DIFF
--- a/deployment/utils.py
+++ b/deployment/utils.py
@@ -43,6 +43,11 @@ def load_tags():
     if 'owner' not in tags:
         tags['owner'] = os.popen("git config --get user.email").read().rstrip()
 
+    for k, v in tags.items():
+        if v is None:
+            raise RuntimeError(
+                "None is not legal, in tag {!r}".format(k))
+
     return [{'Key': k, 'Value': v} for k, v in tags.items()]
 
 


deployment/utils.py
```diff
@@ -43,6 +43,11 @@ def load_tags():
     if 'owner' not in tags:
         tags['owner'] = os.popen("git config --get user.email").read().rstrip()
 
+    for k, v in tags.items():
+        if v is None:
+            raise RuntimeError(
```
hackdna: Should we simply ignore the tags with no value instead?

deployment/utils.py
```diff
@@ -43,6 +43,11 @@ def load_tags():
...
+                "None is not legal, in tag {!r}".format(k))
```
hackdna: Minor formatting change: if an opening parenthesis is on its own line then so should be the closing one.

deployment/utils.py
drj11: yeah i quite like that actually. Should be just as straightforward to do.

deployment/utils.py
drj11: Sorry. I keep forgetting this, and `flake8` is happy to let it pass.

deployment/utils.py
hackdna: As we discussed last week, raising an error for tags with no value is actually more useful. I'm ready to merge when you fix the formatting nitpick below.

jkmarx:
@drj11 Don't forget about this pull request. Close?

PR #2247, opened 2017-10-18
Scottx611x/add force to generate tool definitions

scottx611x:
Useful not only for user sanity while generating their `ToolDefinitions` at first, but also great for idempotent integration tests in the future (thanks for the idea @mccalluc).

### DIFF
--- a/refinery/tool_manager/management/commands/generate_tool_definitions.py
+++ b/refinery/tool_manager/management/commands/generate_tool_definitions.py
@@ -32,6 +32,13 @@ class Command(BaseCommand):
             dest='workflows',
             help='Generate ToolDefinitions for properly annotated '
                  'Galaxy-based Workflows'
+        ),
+        make_option(
+            '--force',
+            action='store_true',
+            dest='force',
+            help='Overwrite ToolDefinitions with the same names instead of '
+                 'skipping their creation'
         )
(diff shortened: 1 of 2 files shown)


refinery/tool_manager/management/commands/generate_tool_definitions.py
```diff
@@ -32,109 +33,122 @@ class Command(BaseCommand):
...
+        ]
+        if tool_annotation["name"] in current_tool_definition_names:
+            if self.force:
+                self.stdout.write(
+                    self.style.NOTICE("Forcing deletion of of `{}`".format(
+                            tool_annotation["name"]))
+                )
+                ToolDefinition.objects.get(
+                    name=tool_annotation["name"]
+                ).delete()
```
mccalluc: Should there be a `return False` here? If the callers are depending on a false-y value, better to be explicit: I'd worry abut the `return True` being outdented at some point.

Or I could be confused.

scottx611x: Nope that make total sense, I'll add thanks!

mccalluc: or maybe retitle to `has_duplicates`?

PR #2255, opened 2017-10-20
Remove canEdit field and ng-if to show tools to all users.

jkmarx:
- Show Tool Button is available for all users


jkmarx:
@scottx611x Easy peasy

PR #2266, opened 2017-10-24
Fix field showing due to includes empty string returning true.

### LINKED ISSUE #2263: Attribute Filter Bug
attribute-filter-bug

jkmarx:
Resolves #2263 

fixed-attribute-filter

### DIFF
--- a/refinery/ui/source/js/file-browser/partials/assay-filters.html
+++ b/refinery/ui/source/js/file-browser/partials/assay-filters.html
@@ -54,7 +54,9 @@
         <div class="panel-body">
           <div ng-repeat="field in attributeObj.facetObj | orderBy: order track by $index">
             <div class="checkbox container"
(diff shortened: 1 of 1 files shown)


refinery/ui/source/js/file-browser/partials/assay-filters.html
```diff
@@ -54,7 +54,9 @@ <h4 class="panel-title">
         <div class="panel-body">
           <div ng-repeat="field in attributeObj.facetObj | orderBy: order track by $index">
             <div class="checkbox container"
-                ng-show="field.name.toLowerCase().includes(search.toLowerCase()) || showField(field.name, attributeObj.internal_name, attribute)">
+                ng-show="(search.toLowerCase().length && field.name
```
mccalluc: or just `search.length`? And maybe only line break on the `||`... or that might be too long? Thanks for fixing this / sorry for breaking it.

PR #2271, opened 2017-10-25
Scottx611x/tool manager test coverage

scottx611x:
Just adding some more test coverage to `tool_manager`


refinery/tool_manager/tests.py
```diff
@@ -1257,6 +1259,42 @@ def test_visualization_annotation_extra_dirs_missing(self):
...
+                self.mock_vis_annotations_reference,
+                return_value=tool_annotation
+        ):
+            with self.assertRaises(CommandError) as context:
+                call_command("generate_tool_definitions", visualizations=True)
+            self.assertIn("BAD FILETYPE", context.exception.message)
+            self.assertIn(
+                str([filetype.name for filetype in FileType.objects.all()]),
+                context.exception.message
+            )
```
mccalluc: Can we isolate just the parts that are different between these two? This seems like a good basic kind of test, and it should be easy to write more without copy-and-paste.
```python
with assert_tool_def_message('visualizations/bad_filetype.json') as message:
  self.assertIn("BAD FILETYPE", message)
  self.assertIn(
    str([filetype.name for filetype in FileType.objects.all()]),
    message
  )
```

refinery/tool_manager/tests.py
```diff
@@ -3260,3 +3298,91 @@ def test_valid_tool_launch_config_PAIR_LIST(self):
...
+        self.assertIn(bad_filetype, file_type_validation_error.message)
+        self.assertIn(error_message, file_type_validation_error.message)
+        self.assertIn(
+            str([f.name for f in FileType.objects.all()]),
+            file_type_validation_error.message
+        )
+
+    def test_get_visualization_annotations_list(self):
+        settings.VISUALIZATION_ANNOTATION_BASE_PATH = os.path.abspath(
+            os.getcwd()
```
mccalluc: Is this right? It seems to make assumptions based on where the test is run from. Do you mean something like `dirname(__file__)`?

refinery/tool_manager/tests.py
```diff
@@ -3260,3 +3298,91 @@ def test_valid_tool_launch_config_PAIR_LIST(self):
...
+        )
+        tool_definition_name = "dummy.json"
```
mccalluc: It looks like this test writes to the actual directory where tool defs are stored? ie, if the user actually had a `dummy.json`, it would be clobbered? Would it be possible to mock the internal setting for this dir, and have it point to a scratch space for the duration of the test?

Not sure if that makes sense.

refinery/tool_manager/tests.py
```diff
@@ -3260,3 +3298,91 @@ def test_valid_tool_launch_config_PAIR_LIST(self):
...
+    @mock.patch(
+        "bioblend.galaxy.workflows.WorkflowClient.show_workflow",
+        return_value={"graph": None}
+    )
+    def test_get_workflows(self, show_workflow_mock, exported_workflow_mock):
+        with mock.patch.object(
+            bioblend.galaxy.workflows.WorkflowClient,
+            "get_workflows",
+            return_value=[{"id": self.GALAXY_ID_MOCK}]
+        ) as bioblend_get_workflows_mock:
```
mccalluc: Sorry if this is basic, but is there a reason to mix the annotation and with syntaxes?

refinery/tool_manager/views.py
```diff
@@ -72,22 +71,18 @@ def get_queryset(self):
...
+        )
+        user_tools.extend(
+            get_objects_for_user(
+                self.request.user,
+                "tool_manager.read_visualizationtool"
             )
-        except GuardianError as e:
-            return HttpResponseBadRequest(e)
-
+        )
```
mccalluc: What changed? Does something else catch the GuardianError, or was that a mistake before?

scottx611x: This was a mistake before

refinery/tool_manager/tests.py
scottx611x: Not sure what you mean by this

refinery/tool_manager/tests.py
scottx611x: I change that behavior (`"writes to the actual directory where tool defs are stored"`) when I overwrite the `settings.VISUALIZATION_ANNOTATION_BASE_PATH`  at the beginning of this test

refinery/tool_manager/tests.py
scottx611x: 👍

PR #2276, opened 2017-10-25
Jkmarx/analysis tab url fix

### LINKED ISSUE #2027: Analysis Filtering url is not picked up by File Browser 2 facet filtering
Commit: `c33d8d5f9dd88168c140c5e99114c5e61e535d72`

@mccalluc Analysis facet filtering in FB2 is not triggered with the newly created urls. It seems like the logic to create the url on facet selection is there, but the selections aren't updated with the inverse manner.

jkmarx:
Resolves #2027

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -8,12 +8,10 @@ from __future__ import absolute_import
 import ast
 from collections import defaultdict
 from datetime import datetime
-import json
 import logging
 import os
 import smtplib
 import socket
-from urllib import quote
 from urlparse import urljoin
 
 from django import forms
@@ -1385,10 +1383,6 @@ class Analysis(OwnableResource):
             self.data_set.get_latest_study().id,
             self.data_set.get_latest_assay().id,
         )
-
-    def data_sets_query(self):
-        analysis_facet_name = self.facet_name()
-        return quote(json.dumps({analysis_facet_name: self.uuid}))
 
     def send_email(self):
         """Sends an email when the analysis is finished"""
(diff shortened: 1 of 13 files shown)


jkmarx:
@scottx611x analysis-redirect-fix

refinery/core/models.py
```diff
@@ -1387,6 +1387,15 @@ def data_sets_query(self):
         )
         return quote(json.dumps({analysis_facet_name: self.uuid}))
 
+    def facet_name(self):
```
scottx611x: Considering how the method is named, it might be worth being explicit and decorating this method as a `@property` ?

scottx611x: Also should reuse `facet_name` [here](https://github.com/refinery-platform/refinery-platform/blob/develop/refinery/core/models.py#L1383) now that we have it

refinery/core/models.py
```diff
@@ -1379,12 +1379,15 @@ def get_input_file_uuid_list(self):
...
-    def data_sets_query(self):
-        analysis_facet_name = '{}_{}_{}_s'.format(
+    def facet_name(self):
+        return '{}_{}_{}_s'.format(
             NodeIndex.ANALYSIS_UUID_PREFIX,
             self.data_set.get_latest_study().id,
             self.data_set.get_latest_assay().id,
         )
+
+    def data_sets_query(self):
```
mccalluc: Do we even want to keep this around at all? Is it ever the right way of doing what it's trying to do?

jkmarx:
@mccalluc Good point, the dash board should be updated accordingly.

PR #2278, opened 2017-10-25
Jkmarx/data set 1 cleanup

jkmarx:


refinery/templates/core/data_set.html
```diff
@@ -87,8 +90,13 @@
 
   <!-- Refinery libraries -->
   <script type="text/javascript">
+    var REFINERY_BASE_URL = document.location.host;
```
mccalluc: Is this just being moved from somewhere else? I'm not sure how it's used, but it might be important to get the protocol as well.

jkmarx: @mccalluc Yes, the globals are going to being revisited as we speak. Focused on just relocating the provenance within the tab.

refinery/static/source/js/refinery/contents.js
```diff
@@ -124,14 +86,8 @@
...
 
-          tableView = new SolrDocumentTable("solr-table-view", "solrdoctab1", query, client, configurator, documentTableCommands, dataSetMonitor);
-          tableView.setDocumentsPerPage(20);
-          analysisView = new SolrAnalysisView("solr-analysis-view", "solranalysis1", query, configurator, analysisViewCommands, dataSetMonitor);
-          facetView = new SolrFacetView("solr-facet-view", "solrfacets1", query, configurator, facetViewCommands);
-
-          pivotMatrixView = new SolrPivotMatrix("solr-pivot-matrix", "solrpivot1", query, {}, pivotMatrixCommands);
-
-          $(function() {
+//** PROVVIS QUERY **//
```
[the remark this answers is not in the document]
jkmarx: I wouldn't worry about the indents now, considering I'm trying to get rid of contents.js

PR #2283, opened 2017-10-27
Remove unused JS global. Fix #1765.

### LINKED ISSUE #1765: Remove REFINERY_USER_AUTHENTICATED if unused
refinery/templates/core/user_files.html :
```
    {% if user.is_authenticated %}
      var REFINERY_USER_AUTHENTICATED = true;
    {% else %}
      var REFINERY_USER_AUTHENTICATED = false;
```
The main problem would be if any kind of security relied on this...

mccalluc:
@jkmarx : I could imagine this causing merge conflicts for you... Thoughts?

### DIFF
--- a/refinery/static/source/js/refinery/contents.js
+++ b/refinery/static/source/js/refinery/contents.js
@@ -74,8 +74,9 @@
 
         // =====================================
 
+        // TODO: This comes from old file browser, and in what's left, download is not supported.
         function updateDownloadButton(button_id) {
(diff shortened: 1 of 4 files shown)


jkmarx:
@mccalluc I've already removed it from the javascript (on data-set-1 clean up) but not the templates yet. I can resolve the conflict later, so feel free to merge to develop.

mccalluc:
Thanks!

One build failed with what looks like a spurious cypress error:
```
CypressError: Timed out retrying: Expected to find content: 'We examined the pattern of gene expression' but never did.
```

I think that's just some AJAX taking too long to return -- I don't think we've hit any cypress errors yet on Travis, so on one side, it's nice to have confirmation that it is running. I'm rerunning now, and will merge if it passes.

PR #2288, opened 2017-10-30
Remove filetypes fq

### LINKED ISSUE #2106: Seeing unexpected values for "type" in file browser
(Started in #2089, but a new bug to reflect new, limited scope.)

Other facets seem fine, but the type facet returns stray zero values (with `mincount` removed). I had expected the list for every dataset to be the same, so I am probably not thinking about it right.

[Transcription factor ChIP-seq in expanded human hematopoietic stem and progenitor cells](https://test.stemcellcommons.org/data_sets/be9af94c-43fb-44a0-b357-12fdf0d8d008/#/files/):

> Raw Data File 10
> Assay Name 0
> Extract Name 0

[Molecular Analysis of antigen-specific B cell responses](https://test.stemcellcommons.org/data_sets/6a497724-acce-4e0d-8907-b98f0542061d/#/files/)

> Array Data File 14
>Derived Array Data Matrix File 4
> Derived Array Data File 2
> Data Transformation Name 0
> Extract Name 0
> Hybridization Assay Name 0
> Labeled Extract Name 0
> Scan Name 0

mccalluc:
Fix #2106?


mccalluc:
3 fails, just because solr query doesn't match:
```
FAIL: test_generate_solr_params_for_user (user_files_manager.tests.UserFilesUtilsTests)
FAIL: test_generate_solr_params_for_assay_with_params (data_set_manager.tests.UtilitiesTest)
FAIL: test_generate_solr_params_no_params (data_set_manager.tests.UtilitiesTest)
```
I'll fix those.

mccalluc:
One of two builds [fails](https://travis-ci.org/refinery-platform/refinery-platform/builds/294918045?utm_source=github_status&utm_medium=notification):
```
======================================================================
FAIL: test_get_dataset_expecting_analyses (core.tests.DataSetResourceTest)
  File "/home/travis/build/refinery-platform/refinery-platform/refinery/core/tests.py", line 1991, in test_get_dataset_expecting_analyses
    self.assertEqual(data['analyses'][0]['name'], a2.name)
AssertionError: u'a1' != 'a2'
```
Haven't seen this before. Just try restarting...

mccalluc:
I think that's where we left it... There are a couple ideas out there:
- Load up some data locally and see what it looks like with these changes.

but also:
- Maybe if we think it through, we'll realize that no amount of solr tweaking will give us the results we want, and it may require changes to the indexing.

mccalluc:
Thinking a little more: The simplest thing is probably just not to index these documents in the first place... but still store the files, in case we want to reindex and make them visible?

Issue #2095, opened 2017-09-06
Run docker containers on separate burstable instance (or ECS?)


mccalluc:
The docker containers need to run on a separate burstable EC2 instance: We don't want their resource use to get in the way of the main Refinery work. There is code in django_docker for making the appropriate calls to CloudFormation to build a stack, but I don't know if that can be reused.

mccalluc:
This does not have buy-in: Ilya believes ECS is a better solution to the problem. In my experience, starting containers through ECS is much slower than starting them directly on a docker engine, but I might have been doing it wrong.

Adding "(or ECS?)" to title and adding `Question` label.

ngehlenborg:
After discussion: expand root volume and create directory there. It will be that way no matter how we resolve the question about launching a separate EC2 or using ECS to run Docker containers.

mccalluc:
Continuing conversation from https://github.com/refinery-platform/refinery-platform/pull/2291#issuecomment-340595672: Ilya suggested I look into BlockDeviceMappings. I do think this is in the right direction, but it looks like there is a choice between ephemeral instance stores and ebs stores, and only for the latter can you set the volume size?  

- [Contrasting examples for the two](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-mapping.html#w2ab2c21c14d514c11b2)
- [Documentation explaining that size can only be specified for ebs volumes](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-mapping.html)
- Diagrams contrasting the two: [root device docs](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/RootDeviceStorage.html), [block device mapping docs](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/block-device-mapping-concepts.html)

mccalluc:
On the AWS instance:
```
ubuntu@ip-172-31-53-83:~$ sudo du -sh /var/lib/docker
2.2G	/var/lib/docker
ubuntu@ip-172-31-53-83:~$ sudo ls -dl /var/lib/docker
drwx--x--x 11 root root 4096 Oct 26 21:41 /var/lib/docker
```

hackdna:
"By default, the root device volume for an AMI backed by Amazon EBS is deleted when the instance terminates."

mccalluc:
[...]

PR #2291, opened 2017-10-30
Just make data volume larger

mccalluc:
Conclusion of #2095 was just to give it a larger volume, and this does that...
- Not sure this is the right setting,
- Or that doubling it is reasonable,
- Or that this is even what we really want to do.

### DIFF
--- a/deployment/aws-config/config.yaml.template
+++ b/deployment/aws-config/config.yaml.template
@@ -45,7 +45,7 @@ SERVER_EMAIL: "refinery-registration@example.org"
 EMAIL_SUBJECT_PREFIX: "[Refinery] "
 
 # size of the data volume in GB (ignored if a DATA_SNAPSHOT is specified)
-DATA_VOLUME_SIZE: 500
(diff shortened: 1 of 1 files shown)


hackdna:
The idea was to increase the root volume size. It looks like you'll need to use [BlockDeviceMappings property of AWS::EC2::Instance resource](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html#cfn-ec2-instance-blockdevicemappings) ([example](https://ig.nore.me/2015/03/increasing-the-size-of-a-root-ebs-volume/)).

mccalluc:
Talking with Scott, it feels like we might have started down the wrong path with this. The most limited space in all of this is the internal volume for the instance: Scott believes we may have had logs from HiGlass filling up this relatively limited space.

At the very least, this particular PR is not the right direction. I think the `BlockDeviceMappings` tweak could still be useful though, but primarily to handle the one-time costs of downloading a particular image.

PR #2300, opened 2017-10-31
Mccalluc/exponential explosion

mccalluc:
There are bugs and a lack of testing on the flat-file imports. This adds some tests... and I'd hope fixing the bugs might be easy, but it hasn't been obvious, yet.


refinery/data_set_manager/tests.py
```diff
@@ -1958,3 +1960,34 @@ def test_post_bad_isa_tab_file(self):
...
+            'test-data', 'single-file', filename
+        )
+        with open(path, 'r') as f:
+            process_metadata_table(
+                username='guest',
+                title='fake',
+                metadata_file=f,
+                source_columns=[0],
+                data_file_column=2,
+            )
```
scottx611x: Assertions here regarding created objects?

mccalluc:
The fix probably will not be in before we freeze... but it might be good to merge a PR even if it just surfaces a problem in a new test, even if we can't fully resolve it right now?

mccalluc:
@scottx611x : Here's something.
- Just put some duck tape on the false positive exponential nodes problem: Still don't understand what's actually going on.
- Single line CSV ingests now, and we assert that we get the right number of nodes out.
- Added a test of `update_annotated_nodes` where the problem really was... though the behavior there isn't what I'd expect.

refinery/data_set_manager/tests.py
mccalluc: see below

PR #2303, opened 2017-10-31
quieter grunt

mccalluc:
Feels like there should be a better way, but this cleans up the output without confusing the status

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,7 +88,12 @@ script:
   - flake8 --exclude=migrations,ui ..
 
   - echo 'travis_fold:start:grunt'
-  - pushd ui && grunt test && popd # TODO: 600 mostly useless lines of output
+  - pushd ui && grunt test > /tmp/grunt-test.txt && popd
+  # 'grunt test' is too verbose, and there's not a flag to quiet it.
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -88,7 +88,12 @@ script:
   - flake8 --exclude=migrations,ui ..
 
   - echo 'travis_fold:start:grunt'
-  - pushd ui && grunt test && popd # TODO: 600 mostly useless lines of output
+  - pushd ui && grunt test > /tmp/grunt-test.txt && popd
+  # 'grunt test' is too verbose, and there's not a flag to quiet it.
+  # Pipelining the output to grep would confuse grunt's exit status and grep's,
+  # so instead write to a temp file so we can get the abbreviated output,
+  # but ignore the grep status.
+  - grep -v SUCCESS /tmp/grunt-test.txt || true
```
hackdna: Grunt would not have a flag for reducing output of tasks that it runs. However, it is possible to reduce Karma output by selecting appropriate reporters in its config file (http://karma-runner.github.io/0.13/config/configuration-file.html). Have you tried removing the 'progress' reporter for example?

mccalluc: No, I didn't think of that. I've filed #2316 so this won't be lost.

PR #2314, opened 2017-11-02
Mount a new volume for docker parallel to /data

mccalluc:
Towards #2095. Began by looking at adding more space to the root volume, but it seems like it would be a good think for now to explicitly limit the space that can be consumed by Docker, and if it exceeds those bounds we'll notice immediately.

I have not tried to run this yet. Thoughts?


hackdna:
Docker needs ephemeral storage. Root volume provides it with minimal infrastructure code changes and this approach will be re-used on a separate Docker instance down the road. Managing additional volumes complicates things needlessly.

mccalluc:
replace with #2318

Issue #2294, opened 2017-10-30
Analysis results are not added to source data set


hackdna:
* Specific code commit: v1.6.0
* Version of the web browser and OS: Chrome on OS X
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): AWS

### Steps to reproduce
1. Set up a new stack (DATA_SNAPSHOT: snap-0b69229e9b0290af1 and RDS_SNAPSHOT: betastemcellcommons-snapshot-rdsinstance-1b7kwlbwn0ocz)
1. [Launch a new CloudMan cluster](https://github.com/stemcellcommons/stemcellcommons/wiki#galaxy-cloudman)
1. Delete old versions of workflows from Refinery
1. Import new versions of workflows from Galaxy
1. Update Solr index
1. Run a ChIP-seq or RNA-seq workflow

### Observed behavior
Analysis results are not added to the source data set

### Expected behavior
Analysis results are added to the source data set

### Notes
* FastQC workflow results are added to the data set
* ~~~May be related to #2293~~~
* ~~~from @scottx611x: "The Derived Nodes are created properly for these WF runs but they don't have `is_refinery_file=True` hence why they aren't being returned as part of the Dataset listing."~~~
* This seems to be due to a bug in Galaxy where one has to manually deselect asterisked Workflow outputs, save the Workflow, reselect said outputs, and save the workflow again when importing workflows from files.

### TO-DO:
- [x] Revert back to using step level annotations for outputs **See: https://github.com/refinery-platform/refinery-platform/pull/2432**
- [x] Make it impossible to import a workflow that doesn't declare `workflow_outputs` #2315 
- [x] Look into #2381 and if its related to this issue
- [x] See if I can get a minimal reproducer and test the bug [shown here](https://gist.github.com/scottx611x/3dd5e79d405547454bc0abbf233017e3) w/ newer versions of Galaxy locally.

scottx611x:
So this seems to be a side effect of utilizing the asterisking feature that Galaxy provides to select workflow outputs.

When a Workflow that specifies `workflow_outputs` is exported and reimported on a different Galaxy instance the UI will render the asterisked outputs but api responses won't yield them causing Refinery to not import any derived results. (This behavior doesn't seem to happen when one exports a Workflow and reimports it on the same Galaxy instance)

[...]

PR #2315, opened 2017-11-02
Scottx611x/ensure workflow outputs exist

scottx611x:
- Ensure that a user has properly set `workflow_outputs` for their Galaxy Workflow before allowing the creation of a ToolDefinition.
- Also Handle one-off error exhibited [here](https://gist.github.com/scottx611x/3dd5e79d405547454bc0abbf233017e3)

Towards #2294


scottx611x:
@hackdna @mccalluc Could you check this out when you get a chance?

hackdna:
Just to clarify: this will be rewritten when we go back to step annotations in 1.6.2, correct?

scottx611x:
@hackdna Yes

refinery/tool_manager/management/commands/generate_tool_definitions.py
```diff
@@ -178,6 +184,47 @@ def _generate_workflows(self):
                     )
                 )
 
+    def ensure_workflow_outputs_are_present(self, workflow):
```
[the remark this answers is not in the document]
mccalluc: Ditto: From the name, someone might think it adds the workflow outputs if missing.

naming: even more concise: `has_workflow_outputs()`. (but `has_outputs()` would be too short?)

refinery/tool_manager/management/commands/generate_tool_definitions.py
```diff
@@ -178,6 +182,47 @@ def _generate_workflows(self):
...
+        """
+        Assert that the given `workflow` has `workflow_outputs` properly
+        defined
+        :param workflow: dict containing a Galaxy Workflow's information
+        :raises: CommandError
+        """
+        workflow_steps = workflow["graph"]["steps"]
+        for step_index in workflow_steps:
+            step_has_workflow_outputs = workflow_steps[step_index].get(
```
mccalluc: This isn't *really* a boolean; probably no harm done, but still.

refinery/tool_manager/management/commands/generate_tool_definitions.py
```diff
@@ -178,6 +182,47 @@ def _generate_workflows(self):
...
+                WorkflowTool.WORKFLOW_OUTPUTS
+            )
+            if step_index == "0":
+                if not step_has_workflow_outputs:
+                    raise CommandError(
+                        "You've encountered known error when trying to "
```
mccalluc: "*a* known error"

PR #2320, opened 2017-11-02
We do not need to see each successful GET under Cypress

mccalluc:
Another piece towards making the log fit in 10K lines.


mccalluc:
(confirmed that the cypress portion of the logs doesn't have the http clutter any longer.)

PR #2327, opened 2017-11-03
Mccalluc/file browser buttons

### LINKED ISSUE #2322: "Show Tool Panel" should be left aligned
* Specific code commit: 4bd26aff83c2550bd9ee503ed21587c29eef0518 
* Version of the web browser and OS: Chrome and FF
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): Local dev

### Steps to reproduce

File browser:
screen shot 2017-11-03 at 9 53 14 am

### Expected behavior
* Shouldn't the "Show Tool Panel" button be left aligned with the div below? I feel like this has changed recently.

### LINKED ISSUE #2326: Data Set Table Configuration modal only triggers when clicking directly on wrench
* Specific code commit: 4bd26aff83c2550bd9ee503ed21587c29eef0518
* Version of the web browser and OS: `Chrome macOS`
* Environment where the error occurred: Vagrant

### Steps to reproduce

1. Got to a FileBrowser for any Dataset.
2. Click on the wrench button.

### Observed behavior
1. The Data Set Table Configuration modal only opens when the wrench icon is clicked directly, not when you click on the button alone.
nov-03-2017 11-46-24

### Expected behavior
* For the button clicks to trigger the modal opening as well.

mccalluc:
Fix #2322 and Fix #2326 

screen shot 2017-11-03 at 5 56 26 pm

A little awkward at narrow widths, but I don't think that's new. Perhaps save room by getting rid of the help if it really matters? Though I think it is fine as is.

### DIFF
--- a/refinery/ui/source/js/file-browser/views/files-tab.html
+++ b/refinery/ui/source/js/file-browser/views/files-tab.html
@@ -43,7 +43,7 @@
         </span>
 
       </div>
-      <div id="table-filter">
+      <div id="table-filter" class="row">
         <div id="filter-panel-body">
           <rp-file-browser-assay-filters>
           </rp-file-browser-assay-filters>
--- a/refinery/ui/source/js/file-browser/views/files-tab.html
+++ b/refinery/ui/source/js/file-browser/views/files-tab.html
@@ -25,8 +25,8 @@
         </button>
 
         <span class="float-right">
-          <span ng-if="FBCtrl.isOwner" class="refinery-base btn btn-default btn-sm">
-            <i class="fa fa-wrench" rp-assay-files-util-modal></i>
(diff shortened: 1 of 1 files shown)


refinery/ui/source/js/file-browser/views/files-tab.html
```diff
@@ -43,7 +43,7 @@
         </span>
 
       </div>
-      <div id="table-filter">
+      <div id="table-filter" class="row">
```
jkmarx: row?

mccalluc: The top div and the bottom div of that column were different widths because the top had `class="row"` and the bottom didn't ... simply removing that class meant the wrench button would sometimes wrap below and sit to the right of the `table-filter` div.

If there's something that would be better, let us know?

PR #2330, opened 2017-11-06
Skip indexing for some files

### LINKED ISSUE #2106: Seeing unexpected values for "type" in file browser

mccalluc:
Fix #2106? Does this seem like a reasonable approach? I still want to:
- Simplify the solr queries we construct, since these documents won't even be in the index now. (ie, add #2288 back in.)
- Move this list to configuration.
- Review other lists of hard-coded document types.


mccalluc:
On second thought, perhaps not a config, but class constant. I think this is all that is needed now?

refinery/data_set_manager/search_indexes.py
```diff
@@ -71,6 +72,8 @@ def _assay_data(self, object):
     # https://groups.google.com/forum/?fromgroups#!topic/django-haystack/g39QjTkN-Yg
     # http://stackoverflow.com/questions/7399871/django-haystack-sort-results-by-title
     def prepare(self, object):
+        if object.type not in Node.INDEXED_FILES:
+            raise SkipDocument()
```
scottx611x: Can `SkipDocument` accept a message? Could be useful for debugging.

mccalluc:
@ngehlenborg : Can you schedule time to give input on this, or delegate someone who can make a call? There are several fairly simple changes we could make that would fix the behavior, but we're not sure which is best.

mccalluc:
@scottx611x : Talked with Nils: Coming out of that, the idea was to show all documents, but for all the doc types we don't recognize, change the type, *but also* add a new field to the object so we could theoretically export the original.

With a little distance, that feels too complicated, and too fragile. How would you feel about merging this original PR, and putting the rest on the back burner?

scottx611x:
@mccalluc I agree on the fragility of: `for all the doc types we don't recognize, change the type, but also add a new field to the object so we could theoretically export the original`

Let me look over this PR again first and I'll give a better opinion

scottx611x:
@mccalluc I'd be okay with merging this as long as we move our future intentions into a new issue

Issue #2293, opened 2017-10-30
Downloading more files from Galaxy than specified by workflow


hackdna:
* Specific code commit: v1.6.0 tag
* Version of the web browser and OS: Chrome on OS X
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): AWS

### Steps to reproduce
Run "RNA-seq (PE) Quantification - Human" workflow on one pair of input files.

### Observed behavior
15 files are downloaded into Refinery. Celery worker2 log:
```
2017-10-30 16:10:02 DEBUG    file_store.tasks:69 import_file() - Importing FileStoreItem with UUID 'b4e83915-d773-46a9-a4e2-b2fea0268385'
2017-10-30 16:10:02 DEBUG    file_store.tasks:153 import_file() - Downloading from 'http://galaxy-prod.aws.stemcellcommons.org/datasets/cc8a3628a74b8797/display?to_ext=txt'
2017-10-30 16:10:03 DEBUG    file_store.tasks:204 import_file() - Finished downloading from 'http://galaxy-prod.aws.stemcellcommons.org/datasets/cc8a3628a74b8797/display?to_ext=txt'
2017-10-30 16:10:03 ERROR    file_store.tasks:250 update_solr_index() - Couldn't retrieve Node: Node matching query does not exist.
2017-10-30 16:10:03 ERROR    file_store.tasks:264 begin_auxiliary_node_generation() - Couldn't retrieve Node: Node matching query does not exist.
2017-10-30 16:10:03 DEBUG    file_store.tasks:69 import_file() - Importing FileStoreItem with UUID 'e3eac908-a1ad-4c7f-872f-bd1ff6e9d395'
2017-10-30 16:10:03 DEBUG    file_store.tasks:153 import_file() - Downloading from 'http://galaxy-prod.aws.stemcellcommons.org/datasets/9813f61fb27a58e8/display?to_ext=txt'
2017-10-30 16:10:04 DEBUG    file_store.tasks:204 import_file() - Finished downloading from 'http://galaxy-prod.aws.stemcellcommons.org/datasets/9813f61fb27a58e8/display?to_ext=txt'
2017-10-30 16:10:04 ERROR    file_store.tasks:250 update_solr_index() - Couldn't retrieve Node: Node matching query does not exist.
2017-10-30 16:10:04 ERROR    file_store.tasks:264 begin_auxiliary_node_generation() - Couldn't retrieve Node: Node matching query does not exist.
[...]

PR #2335, opened 2017-11-07
Scottx611x/galaxy connector pindown tests

scottx611x:
Towards #2293 

We utilize `galaxy_connector.models.Instance.get_history_file_list()` while determining which files to download from our Galaxy History so I thought it would be a good idea to pin down its current functionality w/ some tests.


refinery/galaxy_connector/models.py
```diff
@@ -24,68 +24,35 @@ def get_history_file_list(self, history_id):
...
-        for history_content_entry in connection.histories.show_history(
-                history_id, contents=True):
-
-            if "type" not in history_content_entry:
+        history_content_keys = ["state", "file_size", "visible",
+                                "file_name", "genome_build", "misc_info",
+                                "misc_blurb"]
+        history_contents = connection.histories.show_history(
+            history_id,
+            contents=True
```
hackdna: These two don't need to be on separate lines (see above statement for example).

refinery/galaxy_connector/models.py
```diff
@@ -24,68 +24,35 @@ def get_history_file_list(self, history_id):
...
+        for history_content_entry in history_contents:
+            if ("type" not in history_content_entry or
+                    history_content_entry["type"] != "file"):
                 continue
 
-            if history_content_entry["type"] != "file":
-                continue
+            history_content = connection.histories.show_dataset(
+                history_id,
+                history_content_entry["id"]
```
hackdna: These two don't need to be on separate lines

refinery/galaxy_connector/tests.py
```diff
@@ -0,0 +1,104 @@
...
+class GalaxyInstanceTests(TestCase):
+    def setUp(self):
+        self.GALAXY_HISTORY_ID = str(uuid.uuid4())
+        self.GALAXY_DATASET_ID = str(uuid.uuid4())
+        self.GALAXY_DATASET_FILESIZE = 1024
+        self.MISCELLANEOUS_STRING = "Coffee is tasty"
+
+        self.galaxy_instance = Instance.objects.create(
+            base_url="www.example.com",
+            api_key=str(uuid.uuid4()),
```
hackdna: These two don't need to be on separate lines

refinery/galaxy_connector/tests.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+        self.show_history_mock = mock.patch.object(
+            galaxy.histories.HistoryClient,
+            "show_history"
```
hackdna: These two don't need to be on separate lines

refinery/galaxy_connector/tests.py
```diff
@@ -0,0 +1,104 @@
...
+        ).start()
+        self.show_dataset_mock = mock.patch.object(
+            galaxy.histories.HistoryClient,
+            "show_dataset"
```
hackdna: These two don't need to be on separate lines

PR #2336, opened 2017-11-07
Update test to be more robust

### LINKED ISSUE #2304: Potentially non-idempotent test?
Commit: c67b47e75cc089952cb8cfed9fc41a743cef8856
@mccalluc talked about a flippant test failure here: https://github.com/refinery-platform/refinery-platform/pull/2288#issuecomment-340491336

This looks more like a test relying on an api response that isn't always in the order it is expecting. Will need some looking into, but that is my gut feeling.

I feel like someone else saw this error before as well, but I can't find where.

Wonky test here: [test_get_dataset_expecting_analyses](https://github.com/refinery-platform/refinery-platform/blob/master/refinery/core/tests.py#L1928)

scottx611x:
Fixes #2304

### DIFF
--- a/refinery/core/tests.py
+++ b/refinery/core/tests.py
@@ -1317,19 +1317,24 @@ class DataSetResourceTest(LoginResourceTestCase):
         self.assertValidJSONResponse(response)
         data = self.deserialize(response)
         self.assertEqual(data['uuid'], self.dataset.uuid)
-        self.assertIsNotNone(data['analyses'])
         self.assertEqual(len(data['analyses']), 2)
 
-        self.assertIsNotNone(data['analyses'][0]['is_owner'])
-        self.assertTrue(data['analyses'][0]['is_owner'])
-        self.assertIsNotNone(data['analyses'][0]['owner'])
-        self.assertEqual(
-            data['analyses'][0]['owner'],
-            UserProfile.objects.get(user=self.user).uuid
+        sorted_analyses = sorted(
+            data['analyses'],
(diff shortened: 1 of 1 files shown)


refinery/core/tests.py
```diff
@@ -1317,19 +1317,24 @@ def test_get_dataset_expecting_analyses(self):
...
 
-        self.assertIsNotNone(data['analyses'][0]['is_owner'])
-        self.assertTrue(data['analyses'][0]['is_owner'])
-        self.assertIsNotNone(data['analyses'][0]['owner'])
-        self.assertEqual(
-            data['analyses'][0]['owner'],
-            UserProfile.objects.get(user=self.user).uuid
+        sorted_analyses = sorted(
+            data['analyses'],
+            key=lambda x: x["name"]
```
hackdna: These two don't need to be on separate lines

refinery/core/tests.py
```diff
@@ -1317,19 +1317,24 @@ def test_get_dataset_expecting_analyses(self):
...
         )
-        self.assertEqual(data['analyses'][0]['status'], a2.status)
-        self.assertEqual(data['analyses'][0]['name'], a2.name)
-        self.assertEqual(data['analyses'][0]['uuid'], a2.uuid)
+        for analysis in sorted_analyses:
+            self.assertTrue(analysis['is_owner'])
+            self.assertEqual(
```
hackdna: No need for extra line breaks (see above comment)

Issue #2307, opened 2017-11-01
Replace stack-s3.py with Terraform configuration


hackdna:
Motivated by #1837.

hackdna:
Closed in #2337

Issue #1837, opened 2017-06-26
Find replacement for cfn-pyplates


hackdna:
The last release was on 2016-03-13
Doesn't support some of the latest CloudFormation features like cross-stack references

Potential replacements:
* YAML
* https://github.com/cloudtools/troposphere
* https://www.terraform.io/

Specs:
* Management of existing resources (e.g., S3 buckets)
* Support and features
* Works across cloud providers

See also:
* #1736 
* [Impossible to use a rule with the AbortIncompleteMultipartUpload action](https://forums.aws.amazon.com/message.jspa?messageID=781442)

PR #2337, opened 2017-11-07
Hackdna/terraform prototype

hackdna:
Toward #2307


hackdna:
* Yes, this PR achieves the goal of #2307.
* Yes, stack.py will be replaced eventually.
* I've considered Troposphere along with a couple of other options. It is essentially a replacement for cfn-pyplates, so it still ties us to CloudFormation which is insufficient for our needs (#1837).
* Yes, I'll add documentation to the wiki.
* See above.

PR #2366, opened 2017-11-13
Separate S3 upload bucket

hackdna:
Solves #2370

### DIFF
--- a/deployment/terraform/modules/cognito/main.tf
+++ b/deployment/terraform/modules/cognito/main.tf
@@ -54,7 +54,7 @@ resource "aws_iam_role_policy" "upload_access_policy" {
         "s3:AbortMultipartUpload"
       ],
       "Effect": "Allow",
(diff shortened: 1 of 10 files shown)


hackdna:
Yes, I test `CheckDataFilesView` every time I make changes to file upload code.

scottx611x:
@hackdna I'm talking about programmatic tests

mccalluc:
Talking to Scott: `upload` is just temporary space where files go immediately after upload but before they are in the filestore, and migration isn't something we need to worry about, because we're not going to worry about making sure a half completed ingest can complete.

hackdna:
The upload bucket is special because it must allow writes by the front end code. The media bucket (main data file storage) has very different permissions. Those permissions are applied at the bucket level thus two buckets are needed. More details: https://www.mattdaniels.com.au/javascript-integration-with-aws-s3/

PR #2369, opened 2017-11-13
Command-line multi-file download

### LINKED ISSUE #2029: bash hint for multi file download
The 4DN portal suggests a one-liner that parses the metadata download file and does the download: Do the same thing in refinery.

mccalluc:
Fix  #2029
screen shot 2017-11-13 at 3 08 52 pm

- Buttons seemed more idiomatic than links, but I could be swayed.
- Probably want more space around buttons, but didn't want to invest more time there unless we're pretty sure it's what we want.
- The commandline is really long, because the query arguments reflect the internal solr names. Alternatively, we could have folks download the CSV, and then give them a command line assuming that they have the csv: Prettier, but more space for things to go wrong?
- Text ok?

### DIFF
--- a/refinery/ui/source/js/user-file-browser/ctrls/user-file-browser-files-ctrl.js
+++ b/refinery/ui/source/js/user-file-browser/ctrls/user-file-browser-files-ctrl.js
@@ -69,11 +69,19 @@
     };
 
     gridOptionsService.appScopeProvider = vm;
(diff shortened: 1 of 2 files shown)


mccalluc:
screen shot 2017-11-14 at 4 06 16 pm

Added padding and change to textarea. We can select the text, but copying it is impossible (I think) with browser native technology.

PR #2377, opened 2017-11-14
Scottx611x/galaxy assistant class

scottx611x:
Move methods utilizing `handle_bioblend_exceptions` to their own class (`GalaxyAssitant`) that `WorkflowTool` can now extend.


scottx611x:
Closing for now, this works but there's probably a better design pattern I could choose

Issue #2348, opened 2017-11-09
Remove flakey Satori Cypress test?


mccalluc:
A few times now we've hit
> Satori Visualization Sorts nodes in list graph by name:
>    CypressError: Timed out retrying: Expected to find element: '.node.lock-directly', but never found it.

Options:
- Do nothing
- Remove test
- Try to make Satori more reliable?
- Up the timeout in Cypress, or something like that?

hackdna:
FWIW, I've seen this fail in about half of the Travis runs today. For example: https://travis-ci.org/refinery-platform/refinery-platform/builds/311508803
I would suggest removing this test unless it can be run reliably.

hackdna:
Still fails randomly: https://travis-ci.org/refinery-platform/refinery-platform/builds/347922649
I vote to expedite removal.

scottx611x:
Ran into this again: https://travis-ci.org/refinery-platform/refinery-platform/builds/377882967?utm_source=slack&utm_medium=notification

hackdna:
Still a problem: https://travis-ci.org/refinery-platform/refinery-platform/builds/396934453
I vote to remove ASAP.

PR #2384, opened 2017-11-17
Scottx611x/remove workflow files dl model

### LINKED ISSUE #2173: Remove the `WorkflowFilesDL` model
First off, `WorkflowFilesDL` is not a great name. If we do need it, lets update that.

This model keeps track of Galaxy Dataset's filenames and which Workflow step they correspond to which is then used later on to aid in downloading said datasets, but seems to only need to stick around for legacy `Analysis` runs.

I'm not convinced we need this model at all for new `WorkflowTool`-based `Analysis` runs.

- [x] Look into this here: https://github.com/refinery-platform/refinery-platform/blob/develop/refinery/analysis_manager/tasks.py#L640-L724

---

- [x] Remove code creating `WorkflowFilesDL` instances
- [x] Remove afformentioned code
- [x] Remove ManyToMany in `Analysis`
- [x] Remove `WorkflowFilesDL` Model & Admin interface
- [x] Remove tests

scottx611x:
Closes #2173 

- [x]  Test this on an instance with beta.scc snapshots and check how existing Analyses are affected

### DIFF
--- a/refinery/analysis_manager/tasks.py
+++ b/refinery/analysis_manager/tasks.py
@@ -464,19 +464,8 @@ def _get_galaxy_download_task_ids(analysis):
 
     # retrieving list of files to download for workflow
     tool = _get_workflow_tool(analysis.uuid)
-    tool.create_workflow_file_downloads()
     tool.create_analysis_output_node_connections()
 
-    dl_files = analysis.workflow_dl_files
-    # creating dictionary based on files to download predetermined by workflow
-    # w/ keep operators
-    dl_dict = {}
-    for dl in dl_files.all():
-        temp_dict = {
-            'filename': dl.filename,
-            'pair_id': dl.pair_id
-        }
-        dl_dict[str(dl.step_id)] = temp_dict
     galaxy_instance = analysis.workflow.workflow_engine.instance
 
     try:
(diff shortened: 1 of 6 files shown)


scottx611x:
Failing build  above from cypress test: `Satori Visualization Sorts nodes in list graph by name`
See: #2348

scottx611x:
@hackdna not sure. I'll test this on an instance with `beta.scc` snapshots and check

scottx611x:
@hackdna I've gotten around to testing this out w/ beta data and existing Analyses and their data aren't affected.

The idea behind `WorkflowFilesDL` was to be a temporary placeholder for derived data file info until an actual `FileStoreitem` got created. After the FSI creation it was a non-essential entity

refinery/analysis_manager/tasks.py
```diff
@@ -464,19 +464,8 @@ def _get_galaxy_download_task_ids(analysis):
 
     # retrieving list of files to download for workflow
     tool = _get_workflow_tool(analysis.uuid)
-    tool.create_workflow_file_downloads()
     tool.create_analysis_output_node_connections()
 
-    dl_files = analysis.workflow_dl_files
-    # creating dictionary based on files to download predetermined by workflow
-    # w/ keep operators
-    dl_dict = {}
```
mccalluc: huh: This dict didn't seem to be used, even before your changes? Just vestigial, probably?

scottx611x: @mccalluc It was used way back when we were launching analyses the old way, but looking at this code in particular was the motivator for the removal of the whole model

PR #2392, opened 2017-11-27
Allow deletion of the empty user data S3 buckets

hackdna:
`aws_s3_bucket` resource cannot be deleted if bucket is not empty, so extra protection provided by the `lifecycle` meta-parameter is unnecessary. Also, this makes removal of temporary infrastructures easier.


drj11:
so what happens if you try to delete a terraform stack and it does have a bucket that contains data in it? Do you know?

hackdna:
That bucket will be left intact and TF will show a `BucketNotEmpty` error message. It will still delete other resources.

PR #2385, opened 2017-11-17
Scottx611x/galaxy cleanup on delete or cancel

scottx611x:
Fixes #2375

PR #2410, opened 2017-12-06
Specify `base_url` in GalaxyInstance Factory

### LINKED ISSUE #2408: Assertion error in test_analysis_deletion
* Specific code commit: e2bb3a6a
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): Vagrant VM

### Steps to reproduce
Run:
```shell
manage.py test selenium_testing.tests.UiDeletionTestCase.test_analysis_deletion
```

### Observed behavior
Test run output:
```
2017-12-05 17:00:52 INFO     core.models:1331 galaxy_cleanup() - Purging galaxy of libraries, histories, and workflows related to the execution of Analysis with UUID: 96a8c5f5-989d-4f0c-98f9-799f28ea388c
2017-12-05 17:00:52 ERROR    django.request:231 handle_uncaught_exception() - Internal Server Error: /api/v2/analyses/96a8c5f5-989d-4f0c-98f9-799f28ea388c/
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 57, in wrapped_view
    return view_func(*args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/views/generic/base.py", line 69, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/rest_framework/views.py", line 466, in dispatch
    response = self.handle_exception(exc)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/rest_framework/views.py", line 463, in dispatch
    response = handler(request, *args, **kwargs)
  File "/vagrant/refinery/core/views.py", line 913, in delete
    analysis_deleted = Analysis.objects.get(uuid=uuid).delete()
  File "/vagrant/refinery/core/models.py", line 1196, in delete
    self.cancel()
  File "/vagrant/refinery/core/models.py", line 1359, in cancel
    self.galaxy_cleanup()
  File "/vagrant/refinery/core/models.py", line 1333, in galaxy_cleanup
    self.galaxy_instance().delete_library(self.library_id, self.name)
  File "/vagrant/refinery/galaxy_connector/models.py", line 72, in delete_library
    self.galaxy_connection().libraries.delete_library(library_id)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/bioblend/galaxy/libraries/__init__.py", line 54, in delete_library
    return self._delete(id=library_id)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/bioblend/galaxy/client.py", line 187, in _delete
    r = self.gi.make_delete_request(url, payload=payload, params=params)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/bioblend/galaxyclient.py", line 160, in make_delete_request
    r = requests.delete(url, verify=self.verify, data=payload, params=params)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/requests/api.py", line 147, in delete
    return request('delete', url, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/requests/api.py", line 50, in request
    response = session.request(method=method, url=url, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/requests/sessions.py", line 451, in request
    prep = self.prepare_request(req)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/requests/sessions.py", line 382, in prepare_request
    hooks=merge_hooks(request.hooks, self.hooks),
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/requests/models.py", line 304, in prepare
    self.prepare_url(url, params)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/requests/models.py", line 365, in prepare_url
    raise InvalidURL("Invalid URL %r: No host supplied" % url)
InvalidURL: Invalid URL u'http:///api/libraries': No host supplied
...
FAIL: test_analysis_deletion (selenium_testing.tests.UiDeletionTestCase)
Delete some analyses and make sure the UI updates properly
  File "/vagrant/refinery/selenium_testing/tests.py", line 64, in test_analysis_deletion
    delete_from_ui(self.browser, "analysis", total_analyses)
  File "/vagrant/refinery/selenium_testing/utils.py", line 201, in delete_from_ui
    "{} {}".format(total_objects, object_name))
  File "/vagrant/refinery/selenium_testing/utils.py", line 96, in assert_text_within_id
    selenium.find_element_by_id(search_id).text
AssertionError: "1 analysis" not in total-analyses:
2 analyses
```

### Expected behavior
No test errors or failures.

scottx611x:
Fixes #2408

### DIFF
--- a/refinery/factory_boy/django_model_factories.py
+++ b/refinery/factory_boy/django_model_factories.py
@@ -86,7 +86,7 @@ class GalaxyInstanceFactory(factory.django.DjangoModelFactory):
     """Minimal representation of a GalaxyInstance"""
     class Meta:
         model = "galaxy_connector.Instance"
-    base_url = "http://www.example.com/galaxy"
+    # base_url = "http://www.example.com/galaxy"
 
 
 class WorkflowEngineFactory(factory.django.DjangoModelFactory):
--- a/refinery/factory_boy/django_model_factories.py
+++ b/refinery/factory_boy/django_model_factories.py
@@ -86,6 +86,7 @@ class GalaxyInstanceFactory(factory.django.DjangoModelFactory):
     """Minimal representation of a GalaxyInstance"""
     class Meta:
         model = "galaxy_connector.Instance"
+    base_url = "www.example.com/galaxy/"
 
 
 class WorkflowEngineFactory(factory.django.DjangoModelFactory):
--- a/refinery/factory_boy/django_model_factories.py
+++ b/refinery/factory_boy/django_model_factories.py
@@ -86,7 +86,7 @@ class GalaxyInstanceFactory(factory.django.DjangoModelFactory):
     """Minimal representation of a GalaxyInstance"""
     class Meta:
         model = "galaxy_connector.Instance"
-    # base_url = "http://www.example.com/galaxy"
+    base_url = "http://www.example.com/galaxy"
 
 
 class WorkflowEngineFactory(factory.django.DjangoModelFactory):
--- a/refinery/factory_boy/django_model_factories.py
+++ b/refinery/factory_boy/django_model_factories.py
@@ -86,7 +86,7 @@ class GalaxyInstanceFactory(factory.django.DjangoModelFactory):
     """Minimal representation of a GalaxyInstance"""
     class Meta:
         model = "galaxy_connector.Instance"
-    base_url = "www.example.com/galaxy/"
+    base_url = "http://www.example.com/galaxy"
 
 
 class WorkflowEngineFactory(factory.django.DjangoModelFactory):
--- a/refinery/analysis_manager/tests.py
+++ b/refinery/analysis_manager/tests.py
@@ -20,9 +20,9 @@ from analysis_manager.utils import (fetch_objects_required_for_analysis,
 from analysis_manager.views import analysis_status, run
 from core.models import Analysis, DataSet, Project, Workflow, WorkflowEngine
 from data_set_manager.models import Assay
+from factory_boy.django_model_factories import GalaxyInstanceFactory
 from factory_boy.utils import (create_dataset_with_necessary_models,
                                make_analyses_with_single_dataset)
-from galaxy_connector.models import Instance
 from tool_manager.models import ToolDefinition
 from tool_manager.tests import ToolManagerTestBase
 
@@ -106,7 +106,7 @@ class AnalysisUtilsTests(TestCase):
     different types of valid analysis configs
     """
     def setUp(self):
-        self.galaxy_instance = Instance.objects.create()
+        self.galaxy_instance = GalaxyInstanceFactory()
         self.workflow_engine = WorkflowEngine.objects.create(
             instance=self.galaxy_instance
         )
--- a/refinery/core/tests.py
+++ b/refinery/core/tests.py
@@ -19,9 +19,9 @@ from tastypie.test import ResourceTestCase
 
 from analysis_manager.models import AnalysisStatus
 from data_set_manager.models import Assay, Contact, Investigation, Node, Study
+from factory_boy.django_model_factories import GalaxyInstanceFactory
 from factory_boy.utils import create_dataset_with_necessary_models
 from file_store.models import FileStoreItem, FileType
-from galaxy_connector.models import Instance
 
 from .api import AnalysisResource
 from .management.commands.create_user import init_user
@@ -138,7 +138,7 @@ class AnalysisResourceTest(LoginResourceTestCase):
         ).catch_all_project
         self.dataset = create_dataset_with_necessary_models()
         self.dataset2 = create_dataset_with_necessary_models()
-        self.galaxy_instance = Instance.objects.create()
+        self.galaxy_instance = GalaxyInstanceFactory()
         self.workflow_engine = WorkflowEngine.objects.create(
             instance=self.galaxy_instance
         )
@@ -573,7 +573,7 @@ class WorkflowDeletionTest(TestCase):
             self.username, '', self.password
         )
         self.project = Project.objects.create()
-        self.galaxy_instance = Instance.objects.create()
+        self.galaxy_instance = GalaxyInstanceFactory()
         self.workflow_engine = WorkflowEngine.objects.create(
             instance=self.galaxy_instance
         )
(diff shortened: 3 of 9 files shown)


refinery/factory_boy/django_model_factories.py
```diff
@@ -86,6 +86,7 @@ class GalaxyInstanceFactory(factory.django.DjangoModelFactory):
     """Minimal representation of a GalaxyInstance"""
     class Meta:
         model = "galaxy_connector.Instance"
+    base_url = "www.example.com/galaxy/"
```
hackdna: Curious as to why was this not necessary before e2bb3a6 and also, why is this not necessary when running the tests on Travis?

scottx611x: It wasn't necessary before #2385 because there was no logic in place to run `galaxy_cleanup` for  non-successful deleted or cancelled Analyses.

Our GalaxyInstance needs a `base_url` now that we are able to run this cleanup on Analysis deletion.

As for why things weren't failing on Travis I'm not sure.

hackdna: It would be great to find out why test outcome is different in Travis and Vagrant because this could be a symptom of a larger problem. Also, according to the [docs](https://github.com/refinery-platform/refinery-platform/wiki/setting-up-newer-galaxy#configure-refinery-for-your-galaxy-instance), `base_url` should have no trailing slash.

scottx611x:
@hackdna Regarding the discrepancy between local and travis test runs:

I haven't been able to uncover much, although I've observed a passing full test suite locally with commit: e2bb3a6aad9d4f55ea9ec8988e8ad808ad7d8b7e, `prod` & `dev` settings,  and by running `manage.py test` which is in line with the Travis behavior. 

I was also able to reproduce the test failure with this same setup by running said test on its own: `manage.py test selenium_testing.tests.UiDeletionTestCase.test_analysis_deletion` as @hackdna reported.

hackdna:
Thanks, it sounds like this issue is due not to a difference between Travis and Vagrant but there is an inter-dependency between tests: results of running some test(s) can affect the outcome of `test_analysis_deletion`.
If tests are not isolated/independent from each other it can make debugging very difficult, so this interdependency should be found and eliminated.

hackdna:
Also, running `manage.py test selenium_testing` fails which means that the interdependency is on test(s) outside the `selenium_testing` module.

scottx611x:
I've spent too much time looking into this today to no avail. I think it would be less effort to push forward and write this as a Cypress test like we envisioned.

hackdna:
Rewriting `test_analysis_deletion` is not going to help because the problem is caused by some test(s) that appear to leave the database in a dirty state. This can cause other/future tests to produce incorrect results and thus make the whole test suite useless.

scottx611x:
So I've gotten a db_dump before each test run of the `selenium_testing` app:

Run w/ full test suite (passing): https://gist.github.com/scottx611x/c69411bf25e9dbfb4f89ab4ef6b3e77d#file-selenium_tests_passing-json
Run `selenium_testing` only (failing): https://gist.github.com/scottx611x/c69411bf25e9dbfb4f89ab4ef6b3e77d#file-selenium_tests_failing-json

Diff-ing these files doesn't yield anything interesting other than some different pks, uuids, and timestamps which is expected.

@hackdna How would you feel about merging these changes to allow for the selenium-based tests to run on their own and do some further investigation later?

scottx611x:
@hackdna Interestingly enough I believe I've found the culprit while working in another branch:

Adding the missing super call in the `tearDown()` [here](https://github.com/refinery-platform/refinery-platform/pull/2414/files#diff-36bd008eb6c525a4a81c39dcda428fcfR273) causes the `selenium-testing` tests to properly fail while running the entire test suite.

hackdna:
This appears to work by coincidence because there is no `tearDown()` in `ToolManagerMocks` (the parent of `ToolManagerTestBase`) and the only `tearDown()` is actually in `unittest.TestCase` and it is a no-op.

My guess is that this problem could be caused by a combination of not destroying model instances in `tearDown()` that were created in `setUp()` and using [TransactionTestCase](https://docs.djangoproject.com/en/1.7/topics/testing/tools/#transactiontestcase) since it resets the database by truncating the tables. On the other hand, a `TestCase` guarantees that the rollback at the end of the test restores the database to its initial state.

hackdna:
Great, though it appears more than that is needed: https://docs.python.org/3/library/unittest.mock.html#start-and-stop

scottx611x:
@hackdna I'm not immediately seeing what else is needed from those docs. Could you elaborate?

hackdna:
Several `start()` calls are missing in `ToolManagerMocks.setUp()`. These patches won't be stopped by `stopall()` in tearDown().
Also, there are `start()` calls in other places in the codebase without a corresponding `stop()` call.

scottx611x:
`stopall()` will take care of those other instances since all tool_manager test cases are derived from `ToolManagerMocks` and will gain the benefit of its `tearDown()`. Mocks don't have to be started in the `setUp()`.

I can look into the other `start`s outside of tool_manager that are never stopped.

hackdna:
From https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.stopall:
patch.stopall()
Stop all active patches. Only stops patches started with start.

scottx611x:
@hackdna I think the documentation mentions that because there are a few other ways to mock things that `stopall()` won't terminate : decorators, context managers etc.

If a mock is never started then it won't change the behavior of the mocked reference.

For example:

Assuming a file`mock_test.py` exists:
```
# mock_test.py
def blah():
    return "blah"
```
Drop to a shell:
```
>>> import mock
>>> import mock_test
>>> blah_mock = mock.patch("mock_test.blah", return_value="MOCKED")
>>> blah_mock
<mock.mock._patch object at 0x7f712edc3590>
>>> assert mock_test.blah() == "blah"
>>> blah_mock.start()
<MagicMock name='blah' id='140124050715792'>
>>> assert mock_test.blah() == "MOCKED"
```

scottx611x:
Build failure is known. See: #2348

Issue #2427, opened 2017-12-13
How should we handle the Exponential Explosion error state?


scottx611x:
[Hotfix v1.6.1.1](https://github.com/refinery-platform/refinery-platform/releases/tag/v1.6.1.1) was introduced to allow for some ISATabs to be imported without failure.

@fritz introduced the logic to handle this "Exponential Explosion" error state [here](https://github.com/refinery-platform/refinery-platform/pull/1763). Along with a [good explanation](https://github.com/refinery-platform/refinery-platform/pull/1763#issuecomment-313731061) of what state this "Expontential Explosion" should be catching.

While exploring this the other day, I found the [following information](https://gist.github.com/scottx611x/ad4d73e628a0c40fd10a431576b4916c) while testing this out with  [Shannan's ISATab](https://github.com/refinery-platform/refinery-platform/files/1556368/Kim-BRG1-test.2.zip).

TODO:
- [ ] When a solution is found we should reintroduce tests similar to what we had [here](https://github.com/refinery-platform/refinery-platform/blob/v1.6.1/refinery/data_set_manager/tests.py#L1925-L1943)

PR #2428, opened 2017-12-13
Hotfix 1.6.1.1

scottx611x:
This hotfix allows for @sjhosui's ISATabs to be imported without failure.

I've opened https://github.com/refinery-platform/refinery-platform/issues/2427 to be able to properly address this moving forward

- [x] Test on AWS with [Shannan's ISATab](https://github.com/refinery-platform/refinery-platform/files/1556368/Kim-BRG1-test.2.zip)


refinery/data_set_manager/utils.py
```diff
@@ -464,11 +464,10 @@ def update_annotated_nodes(
         # TODO: Not happy about this hack at all.
         error_message = (
             "Exponential explosion! Creation of {} annotated nodes for {} "
-            "nodes of type {} is stopped!"
+            "nodes of type {}"
         ).format(total_attrs, num_nodes_of_type, node_type)
 
         logger.error(error_message)
```
hackdna: Should this be `logger.warning(warning_message)` since import is no longer prevented?

scottx611x: Would make sense to leave it as `logger.error` since this state, as far as we know, is an error? 

Either way is fine with me.

hackdna: Well, it turned out that it is definitely not an error for some data sets.

scottx611x: Even though these Datasets can now import without this check in place there is still the deeper issue of many `Attributes` being created with duplicated info from parsing their ISATabs and this logic still catches that. Maybe leaving it as an error message would keep us on our toes when seeing this in the future.

PR #2437, opened 2017-12-17
Jkmarx/add visualization tab

### LINKED ISSUE #2260: Display Users launched Visualizations in the UI somewhere
Similar to how we display a User's launched `Analyses`, there should be somewhere in the UI where a `User's` launched `VisualizationTools` are shown as well.

Things to consider:
- Should the middle column on the dashboard display a merged list of `Analyses` and `Visualizations`?
- Should that column exist at all? I think we've talked before about having this information about `Analysis` and `Visualization` launches shown in the `Dataset Preview`
- A `User` should be able to click and navigate to a previously launched `VisualizationTool` if it is still running #2433
- ~~~A `User` may want to reanimate a "dead" `VisualizationTool`~~~ #2433 
- A deletion button & UI tests will be needed for items in the `VisualizationTool` listing as well
- A change like this could warrant the removal of the 2nd and 3rd columns of the dashboard which could make room for the editable markdown-based landing page that was demonstrated [here](https://github.com/scottx611x/django_dynamic_markdown_pages) ([Requires Django 1.8](https://github.com/refinery-platform/refinery-platform/issues/449))

Notes I've written earlier:
```
Tool.has_container ->Default False
Tool.is_new -> default True

Utilize these two fields to be able to reanimate Visualization Tools
```
```
VisualizationTool urls to utilize their own uuid so that we can destroy & dissassociate containers but maintain urls on relaunch of said Tools/containers
```

TODO:
- [ ] Display useful info in the ui: launch params, input files, .etc
- [x] Be able to relaunch a visualization #2433
- [x] Visualizations tab that includes a listing of visualizations that have been launched on a given dataset #2437
- [x] Tweak tools api to be able to filter on just visualiztions #2437
- [x] Track status: Running/not running etc. https://github.com/refinery-platform/refinery-platform/pull/2433

jkmarx:
Resolves #2260


refinery/tool_manager/models.py
```diff
@@ -400,6 +416,33 @@ def update_file_relationships_with_urls(self):
             )
         self.set_tool_launch_config(tool_launch_config)
 
+    def is_running(self):
+        if self.is_workflow():
+            return True if self.analysis.running() else False
```
mccalluc: Don't need the bracketing if-else. (Or use `bool()` if the point really is to cast.)

refinery/tool_manager/models.py
```diff
@@ -400,6 +416,33 @@ def update_file_relationships_with_urls(self):
...
+
+        try:
+            self._django_docker_client.lookup_container_url(
+                self.container_name
+            )
+            return True
+        except NotFound:
+            return False
```
mccalluc: Sorry: There should be a better method for this. I've filed https://github.com/refinery-platform/django_docker_engine/issues/94, but obviously not a blocker for this PR.

refinery/tool_manager/views.py
```diff
@@ -54,41 +46,82 @@ def list(self, request, *args, **kwargs):
...
+    http_method_names = ['get']
+
+    def get_queryset(self):
+        if self.request.user.has_perm('core.share_dataset', self.data_set):
+            return ToolDefinition.objects.all()
+
+        elif self.request.user.has_perm('core.read_dataset', self.data_set):
+            return ToolDefinition.objects.filter(
+                tool_type=ToolDefinition.VISUALIZATION
+            )
```
mccalluc: Is the fall-through case ok or should there be an error returned?

refinery/ui/source/js/commons/services/time.js
```diff
@@ -0,0 +1,58 @@
...
+      }
+
+      if (sec < 10) {
+        sec = '0' + sec;
+      }
+
+      var dateStr = year + '-' + month + '-' + day;
+      var timeStr = '@' + hour + ':' + mins + ':' + sec;
+
+      return (dateStr + timeStr);
```
mccalluc: Did you consider  just using `toISOString()`?

refinery/ui/source/js/data-set-visualization/partials/visualization.html
```diff
@@ -0,0 +1,122 @@
...
+      <div class="col-md-1">
+       <a class="analysis-header">
+         <strong>Creation Time</strong>
+         </span>
+       </a>
+      </div>
+      <div class="col-md-2">
+        <strong>Owner Name</strong>
+      </div>
+      <div class="col-md-2">
```
mccalluc: What's the empty div for? If it's filler, I think the column count is still only 11, not 12.

refinery/ui/source/js/data-set-visualization/partials/visualization.html
```diff
@@ -0,0 +1,122 @@
...
+              <span ng-if="!vis.is_running">
+                <i class="fa fa-pause-circle analyses-status"
+                 refinery-tooltip
+                 refinery-tooltip-container="body"
+                 refinery-tooltip-placement="right"
+                 title="Visualization is paused.">
+                </i>
+                Paused
+              </span>
+          </div>
```
mccalluc: check nesting.. I think this corresponds to `span[class=spinner]` above.

refinery/ui/source/js/data-set-visualization/partials/visualization.html
```diff
@@ -0,0 +1,122 @@
...
+           <div class="col-md-2">
+              <a ng-href="/users/{{ vis.owner.user_profile_uuid }}">
+                <!-- avoid error with white spaces or single letter names-->
+                <span ng-if="vis.owner.full_name.length > 2">
+                  {{ vis.owner.full_name }}
+                </span>
+                <span ng-if="!vis.owner.full_name.length < 2">
+                  {{ vis.owner.username }}
+                </span>
```
mccalluc: What if vis.owner.full_name.length == 2? And check order of operations:
```
> f = 'foo'
> f.length > 4
false
> !f.length > 4
false
```

refinery/ui/source/js/data-set-visualization/partials/visualization.html
```diff
@@ -0,0 +1,122 @@
...
+      <ul class="pagination analyses">
+        <li ng-class="prevPageDisabled()">
+          <a ng-click="prevPage()">«</a>
+        </li>
+        <li ng-repeat="pageNum in range()" ng-class="{active:pageNum==currentPage}"
+          ng-click="setPage(pageNum)">
+          <a> {{pageNum + 1}}  </a>
+        </li>
+        <li ng-class="nextPageDisabled()">
+          <a href ng-click="nextPage()">»</a>
```
mccalluc: Does the href belong here? Expecting this to match prevPage above.

jkmarx:
@mccalluc Sorry, I thought you finished reviewing. My mistake. Thanks for catching those, I will fix in a separate pull request and merge.

refinery/tool_manager/views.py
jkmarx: @mccalluc Good point. Will discuss with @scottx611x.

PR #2443, opened 2017-12-18
Set EC2 instance root volume type to gp2

### LINKED ISSUE #2442: EC2 data volume type is set to magnetic
* Specific code commit: v1.6.1
* Environment where the error occurred: AWS

### Steps to reproduce
Create a Refinery stack

### Observed behavior
EC2 instance root volume type is set to Magnetic (standard)

### Expected behavior
EC2 instance root volume type is set to General Purpose SSD (gp2)

hackdna:
Fix #2442

### DIFF
--- a/deployment/stack.py
+++ b/deployment/stack.py
@@ -195,8 +195,9 @@ def make_template(config, config_yaml):
                 {
                     'DeviceName': '/dev/sda1',
                     'Ebs': {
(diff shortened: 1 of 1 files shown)


hackdna:
This is for the root volume. It was always of type gp2 which is the recommended type for root/boot volumes (https://aws.amazon.com/ebs/details/).

PR #2446, opened 2017-12-19
Scottx611x/parameter updates

scottx611x:
- Fix bug when casting boolean parameters
- Add test coverage

### DIFF
--- a/refinery/tool_manager/models.py
+++ b/refinery/tool_manager/models.py
@@ -87,7 +87,7 @@ class Parameter(models.Model):
         if self.value_type in self.STRING_TYPES:
             return str(parameter_value)
         elif self.value_type == self.BOOLEAN:
-            return bool(parameter_value)
+            return True if parameter_value.lower() == "true" else False
         elif self.value_type == self.INTEGER:
             return int(parameter_value)
         elif self.value_type == self.FLOAT:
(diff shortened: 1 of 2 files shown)


refinery/tool_manager/models.py
```diff
@@ -87,7 +87,7 @@ def cast_param_value_to_proper_type(self, parameter_value):
         if self.value_type in self.STRING_TYPES:
             return str(parameter_value)
         elif self.value_type == self.BOOLEAN:
-            return bool(parameter_value)
+            return True if parameter_value.lower() == "true" else False
```
mccalluc: You don't need the bracketing if-then... but is this the right direction? I had thought we were moving towards allowing native type, rather than away?

refinery/tool_manager/models.py
```diff
@@ -469,9 +469,13 @@ def _get_visualization_parameters(self):
                 {
                     "uuid": parameter.uuid,
                     "description": parameter.description,
-                    "default_value": parameter.default_value,
+                    "default_value": parameter.cast_param_value_to_proper_type(
+                        parameter.default_value
```
mccalluc: Again, I'm a little confused: I had thought this was moving in a direction to support native types?

mccalluc: (Scott explained in person: This is coming from the database, where everything is string.)

refinery/tool_manager/tests.py
```diff
@@ -3473,3 +3473,75 @@ def test_get_workflows_with_connection_error(self):
...
+            default_value="1.0"
+        )
+        self.assertEqual(
+            float,
+            type(
+                parameter.cast_param_value_to_proper_type(
+                    parameter.default_value
+                )
+            )
+        )
```
mccalluc: In the tests, I think it might be better just to pass in plain values, rather than pulling `parameter.default_value`: Imagine a scenario where default_value starts returning something that's already been cast to the right type, and then at some later point cast_param stops casting: These tests might not catch either of those changes.

Does the ParameterFactory require all those fields, or could you make it more concise for the sake of readablity? Nothing by `value_type` really matters here...

And maybe just do the assertEqual on the value, instead of the type of the value? Shorter, and stronger. (Though I get that type is the thing you're really interested in here.)

scottx611x:
@mccalluc sorry I didn't realize you were still reviewing. I'll make changes in #2447 after reading your comment

PR #2447, opened 2017-12-19
Scottx611x/expose extra directories

scottx611x:
@mccalluc Please review/merge #2446 Before this one

- Add `extra_directories` information to `input.json`

Note: [`Heatmap Scatterplot`](https://github.com/mccalluc/heatmap-scatter-dash/) Tool launches properly with these changes, but we run into the relative paths issue again (when specifying a relative path in a VisTool things get routed through Refinery instead of the proxy) . We should really begin looking into using[ subdomain based routing for Vis Tools ](https://github.com/refinery-platform/django_docker_engine/issues/71)

### DIFF
--- a/refinery/tool_manager/models.py
+++ b/refinery/tool_manager/models.py
@@ -431,10 +431,8 @@ class VisualizationTool(Tool):
             self.FILE_RELATIONSHIPS: self.get_file_relationships_urls(),
             ToolDefinition.PARAMETERS: self._get_visualization_parameters(),
             self.NODE_INFORMATION: self._get_detailed_input_nodes_dict(),
-            ToolDefinition.EXTRA_DIRECTORIES: (
+            ToolDefinition.EXTRA_DIRECTORIES:
                 self.tool_definition.get_extra_directories()
-            )
-
         }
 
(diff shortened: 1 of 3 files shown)


refinery/tool_manager/models.py
```diff
@@ -428,7 +430,11 @@ def _create_container_input_dict(self):
         return {
             self.FILE_RELATIONSHIPS: self.get_file_relationships_urls(),
             ToolDefinition.PARAMETERS: self._get_visualization_parameters(),
-            self.NODE_INFORMATION: self._get_detailed_input_nodes_dict()
+            self.NODE_INFORMATION: self._get_detailed_input_nodes_dict(),
+            ToolDefinition.EXTRA_DIRECTORIES: (
+                self.tool_definition.get_extra_directories()
+            )
```
mccalluc: I don't think you need the surrounding paren here: There's no syntactic ambiguity without it.

mccalluc: I don't think this should block, but the mix of `self` and `ToolDefinition`, or public and underscored confuses me, a little. Are those distinctions useful, and meaningful?

scottx611x:
Failing build See: #2348

PR #2449, opened 2017-12-20
Jkmarx/fixes for vis tab

jkmarx:
Ref #2437 
- Addresses comments by @mccalluc


refinery/ui/source/js/data-set-visualization/partials/visualization.html
```diff
@@ -74,7 +72,7 @@
                 <span ng-if="vis.owner.full_name.length > 2">
                   {{ vis.owner.full_name }}
                 </span>
-                <span ng-if="!vis.owner.full_name.length < 2">
+                <span ng-if="!(vis.owner.full_name.length < 2)">
```
mccalluc: I'm still confused here... Is the intention to have this and the condition above be mutually exclusive?

jkmarx: @mccalluc Gotcha. Thanks

PR #2473, opened 2018-01-17
Remove slug edit from import form.

### LINKED ISSUE #1629: Move "URL shortcut" field from upload page to data set details in file browser
It is not necessary to set this during upload and is distracting to users without prior experience with the system.

This feature should be located next to edit features for other data set details, such as title, description etc.

_This is dependent on availability of other data set detail editing capabilities in the data set preview and new file browser._

jkmarx:
Resolves #1629


hackdna:
Slug handling should also be removed from the corresponding server side code, for example: https://github.com/refinery-platform/refinery-platform/blob/e53a8a577f460107a8bcefd3044d4ea4d7326f1c/refinery/data_set_manager/views.py#L540

PR #2474, opened 2018-01-17
Remove slug reference when importing meta data form.

jkmarx:
Ref #1629

### DIFF
--- a/refinery/data_set_manager/single_file_column_parser.py
+++ b/refinery/data_set_manager/single_file_column_parser.py
@@ -297,7 +297,6 @@ def process_metadata_table(
     annotation_column=None,
     sample_column=None,
     assay_column=None,
-    slug=None,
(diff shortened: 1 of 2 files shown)


refinery/data_set_manager/single_file_column_parser.py
```diff
@@ -297,7 +297,6 @@ def process_metadata_table(
     annotation_column=None,
     sample_column=None,
     assay_column=None,
-    slug=None,
```
hackdna: `process_metadata_table` management command calls `process_metadata_table()` function with the slug argument: https://github.com/refinery-platform/refinery-platform/blob/8a8c6fa301e170d88b0ac6b7d4af44d55f7691c2/refinery/data_set_manager/management/commands/process_metadata_table.py#L100

PR #2490, opened 2018-01-18
Scottx611x/fix 2481

### LINKED ISSUE #2481: Incorrect REFINERY_DOWNLOAD_URL_s displayed for File Nodes without an associated file_import task
* Specific code commit: 23515c12778dd91e10fc78bbafff0ea1a4947ca2
* Version of the web browser and OS: Any
* Environment where the error occurred: https://test.stemcellcommons.org

### Steps to reproduce
* Bring up a stack referencing the latest beta site snapshots
* [Query Solr's `data_set_manger` core for `Nodes` with `REFINERY_DOWNLOAD_URL_s:PENDING` ](https://test.stemcellcommons.org/solr/data_set_manager/select/?fq=REFINERY_DOWNLOAD_URL_s:PENDING&q=django_ct:data_set_manager.node&wt=json)

### Observed behavior
* 31,812/37,545 `Nodes` have their `REFINERY_DOWNLOAD_URL_s == PENDING`
* None of these `Nodes` has a corresponding `file_import` task
* None of these `Nodes` point to a valid `FileStoreItem.datafile`
* None of these `Nodes` have a valid `FileStoreItem.source`
* See the above info here: [pending_nodes_filestore_item_info.txt](https://github.com/refinery-platform/refinery-platform/files/1643341/bad.txt)

### Expected Behavior
* We should see the lightning bolt icon next to these Nodes

### Questions
* Why are there so many of these `Nodes`?
* Can we get rid of these `Nodes`?
* How should the `data_set_manager.search_indexes.NodeIndex.prepare` be updated to accomodate these `Nodes`?

### Other Thoughts
* We currently don't run our `DataSet` uploads within the scope of db transactions so its possible that many of these `Nodes` exist due to failed DataSet uploads. (Also many of the corresponding `FileStoreItem's sources` refer to `/data/import/` which led me to believe this as well).

scottx611x:
Fixes #2481

TODO: 
- [x] @jkmarx Also observed similar behavior with ISATabs that reference local data so handle that case too
- [x] Test on AWS instance with beta site data
- [x] Add test coverage


refinery/data_set_manager/search_indexes.py
```diff
@@ -141,14 +143,43 @@ def prepare(self, object):
...
+                        except TaskMeta.DoesNotExist:
+                            logger.debug(
+                                "No file_import task for FileStoreItem with "
+                                "UUID: %s",
+                                file_store_item.uuid
+                            )
+                            download_url = NOT_AVAILABLE
+                        else:
+                            download_url = PENDING
+                else:
```
mccalluc: If the `else` is much shorter than the `then` consider swapping them: Easier to understand where it's coming from that way.

refinery/data_set_manager/tests.py
```diff
@@ -2000,36 +2006,71 @@ def test_prepare_node_remote_datafile_source(self):
...
-    def test_prepare_node_pending_file_import_task(self):
+    def test_prepare_node_pending_yet_existing_file_import_task(self):
+        self.file_store_item.datafile.delete()
+        self.file_store_item.save()
+        with mock.patch.object(
+            FileStoreItem,
+            "get_import_status",
+            return_value=PENDING
+        ):
+            self._prepare_index(self.node, expected_download_url=PENDING)
```
mccalluc: Do you want to make an assertion in this test? Or maybe that's in `_prepare_index`?

Issue #2529, opened 2018-01-23
`load_tools --visualizations` Error message could be more helpful


scottx611x:
* Specific code commit: c8a9cb3f
* Version of the web browser and OS: Any
* Environment where the error occurred: Any

### Steps to reproduce
Please list all the actions and the input data used:

1. Run `./manage.py load_tools --visualizations non_existent_tool_name`

### Observed behavior
* The current Error message looks like:
```
CommandError: "non_existent_tool_name" not a file and "https://raw.githubusercontent.com/refinery-platform/visualization-tools/master/tool-annotations/non_existent_tool_name.json" not a valid URL
```

### Expected behavior
* A more helpful error message
* Something along the lines of:
```
CommandError: "non_existent_tool_name" not a local file path and "https://raw.githubusercontent.com/refinery-platform/visualization-tools/master/tool-annotations/non_existent_tool_name.json" does not point to a valid Visualization Tool Registry URL.
 Available Visualization Tools from the Registry are: heatmap-scatterplot, higlass, igv, refinery-developer-tool
```
A user would then immediately have an idea of what Visualizations Tools are available to them.

PR #2530, opened 2018-01-24
Scottx611x/better load tools error message

scottx611x:
Fixes: #2529 

Error message displayed is now:
```
CommandError: "blah" not a local file path and "https://raw.githubusercontent.com/refinery-platform/visualization-tools/master/tool-annotations/blah.json" does not point to a valid Visualization Tool Registry URL.
 Available Visualization Tools from the Registry (https://github.com/refinery-platform/visualization-tools/) are: heatmap-scatterplot, higlass, igv, refinery-developer-tool
```

### DIFF
--- a/refinery/tool_manager/management/commands/load_tools.py
+++ b/refinery/tool_manager/management/commands/load_tools.py
@@ -3,6 +3,7 @@ from optparse import make_option
 import os
 import sys
 
+from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
 
 import requests
@@ -114,6 +115,9 @@ class Command(BaseCommand):
 
     def _load_visualization_definitions(self, names, branch='master'):
         visualization_annotations = []
+        availible_registry_tool_names = \
+            self._get_available_visualization_tool_registry_names()
+
         for name in names:
             if os.path.exists(name):
                 with open(name) as f:
@@ -125,8 +129,13 @@ class Command(BaseCommand):
                 response = requests.get(url)
                 if response.status_code != 200:
                     raise CommandError(
-                        '"{}" not a file and "{}" not a valid URL'.format(
-                            name, url
+                        '"{}" not a local file and "{}" does not point to'
+                        ' a valid Visualization Tool Registry URL.\n '
+                        'Available Visualization Tools from the '
+                        'Registry ({}) are: {}'.format(
+                            name, url,
+                            settings.REFINERY_VISUALIZATION_TOOL_REGISTRY_URL,
+                            availible_registry_tool_names
                         )
                     )
                 annotation = response.json()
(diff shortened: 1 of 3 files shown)


refinery/tool_manager/management/commands/load_tools.py
```diff
@@ -112,21 +116,32 @@ def _confirmation_loop(self):
...
                     raise CommandError(
-                        '"{}" not a file and "{}" not a valid URL'.format(
-                            name, url
+                        '"{}" not a local file path and "{}" does not point to'
+                        ' a valid Visualization Tool Registry URL.\n '
+                        'Available Visualization Tools from the '
+                        'Registry ({}) are: {}'.format(
+                            name, raw_asset_url,
+                            settings.REFINERY_VISUALIZATION_TOOL_REGISTRY_URL,
+                            availible_registry_tool_names
```
mccalluc: maybe use the kwargs form? your call.
```python
>>> 'test: {f}'.format(f='foo')
'test: foo'
```

refinery/tool_manager/management/commands/load_tools.py
```diff
@@ -198,6 +213,36 @@ def _generate_workflows(self):
...
+                    "tree/{}/tool-annotations".format(
+                        self.visualization_registry_branch
+                    )
+                )
+            )
+        except requests.exceptions.RequestException as e:
+            raise CommandError(
+                "Unable to fetch Visualization Tools from the Registry "
+                "({}): {}".format(
+                    settings.REFINERY_VISUALIZATION_TOOL_REGISTRY_URL,
```
mccalluc: Maybe the constant could be shorter?: `VIS_REGISTRY` might be too laconic, but gets the idea across. Your call.

refinery/tool_manager/management/commands/load_tools.py
```diff
@@ -198,6 +213,36 @@ def _generate_workflows(self):
...
+                    e
+                )
+            )
+        else:
```
mccalluc: Since the `except` raises, I don't think the `else` is needed here: you could get rid of it and just drop one level of indentation.

refinery/tool_manager/management/commands/load_tools.py
```diff
@@ -198,6 +213,36 @@ def _generate_workflows(self):
...
+            return ", ".join(
+                [
+                    tool_name.split("/")[1].replace(".json", "")
+                    for tool_name in re.findall(
+                        r"tool-annotations/.+?\.json",
+                        response.content
+                    )
+                ]
```
mccalluc: DRY: capture the substring you're interested in:
```python
>>> content = 'lorem tool-annotations/foo.json ipsum tool-annotations/bar.json !!!'
>>> ', '.join(re.findall(r"tool-annotations/(.+?)\.json", content))
'foo, bar'
```

refinery/tool_manager/tests.py
```diff
@@ -267,9 +267,12 @@ def setUp(self):
...
         )
 
-    def load_visualizations(self):
+    def load_visualizations(self, visualizations=None):
         # TODO: More mocking, so Docker image is not downloaded
-        visualizations = ["{}/visualizations/igv.json".format(TEST_DATA_PATH)]
+        if visualizations is None:
+            visualizations = [
+                "{}/visualizations/igv.json".format(TEST_DATA_PATH)
+            ]
```
mccalluc: Maybe move this into the signature? So it's clear just looking at that what the default is.

Issue #2531, opened 2018-01-24
Avoid relying on data migrations in Python tests


hackdna:
This would:
* speed up test runs significantly by not applying migrations for every test run
* make tests less brittle by removing dependence on local database contents (see also #2235)

Test data can be easily created using Factory Boy.

PR #2536, opened 2018-01-24
Add a Django config for running tests

hackdna:


hackdna:
No, this does not work with any tests that rely on data included in migrations but that should be avoided anyway (#2531). However, this change doesn't affect anything unless used explicitly.

It does make the test runs much faster and test result output a lot more concise: actual test failures are still displayed and are much easier to see because logging is suppressed.

Issue #2393, opened 2017-11-27
Remove file_server Django app


hackdna:
It appears to be no longer in use.

* Database migration that removes the corresponding database instances will not exist after the app is deleted
* Search for usages of all symbols in `file_server/models.py` plus `file_server` string
* Remove corresponding Django HTML templates
* There are 30 `BAMItem` instances on beta.stemcellcommons.org
* Remove stale Django `ContentType` instances

Issue #2464, opened 2018-01-12
Remove visualization_manager Django app


mccalluc:
We think VisualizationManager was not the first iteration, but it has been supplanted by all the django_docker_proxy stuff. It should be removed.

It appears to be unused, except for `igv_multi_species()` function in [core/views.py](https://github.com/refinery-platform/refinery-platform/blob/develop/refinery/core/views.py#L628)

See also #2393

hackdna:
I still see this app present in the codebase (d87a66f).

jkmarx:
@hackdna It was resolved by the remove_solr_igv dependencies which is merged in the django 1.8 release branch. It should be merged to develop this week.

hackdna:
Shouldn't issues be closed when the code is merged into the develop branch?

PR #2541, opened 2018-01-25
Remove file server

hackdna:
Toward #2393 

Note: the rest of the code will be removed in the next release (to make sure that corresponding database tables are removed during migration in all instances like production, etc)


refinery/annotation_server/management/commands/add_genome_annotation.py
```diff
@@ -16,7 +16,6 @@
 # Helper command to deal with additional annotation tracks not available from
 # UCSC genome browser
 #
-# TODO: Should we import annotation files into file_server?
 # how to cleanup directories used to store and download files
```
mccalluc: This line is probably a TODO in its own right?

refinery/visualization_manager/views.py
```diff
@@ -535,16 +526,6 @@ def get_file_name(nodeuuid, samp_file=None, is_file_uuid=False):
                 FileStoreItem.MultipleObjectsReturned) as e:
             logger.error("Could not fetch FileStoreItem from %s: %s",
                          nodeuuid, e)
```
mccalluc: With this change, have our expectations around `is_file_uuid` changed? As it was, either temp_fs would have been set, or there would have been an error, and that's no longer the case.

refinery/annotation_server/management/commands/add_genome_annotation.py
hackdna: I was confused about this line. Should I prefix it with a TODO?

refinery/visualization_manager/views.py
hackdna: This whole function (`create_igv_session`) appears to be used only by `igv_session` view which in turn appears to be unused (it looks like it was used for Java-based IGV).

scottx611x: We definitely don't need `create_igv_session` anymore I'd remove it and any usages as well

hackdna:
@scottx611x We'll need to remove file_server entries from the django_migrations table (`manage.py migrate --fake file_server zero`) from production and other instances after this release is deployed. Then the next release would contain the removal of the rest of the file_server code.

I've decided to keep `create_igv_session` to limit the scope of this PR - it will be removed as part of #2464.

PR #2562, opened 2018-02-06
Jkmarx/registration redux upgrade

### LINKED ISSUE #2561: Registration Redux Upgrade
1.3 introduced a breaking change with register()
Seems worth spending the time to upgrade to 1.9 (last release before <1.11 support)

jkmarx:
Resolve #2561

### DIFF
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -926,7 +926,7 @@ class AnalysesViewSet(APIView):
 
 class CustomRegistrationView(RegistrationView):
 
-    def register(self, request, **cleaned_data):
+    def register(self, request):
         """
         Given a username, email address, password, first name, last name,
         and affiliation, register a new user account, which will initially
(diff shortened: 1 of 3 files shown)


refinery/core/views.py
```diff
@@ -974,7 +974,7 @@ class of this backend as the sender.
                                      request=request)
         return new_user
 
-    def get_success_url(self, request, user):
+    def get_success_url(self, user):
```
mccalluc: Did the signature change between versions? Would a test that breaks under the old but works with the new be appropriate?

refinery/templates/core/about.html
```diff
@@ -1,6 +1,6 @@
 {% extends "base.html" %}
 {% load markup %}
-{% load flatblock_tags %}
```
scottx611x: Are the `flatblocks` and `registration` apps intertwined somehow? Or are these just changes due to the `flatblocks` upgrade in `jkmarx/django-upgrade-1.8`

refinery/core/views.py
jkmarx: We upgraded registration_redux from 1.1 to 1.9...the signature changed.

refinery/templates/core/about.html
jkmarx: They are not intertwined. I just didn't make another pull request for flatblocks. The changes are due to the apps upgrade.

Issue #2413, opened 2017-12-07
VisualizationTool launches should have access to **all** of a Dataset's Node metadata information


scottx611x:
The [current implementation](https://github.com/refinery-platform/refinery-platform/blob/master/refinery/tool_manager/models.py#L423-L432) only sends `Node` information over for `Nodes` that were selected in the Tool Launch UI. 

Out of @mccalluc & @ngehlenborg  conversation today we now want to send over all the Node info related to the Dataset.

Implementation Notes:
- [ ] `file_relationship_urls` shouldn't be used in `_create_container_input_dict` anymore since we have access to file url info through Nodes
- [ ] We should still preserve the file relationship structure of the `file_relationships` key, except with Node uuids instead of urls, and the name in json should change. (`node_relationships`?)
- [ ] `update_file_relationships_with_urls` should be refactored to update the file relationships with `Node` uuids only , and should be renamed (`update_file_relationships_with_node_uuids` ?)
- [ ] Some of our existing Vis Tools probably won't like these changes so those should be updated to work properly

#### Edit: Not worrying so much about these checkboxes as theres no concrete use case for all of this extra Node information atm, and addressing the checkboxes would cause lots of backwards incompatibility with current vis tools. [Getting all of the DataSet's Node info into the input.json](https://github.com/refinery-platform/refinery-platform/pull/2569) should suffice for now.

scottx611x:
Ping @mccalluc in case I missed anything

mccalluc:
Looks good; one small edit.

PR #2569, opened 2018-02-07
Scottx611x/send over all dataset node info

scottx611x:
Fixes: #2413 

@mccalluc There would be some backwards compatibility issues with existing VisualizationTools if all of the checkboxes from #2413 were addressed. It still feels kind of dirty to me to send all this extra information over without a concrete use case for it (that I know of). 

Would like to have a quick chat about this again, no rush though.

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -833,12 +833,14 @@ class DataSet(SharableResource):
         else:
             return False
 
-    def get_nodes(self, uuids_only=False):
-        nodes = Node.objects.filter(
+    def get_nodes(self):
+        return Node.objects.filter(
             assay=self.get_latest_assay(),
             study=self.get_latest_study()
         )
-        return nodes.values_list('uuid', flat=True) if uuids_only else nodes
+
+    def get_node_uuids(self):
+        return self.get_nodes().values_list('uuid', flat=True)
 
 
 @receiver(pre_delete, sender=DataSet)
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -833,6 +833,13 @@ class DataSet(SharableResource):
         else:
             return False
 
+    def get_nodes(self, uuids_only=False):
+        nodes = Node.objects.filter(
+            assay=self.get_latest_assay(),
+            study=self.get_latest_study()
+        )
+        return nodes.values_list('uuid', flat=True) if uuids_only else nodes
+
 
 @receiver(pre_delete, sender=DataSet)
 def _dataset_delete(sender, instance, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


refinery/core/models.py
```diff
@@ -833,6 +833,13 @@ def is_valid(self):
         else:
             return False
 
+    def get_nodes(self, uuids_only=False):
```
hackdna: It's probably best to avoid flag arguments in favor of separate functions, especially if the return type is different (like in this case)
https://softwareengineering.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior
https://martinfowler.com/bliki/FlagArgument.html

refinery/tool_manager/models.py
```diff
@@ -485,17 +489,20 @@ def _check_max_running_containers(self):
         if len(self._django_docker_client.list()) >= max_containers:
             raise VisualizationToolError('Max containers')
 
-    def _get_detailed_input_nodes_dict(self):
+    def _get_detailed_nodes_dict(self, fetch_all_nodes=False):
```
hackdna: It's probably best to avoid flag arguments in favor of separate functions
https://softwareengineering.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior
https://martinfowler.com/bliki/FlagArgument.html

refinery/core/tests.py
```diff
@@ -1543,6 +1543,27 @@ def test_get_latest_study_no_studies(self):
...
+        [self.assertIn(node, self.dataset.get_nodes()) for node in nodes]
+
+    def test_get_nodes_no_nodes_available(self):
+        Node.objects.all().delete()
+        self.assertQuerysetEqual(self.dataset.get_nodes(), [])
+
+    def test_get_node_uuids(self):
+        node_uuids = Node.objects.all().values_list("uuid", flat=True)
+        self.assertGreater(node_uuids.count(), 0)
+        [
```
scottx611x: why not loop?

refinery/core/tests.py
```diff
@@ -1543,6 +1543,27 @@ def test_get_latest_study_no_studies(self):
             dataset.get_latest_study()
             self.assertIn("Couldn't fetch Study", context.exception.message)
 
+    def test_get_nodes(self):
+        nodes = Node.objects.all()
+        self.assertGreater(nodes.count(), 0)
+        [self.assertIn(node, self.dataset.get_nodes()) for node in nodes]
```
scottx611x: Why not loop?

refinery/tool_manager/tests.py
```diff
@@ -283,20 +283,21 @@ def tearDown(self):
             django_docker_cleanup()
         super(ToolManagerTestBase, self).tearDown()
 
-    def create_solr_mock_response(self, tool):
+    def create_solr_mock_response(self, tool, all_dataset_nodes=False):
```
scottx611x: defaults should be defaults

PR #2577, opened 2018-02-08
Update user/file facet/columns per user request.

### LINKED ISSUE #2458: Add columns/facets to cross filebrowser to meet Shivdasani lab needs
Talking with @gmnelson: Currently we have a column for the filename. Instead:
- Have a column for the sample name: This should be filled in, and is more likely to be useful: The filename could have been chosen arbitrarily by the processing center.
- Have a column for the original local path: This is actually what the lab needs, since they will only be using it for metadata.
- Add a filetype facet, and a column, because a facet without a column might be odd?
- and *remove* the current filename column: It is redundant with path.

And:
- Change "Date" to "Date Submitted" (but there seems to be some deeper problem with the rendering of the Date column? #2350)

### LINKED ISSUE #2459: Add "cell type" to cross dataset browser
Talking with @gmnelson, it seems like adding cell-type to the cross dataset browser could be useful.

jkmarx:
Resolves #2458 and Resolves #2459


jkmarx:
@gmnelson yes

Issue #449, opened 2014-09-12
Upgrade to Django 1.8


hackdna:
https://docs.djangoproject.com/en/1.8/howto/upgrade-version/
https://docs.djangoproject.com/en/1.8/releases/

-*Moved UUIDField transfer to own issue*

- [x] Switch to using:
* https://docs.djangoproject.com/en/1.8/ref/settings/#secure-ssl-redirect
* from optparse to argparse

- [x] Update django-storages to the latest version that supports Django 1.8
     The latest supports 1.8 but has dropped it for 1.9 in version 1.6.1

It might be useful to start a brand new 1.8 project using `startproject` management command to check what has changed in the project structure, settings, wsgi.py, etc

scottx611x:
**Deprecation checks for upgrading Django 1.7 to 1.8**
- [ ] django.contrib.comments will be removed.
- [ ] The following transaction management APIs will be removed:
TransactionMiddleware,
the decorators and context managers autocommit, commit_on_success, and commit_manually, defined in django.db.transaction,
the functions commit_unless_managed and rollback_unless_managed, also defined in django.db.transaction,
the TRANSACTIONS_MANAGED setting.
- [ ] The cycle and firstof template tags will auto-escape their arguments. In 1.6 and 1.7, this behavior is provided by the version of these tags in the future template tag library.
- [ ] The SEND_BROKEN_LINK_EMAILS setting will be removed. Add the django.middleware.common.BrokenLinkEmailsMiddleware middleware to your MIDDLEWARE_CLASSES setting instead.
- [ ] django.middleware.doc.XViewMiddleware will be removed. Use django.contrib.admindocs.middleware.XViewMiddleware instead.
Model._meta.module_name was renamed to model_name.
- [ ] Remove the backward compatible shims introduced to rename get_query_set and similar queryset methods. This affects the following classes: BaseModelAdmin, ChangeList, BaseCommentNode, GenericForeignKey, Manager, SingleRelatedObjectDescriptor and ReverseSingleRelatedObjectDescriptor.
Remove the backward compatible shims introduced to rename the attributes ChangeList.root_query_set and ChangeList.query_set.
- [ ] django.views.defaults.shortcut will be removed, as part of the goal of removing all django.contrib references from the core Django codebase. Instead use django.contrib.contenttypes.views.shortcut. django.conf.urls.shortcut will also be removed.
[...]

PR #2586, opened 2018-02-12
Update config with the new 1.8 setting.

jkmarx:
Ref #449


jkmarx:
@hackdna @scottx611x Scott and I were talking and he mentioned removing the REFINERY_URL_SCHEME setting. It's used in the core/utils/get_full_url
Thoughts?

hackdna:
REFINERY_URL_SCHEME is needed to construct correct absolute URLs outside of the web browser context (for example, in emails).
Btw, I have already edited `get_full_url()` for clarity in https://github.com/refinery-platform/refinery-platform/compare/hackdna/refactor_file_store.

PR #2590, opened 2018-02-14
Scottx611x/handle lack of galaxy input step inputs

scottx611x:
Workflows created with our previous Galaxy version would include information about their input step like this in their dict representation:
```
"0": {
      "inputs": [
        {
          "name": "Input Dataset Collection",
          "description": ""
        }
      ],
      "type": "data_collection_input",
      "id": 0,
      "name": "Input dataset collection"
    }
}
```

I observed while updating our Test Workflows on the new 17.05 cluster a structure like this:
```
"0": {
      "inputs": [],
      "type": "data_collection_input",
      "id": 0,
      "name": "Input dataset collection"
    }
}
```
Which would yield an `IndexError` at `inputs`here:
```
parent_node_output_name = (
    steps[str(parent_node_id)]['inputs'][0]['name']
)
```

This is backwards compatible with our existing Workflows


scottx611x:
@hackdna when reviewing this it may be best to look at https://github.com/refinery-platform/refinery-platform/pull/2590/commits/8960a3e96a868f828e95ec43e44739656b25fa9a first since its easy to see the meaningful change there.

refinery/tool_manager/utils.py
```diff
@@ -10,6 +10,7 @@
 from bioblend.galaxy.client import ConnectionError
 from django_docker_engine.docker_utils import DockerClientWrapper
 from jsonschema import RefResolver, ValidationError, validate
+import networkx as nx
```
hackdna: Just `import networkx` is sufficient

refinery/tool_manager/utils.py
```diff
@@ -507,3 +508,60 @@ def validate_tool_launch_configuration(tool_launch_config):
...
+    # input_connections, i.e. only by looking at tool nodes)
+    for current_node_id, step in steps.iteritems():
+        # ensure node id is an integer
+        current_node_id = int(current_node_id)
+        input_connections = step['input_connections'].iteritems()
+        for current_node_input_name, input_connection in input_connections:
+            parent_node_id = input_connection["id"]
+            # test if parent node is a tool node or an input node to pick the
+            # right name for the outgoing edge
+            if parent_node_id == 0:
```
hackdna: No more checking for `GALAXY_INPUT_TYPES`?

scottx611x: No, that ended up being overkill

Issue #2415, opened 2017-12-08
Refactor `Analysis.attach_outputs_dataset`


scottx611x:
I remember being in the trenches of [Analysis.attach_outputs_dataset](https://github.com/refinery-platform/refinery-platform/blob/master/refinery/core/models.py#L1519-L1657) a while back, and it not being a particularly good time. 

Now that there is nearly 100% test coverage I think it would be a good opportunity to refactor.

- [x] Rename method to be more clear
- [x] Pull out logic into smaller digestible chunks

PR #2591, opened 2018-02-14
Scottx611x/refactor attach outputs dataset

scottx611x:
I was in this area of the code today and figured it was high time for some refactoring
Fixes: #2415

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -1161,7 +1161,7 @@ class Analysis(OwnableResource):
         )
         ordering = ['-time_end', '-time_start']
 
-    def get_workflow_graph(self):
+    def get_expanded_workflow_graph(self):
         return tool_manager.utils.create_expanded_workflow_graph(
             ast.literal_eval(self.workflow_copy)
         )
@@ -1512,7 +1512,9 @@ class Analysis(OwnableResource):
 
     def attach_derived_nodes_to_dataset(self):
         graph_with_data_transformation_nodes = (
-            self._create_data_transformation_nodes(self.get_workflow_graph())
+            self._create_data_transformation_nodes(
+                self.get_expanded_workflow_graph()
+            )
         )
         graph_with_input_nodes_linked = (
             self._link_input_nodes_to_data_transformation_nodes(
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -1827,10 +1827,12 @@ class AnalysisNodeConnection(models.Model):
                                            max_length=250)
 
     def __unicode__(self):
-        return (
-            self.direction + ": " +
-            str(self.step) + "_" +
-            self.name + " (" + str(self.is_refinery_file) + ")"
+        return "{}: {}_{} ({}) {}".format(
+            self.direction,
+            self.step,
+            self.name,
+            self.is_refinery_file,
+            self.filetype
         )
 
     def get_input_connection_id(self):
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -1653,10 +1653,7 @@ class Analysis(OwnableResource):
         ]
         for data_transformation_node in data_transformation_nodes:
             # TODO: incorporate subanalysis id in tool name???
-            node_name = "{}_{}".format(
-                data_transformation_node['tool_id'],
-                data_transformation_node['name']
-            )
+            node_name = "{tool_id}_{name}".format(**data_transformation_node)
             data_transformation_node['node'] = (
                 Node.objects.create(
                     study=self.get_input_node_study(),
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -1161,6 +1161,12 @@ class Analysis(OwnableResource):
         )
         ordering = ['-time_end', '-time_start']
 
+    @property
+    def workflow_graph(self):
+        return tool_manager.utils.create_expanded_workflow_graph(
+            ast.literal_eval(self.workflow_copy)
+        )
+
     def delete(self, **kwargs):
         """
         Overrides the Analysis model's delete method and checks if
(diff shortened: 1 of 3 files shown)


refinery/core/models.py
```diff
@@ -1161,6 +1161,12 @@ class Meta:
         )
         ordering = ['-time_end', '-time_start']
 
+    @property
```
hackdna: What's the benefit of making this a property?

refinery/core/models.py
```diff
@@ -1505,145 +1511,17 @@ def rename_results(self):
...
-        assay = AnalysisNodeConnection.objects.filter(
-            analysis=self,
-            direction=INPUT_CONNECTION
-        ).first().node.assay
-        # 1. read workflow into graph
-        graph = tool_manager.utils.create_expanded_workflow_graph(
-            ast.literal_eval(self.workflow_copy)
+    def attach_derived_nodes_to_dataset(self):
+        graph_with_data_transformation_nodes = (
+            self._create_data_transformation_nodes(self.workflow_graph)
```
mccalluc: We just talked about this... probably more clear just to call?

refinery/core/models.py
scottx611x: No benefit, changing it now

refinery/core/models.py
```diff
@@ -1753,6 +1631,165 @@ def _create_derived_data_file_node(self, study,
...
+        data_transformation_nodes = [
+            graph.node[node_id] for node_id in graph.nodes()
+            if graph.node[node_id]['type'] == "tool"
+        ]
+        for data_transformation_node in data_transformation_nodes:
+            # TODO: incorporate subanalysis id in tool name???
+            node_name = "{}_{}".format(
+                data_transformation_node['tool_id'],
+                data_transformation_node['name']
+            )
```
mccalluc: maybe:
```python
"{tool_id}_{name}".format(**data_transformation_node)
```
and then it's almost short enough to just put in the `create` below.

refinery/core/models.py
```diff
@@ -1753,6 +1631,165 @@ def _create_derived_data_file_node(self, study,
...
+            analysis=self,
+            direction=INPUT_CONNECTION
+        )
+        for input_connection in input_node_connections:
+            for edge in graph.edges_iter([input_connection.step]):
+                input_id = input_connection.get_input_connection_id()
+
+                if graph[edge[0]][edge[1]]['output_id'] == input_id:
+                    input_node_id = edge[1]
+                    data_transformation_node = (
```
mccalluc: I would tend to just back-slash in a case like this... maybe to avoid the closing paren, or the risk of a stray comma turning it into a tuple, but this could be the wrong call.

refinery/core/models.py
```diff
@@ -1161,6 +1161,11 @@ class Meta:
         )
         ordering = ['-time_end', '-time_start']
 
+    def get_workflow_graph(self):
```
hackdna: `get_expanded_workflow_graph()`?

refinery/core/models.py
```diff
@@ -1753,6 +1631,165 @@ def _create_derived_data_file_node(self, study,
...
+                    input_connection.node.add_child(data_transformation_node)
+        return graph
+
+    def _create_derived_data_file_nodes(self, graph):
+        """create derived data file nodes for all entries and connect to data
+            transformation nodes"""
+        output_connection_to_analysis_result_mapping = (
+            self._get_output_connection_to_analysis_result_mapping()
+        )
+        output_mappings = output_connection_to_analysis_result_mapping
```
mccalluc: just use this name in the first place?

refinery/core/models.py
```diff
@@ -1753,6 +1631,165 @@ def _create_derived_data_file_node(self, study,
...
+                self.get_input_node_assay(),
+                output_connection
+            )
+            if output_connection.is_refinery_file:
+                # retrieve uuid of corresponding output file if exists
+                logger.info(
+                    "Results for '%s' and %s.%s: %s",
+                    self.uuid,
+                    output_connection,
+                    output_connection.filetype,
```
mccalluc: Below, we debug with just output_connection... maybe that's enough?

refinery/core/models.py
```diff
@@ -1753,6 +1631,165 @@ def _create_derived_data_file_node(self, study,
...
+            self,
+            graph,
+            output_connection,
+            derived_data_file_node
+    ):
+        """get graph edge that corresponds to this output node:
+        a. attach output node to source data transformation node
+        b. attach output node to target data transformation node
+        (if exists)"""
+        if len(graph.edges([output_connection.step])) > 0:
```
mccalluc: but if it were zero, then the loop below would just exit quickly?

refinery/core/models.py
```diff
@@ -1753,6 +1631,165 @@ def _create_derived_data_file_node(self, study,
...
+            for edge in graph.edges_iter([output_connection.step]):
+                output_id = output_connection.get_output_connection_id()
+
+                if graph[edge[0]][edge[1]]['output_id'] == output_id:
+                    input_node_id = edge[0]
+                    output_node_id = edge[1]
```
mccalluc: a lot like line 1681? maybe define the variables first?

refinery/core/models.py
```diff
@@ -1753,6 +1631,165 @@ def _create_derived_data_file_node(self, study,
...
+        if (output_connection.is_refinery_file and
+                derived_data_file_node.parents.count() == 0):
+            graph.node[output_connection.step]['node'].add_child(
+                derived_data_file_node
+            )
+        # delete output nodes that are not refinery files and don't have
+        # any children
+        if (not output_connection.is_refinery_file and
+                derived_data_file_node.children.count() == 0):
+            output_connection.node.delete()
```
mccalluc: maybe:
```python
if derived_data_file_node.children.count() == 0:
  if output_connection.is_refinery_file:
    ...
  else:
    ...
```

scottx611x: I thought we could make the split here too, but looking at theses two ifs one is worried about ` derived_data_file_node.parents` and the other ` derived_data_file_node.children`

refinery/core/models.py
scottx611x: I don't think the repr() of output_connection has the filetype information

hackdna: Update the repr?

PR #2597, opened 2018-02-15
Mccalluc/terraform ec2 for docker engine

mccalluc:


deployment/terraform/modules/docker_host/main.tf
```diff
@@ -0,0 +1,57 @@
...
+    from_port = 0
+    to_port = 0
+    protocol = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  tags {
+    Name = "${var.security_group_name}"
+  }
+}
```
hackdna: VPC config should be in a separate module ("network" or "vpc") - will be used by all resources of the stack (see #2600).

deployment/terraform/modules/docker_host/main.tf
```diff
@@ -0,0 +1,57 @@
...
+variable "security_group_id" {
+  type = "string"
+  default = "my_security_group_id"
+}
+
+variable "security_group_name" {
+  type = "string"
+  default = "allow_docker"
+}
+
```
hackdna: All variable declarations should be in vars.tf

deployment/terraform/modules/docker_host/main.tf
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  ingress {
+    from_port   = 0
+    to_port     = 80
+    protocol    = "tcp"
+    cidr_blocks = ["${var.cidr_block}"]
+  }
+
+  egress {  # Implicit with AWS, but Terraform requires that it be explicit:
```
hackdna: Does docker_host need to talk to everything?

deployment/terraform/modules/docker_host/main.tf
```diff
@@ -0,0 +1,57 @@
+variable "cidr_block" {
+  type = "string"
+  default = "10.0.0.0/16"
+}
+
+variable "security_group_id" {
+  type = "string"
+  default = "my_security_group_id"
```
hackdna: I'd simply assign a value in the corresponding `module` section of `live/main.tf` based on `${terraform.workspace}` ([example](https://github.com/refinery-platform/refinery-platform/blob/ff1167230df226f27f932f846452cdbd8f901dd0/deployment/terraform/live/main.tf#L25))

deployment/terraform/modules/docker_host/main.tf
```diff
@@ -0,0 +1,57 @@
...
+}
+
+variable "security_group_name" {
+  type = "string"
+  default = "allow_docker"
```
hackdna: I'd simply assign a value in the corresponding `module` section of `live/main.tf` based on `${terraform.workspace}` ([example](https://github.com/refinery-platform/refinery-platform/blob/ff1167230df226f27f932f846452cdbd8f901dd0/deployment/terraform/live/main.tf#L25))

deployment/terraform/modules/docker_host/main.tf
```diff
@@ -0,0 +1,57 @@
+variable "cidr_block" {
+  type = "string"
+  default = "10.0.0.0/16"
```
hackdna: It looks like both Orchestra and Odyssey are using 10.0.0.0/8, so I'd change this to either 172.16.0.0/16 or 192.168.0.0/16 to prevent any potential conflicts if there is ever a need for peering.
https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Subnets.html#VPC_Sizing

deployment/terraform/modules/docker_host/main.tf
mccalluc: Will do.

deployment/terraform/modules/docker_host/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "docker_hostname" {
+  value = "${aws_instance.docker_host.public_ip}"
```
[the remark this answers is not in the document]
mccalluc: Down the road, yes: It goes into the envvar that's another part of this PR. Not automatic right now, so not strictly necessary, but it will need to be added eventually, so leave it for now, unless you object?

deployment/aws-config/config.yaml.template
```diff
@@ -30,6 +30,9 @@ S3_BUCKET_NAME_BASE: "refinery"
 # Get from Terraform output
 COGNITO_IDENTITY_POOL_ID: ""
 
+# Get Docker IP from terraform
+DOCKER_HOST_IP: ""
```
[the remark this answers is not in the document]
mccalluc: Bad naming on my part: What we want at the end of the line is a string like `tcp://1.2.3.4:2376` as the value of DOCKER_HOST... which is an unfortunate name (DOCKER_URL would be more accurate) but that's the convention. Should be more clear with next commit.

deployment/terraform/modules/docker_host/outputs.tf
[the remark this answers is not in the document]
hackdna: Right, the output is useful. I was referring to the `public_ip` part. Resources inside a VPC should not be Internet accessible in general unless it is something like a web server.
https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Scenario2.html#Configuration-2

deployment/terraform/modules/docker_host/main.tf
mccalluc: Among other things, visualizations will need to pull data from S3, or from another source. Could also imagine the server side wanting to hit 3rd party APIs

deployment/terraform/modules/docker_host/main.tf
mccalluc: but both of those examples still have a corresponding definition in vars: for example, cognito/vars.tf:
```
variable "iam_resource_name_prefix" {
  description = "Name prefix of the S3 upload role"
}
```

I think I have what I want: This particular variable is called out as something that might be changed, but the default is fine.

mccalluc:
@hackdna : I think this addresses everything?

hackdna:
Attribute assignments should be aligned (easy with `Cmd-Option-L` in PyCharm if you have the Terraform plugin installed).
Also, if the docker host only needs to be accessible from within the VPC then it should be placed into a private subnet (no public IP).
Finally, have you tested with `terraform apply`?

deployment/terraform/modules/docker_host/vars.tf
```diff
@@ -0,0 +1,13 @@
+variable "cidr_block" {}
+
+variable "security_group_id" {
```
hackdna: Unused?

deployment/terraform/live/main.tf
```diff
@@ -24,3 +24,14 @@ module "identity_pool" {
...
+
+module "docker_host" {
+  source = "../modules/docker_host"
+  cidr_block = "${var.cidr_block}"
+  vpc_id = "${module.vpc.vpc_id}"
+}
+
+module "vpc" {
+  source = "../modules/vpc"
+  cidr_block = "${var.cidr_block}"
```
hackdna: add something like: `security_group_name = "${terraform.workspace}-docker"` to avoid multiple SGs in the account with the same (default) name

mccalluc:
@hackdna or @scottx611x : I've made some fixes, but hit this error:
```
 $ terraform state rm module.s3
 $ terraform state rm module.vpc
 $ terraform state rm module.docker_host
 $ terraform state rm module.cognito
 $ terraform apply -var 'identity_pool_name=mccallucTest' -var 'region=us-east-1'
* aws_instance.docker_host: Error launching source instance: InvalidParameter: Security group sg-7c15130b and subnet subnet-367c0153 belong to different networks.
	status code: 400, request id: 2b311ee8-187f-4566-8e09-60eb0decc45e
```

I'm not explicitly creating the subnet, so I'm not sure how to get a handle on that.

Also: I thought the workspace prefix would give me a unique name, but I was running into name collisions when creating the bucket: I've put a kludge in for now, but not sure of the right way of handling that.

hackdna:
First, `state` command is for advanced state management only and [should not be necessary](https://github.com/refinery-platform/refinery-platform/wiki/Terraform)

Second, you need to create a subnet: https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Subnets.html

Third, `$workspace` will give you a unique name based on the name of current workspace (it should not be the 'default' workspace).

mccalluc:
@hackdna / @scottx611x : I think this now does what we need. This much works:
```
ssh ubuntu@$REFINERY_HOST -i ~/.ssh/$KEY.pem
docker info | grep Name  # Remote machine
docker ps                # All docker commands seem to work
```

I will remove port 22 from the docker host, since I only needed that for debugging, and I'll sit down with Scott and see if he has other comments.

deployment/terraform/live/main.tf
```diff
@@ -24,3 +23,35 @@ module "identity_pool" {
   upload_bucket_name       = "${module.object_storage.upload_bucket_name}"
   iam_resource_name_prefix = "${terraform.workspace}"
 }
+
+module "docker_host" {
+  source              = "../modules/docker_host"
+  name                = "${var.name}"
```
hackdna: Replace with "${terraform.workspace}"

deployment/terraform/live/main.tf
```diff
@@ -24,3 +23,35 @@ module "identity_pool" {
...
+  vpc_cidr_block      = "${var.vpc_cidr_block}"
+  private_subnet_id   = "${module.vpc.private_subnet_id}"
+  vpc_id              = "${module.vpc.vpc_id}"
+  security_group_name = "${terraform.workspace}-docker"
+  key_name            = "${var.key_name}"
+}
+
+module "refinery_host" {
+  source              = "../modules/refinery_host"
+  name                = "${var.name}"
```
hackdna: Replace with "${terraform.workspace}"

deployment/terraform/live/main.tf
```diff
@@ -24,3 +23,35 @@ module "identity_pool" {
...
+  vpc_id              = "${module.vpc.vpc_id}"
+  security_group_name = "${terraform.workspace}-refinery"
+  refinery_host_count = "${var.refinery_host_count}"
+  key_name            = "${var.key_name}"
+  docker_hostname     = "${module.docker_host.docker_hostname}"
+}
+
+module "vpc" {
+  source             = "../modules/vpc"
+  name               = "${var.name}"
```
hackdna: Replace with "${terraform.workspace}"

deployment/terraform/live/vars.tf
```diff
@@ -1,7 +1,50 @@
...
+  type    = "string"
+  default = "10.0.0.128/25"
+}
+
+variable "refinery_host_count" {
+  type    = "string"
+  default = "0"
+}
+
+variable "key_name" {
```
hackdna: Add to terraform.tfvars.template

deployment/terraform/live/vars.tf
```diff
@@ -1,7 +1,50 @@
...
+
+variable "availability_zone" {
+  default = "us-east-1a"
 }
 
 variable "identity_pool_name" {
   description = "Cognito federated identity pool name"
 }
+
+variable "name" {
```
hackdna: Not needed, should use "${terraform.workspace}" instead.

refinery/config/settings/base.py
```diff
@@ -220,7 +220,6 @@ def get_setting(name, settings=local_settings, default=None):
     'rest_framework',
     'rest_framework_swagger',
     'django_docker_engine',
-    'httpproxy',
```
scottx611x: Double-check this... Is the idea that `django-docker-engine` specifies this internally somewhere now?

scottx611x: @mccalluc I believe we added this initially to suppress `httpproxy` logs polluting refinery logs

for example:
```
==> /var/log/apache2/refinery_error.log <==
2018-04-12 15:32:37 INFO     httpproxy.views:160 create_request() - GET http://10.0.0.95:32770/api/v1/tileset_info/?s=eQBZYJygQQ66ELEnt1jA7w&d=a5AQbI84Tw-yBSTQLsnOHg
2018-04-12 15:32:37 DEBUG    httpproxy.views:170 get_response() - Response body:
{"a5AQbI84Tw-yBSTQLsnOHg": {"min_pos": [0, 0], "max_pos": [2654911517, 2654911517], "max_zoom": 7, "max_width": 3276800000, "bins_per_dimension": 256, "transforms": [{"name": "ICE", "value": "weight"}], "name": "Dixon2012-J1-NcoI-R1-filtered.100kb.multires.cool", "coordSystem": "", "coordSystem2": ""}}
2018-04-12 15:32:38 INFO     httpproxy.views:160 create_request() - GET http://10.0.0.95:32770/api/v1/tiles/?s=eQBZYJygQQ66ELEnt1jA7w&d=a5AQbI84Tw-yBSTQLsnOHg.3.4.6&d=a5AQbI84Tw-yBSTQLsnOHg.3.5.6&d=a5AQbI84Tw-yBSTQLsnOHg.3.2.7&d=a5AQbI84Tw-yBSTQLsnOHg.3.3.7&d=a5AQbI84Tw-yBSTQLsnOHg.3.4.7&d=a5AQbI84Tw-yBSTQLsnOHg.3.5.7
2018-04-12 15:32:38 INFO     httpproxy.views:160 create_request() - GET http://10.0.0.95:32770/api/v1/tiles/?s=eQBZYJygQQ66ELEnt1jA7w&d=a5AQbI84Tw-yBSTQLsnOHg.3.0.2&d=a5AQbI84Tw-yBSTQLsnOHg.3.0.3&d=a5AQbI84Tw-yBSTQLsnOHg.3.0.4&d=a5AQbI84Tw-yBSTQLsnOHg.3.0.5&d=a5AQbI84Tw-yBSTQLsnOHg.3.1.2&d=a5AQbI84Tw-yBSTQLsnOHg.3.1.3&d=a5AQbI84Tw-yBSTQLsnOHg.3.1.4&d=a5AQbI84Tw-yBSTQLsnOHg.3.1.5&d=a5AQbI84Tw-yBSTQLsnOHg.3.2.2&d=a5AQbI84Tw-yBSTQLsnOHg.3.2.3&d=a5AQbI84Tw-yBSTQLsnOHg.3.2.4&d=a5AQbI84Tw-yBSTQLsnOHg.3.2.5&d=a5AQbI84Tw-yBSTQLsnOHg.3.3.3&d=a5AQbI84Tw-yBSTQLsnOHg.3.3.4&d=a5AQbI84Tw-yBSTQLsnOHg.3.3.5&d=a5AQbI84Tw-yBSTQLsnOHg.3.4.4&d=a5AQbI84Tw-yBSTQLsnOHg.3.4.5&d=a5AQbI84Tw-yBSTQLsnOHg.3.5.5&d=a5AQbI84Tw-yBSTQLsnOHg.3.2.6&d=a5AQbI84Tw-yBSTQLsnOHg.3.3.6
2018-04-12 15:32:38 DEBUG    httpproxy.views:170 get_response() - Response body:
{"a5AQbI84Tw-yBSTQLsnOHg.3.4.6": {"min_value": 0.0, "max_value": 0.03544166684150696, "dense": "gA5WDPsO2Qw0EA0QVwy2DPwLQwwCDt4UTwxWDOMQiw4zDgURug7vDg0QNg6HC/cIWwg4FOoKiA66DdsOAQ7KDFoOlQ1zDtgO+Ap4DB4Ougy0D5UOugv0DfoNGg19CuoMSA0nDkYN2A7hDEkQ+A3vCvoLEAs2D94O0w9zDLkNTxBzCZ8JwAkcDGYQbQ/yDcgOgA/VDFYNqQ5YDsIMABAyDW0OFgvNDLoPFhDwD8sQAADzE0MMrQ+sDjMNdQykDWYMvA3JDWQNHAxPDAkOQAvgCAAM5wj5DOENCQ2oDG4MZw6tDroN6QxMDf8M9ArEC44LKw3GDCIMewyyDOQMMgvDDAAAcQk/Db8KcQ2iCtUMgQ4uCmUPbApYDd8N5wylDssMBg0sDbAMxA0oClsMbg8ODtoMPwxADpAO2w69CVcMdA4GDvwNBQ4AAJYWtA61FlUKzQ/HDbIOkg4xDZINTworCwsMhhAVDVALTA+iDEwMSg3kD74N2goWEJoP6xLNDD8MRwzxDAQNoA5wDTYNEQ1eDxUL/QrZDE8O1QuzDIMOxwy5DncMWg1xDawKwwnpDn8ObwvhBTMN1A1QDc4NHw41EJQMAg1DDgoRow2ADtENoAxEEBENjw91DhUOrg3xDDgPeA9/EugLdhCTD/oNpw5aDGoNfwu0Dc8NQw8tDPEMeQrjEC4MXxD2DSERnBC4DWAKIg08C/YM4haSFBcMAw5tEOAO3xCjDm4PAxAIC1QNRRFRDAAAhg69DIANFQ2UDi8OIA9xEPcNdQ4BDCMNIA8GD3YO/A7rEIINPwudDZcIng+GDRkM3Q3YELkM7w8kCegMYw1GDZAN/QykCokMtAwZDqAMdQtTC10QAA+2DMgNFBA8DxMOQA7cEC8Ppg3RDO8O5g3HDNYL1g6ZD0kPbRAAAAAAbAxuDb4OBA3qDcAOqA1+DhcQMw19DZIIfgwuC+YOZxF4C3kN5Q37B+cLLgzQDJYMKA10ChYMwgn9CbELagw1CtMIrworCR8LOw4MC94MqAjGDV0KRA4ODboHBQxUERsPPgtFBwsNMQ0qDL4LEg4XDq8M3Q2TC1sMBAxtDdIP5wyUDRkObQ/uDmEKlQ3TD2IN+Qt0EwAAqRi+FAAAYAkRD1EO3A4vEMAQ9A7NDVYJ+g3LDogNoQwHEMoOBAv+DMENBA+cDkMO7hAAACAQwAnrCpUOmAygDQsK5gkcDlkNmA4LDoAMlgsPDYsNRA3CDgwP6gwPDEcMDw3TDecOUg5QCkcLNAzyDIUPnxAUEMoPyQzADnwNIg8uEK4RtAzzCS0RnAw0EDYOhw4XDaYNihBXDiEOMg0PD7QMZhBHEH8Pkg3VDQkQpg1oDewM/g79DcgN5gkIEKAO8hJiDycN6AwQC0EM/ww+EnMSCQv/CtAOhg4DEn4NjxAOEaYMwgMAADgJAAAtDQUPvQ6fDYMNKw+PDloQFhD6DZAQbw3tDZoMswxkC1YQMw8SDoYMoguODNgNEBDlD44MxA1rDjcOjwqaC+AJMg8ADc8PRAy9BxMMhgpQCrwM1w5pD4QOyg9QEKgP9wo8DXsPWAyqDloNHQ3DChMLQA2zEP0OVhBfEAAAAAAbDTAKSAyaDGwNMg0vD3AOeA0iDXcNpQnBDJ0Meg3mDDUKsAyADW4NLgz6DDAPkQw8Cv0MxAkYCuIJERAqDZEKFQyDC3kICAsgDXULIQvtDG4JNQxrDJUMFgxiDmIRDBCvDDYP7w4kDyQN2gnpDEwQ4g2OC8gKTQgaDdEOHw2bD/4MIQ6IDwgQqw9RDYUNHg/oD8IOAAAAAAAABRGiDosOSgY5DmAPBw1lDZwKVwv6D4cNcAoNDlQNcwm0DFIMKQ9MDIcOoA6xCS4NGwzODVIL9w4RB+8PZQquDJ8JYhCUDfkKFQyOC6EKoAsEDToQ2QzuCrkJ0whaDYsNsg5LD2sLZgw6DzwNOhCEDowLXQxdDUkQsBCMC24OlBALEGgQkBBrEQsMvA+nDV4LZw80DZcRAABYDhwNFg4pD24NhA4AD30MjQxvEQEQSws6DDgNzg5NDgYQ7w0QD4IPnA6kD7wPDhAhEMMY6xC+DngOMg07Df0OhQ5wDSsOBRCwDqwNpQwAAGMQYA47DvkMdQ07DVYPqw0nDCAMoAutDKsLjwtYDBIMHgwvDF0NvwxpDKIM8Q0qCwQM8wp/DIYNWQxPDI4MOg1+DIAO9Q3qDg8NNw45DrAOOg+FDJsMpw0+DCEMwAp/DJYOLg4/DV0PIQ+PDhUOTA6bDp4OCw5eDqsOAAAAAFsLnw29C2YNXQrHDeoPjQqQDfgNhQy3DBwMhQ1oDLUMKQx9DBsMJAwcDMcMTwuNC0gNEw0nDH4NuAzpDcEM6w59DmUNUg4ADjcNRA2nC0MOfQyRDCwMiAy5C8oLNg0PC38KZgxXDJYJuA6MDogKjgoKDSMLEwzgDusONA43D2ILngxuDKkNeQyGDPUMfAsSDbQOuhEAAAAAAAAAANIMdw3MDq4OQAwvDL0N9g7BD8EMWA3NDU0O9g0oDHgQJgylDfMLgQzsDE4NRAn2DN8LNA7fDCgQjA2eDRQN3AwvDocOrw/YDEgNtA0SDWINSA4eDSkNXgzBDIgKpA3iDVQM6AwnCloLUw0UCy4LYwxlC8IN/g3nDZsM7Q0TDPYKZw6RDSIMhgxLDmMOeg1FEOsO9g8AAE8L/wnADo4OIw4NDi8OtQmhCC8L+wrEDP0K9gtMDpcOCw2QD3ANwg/WDtkNXQ7zDqwOFhcHCxMPuA2oD4UPrhBSDjIPtg8+EB4NAACwDQAADQ8+Dy0P2w+WD00PUw8UEEINgwqeDBAPsgz6DdENtA1bDUwMGA5lDXsNKA0+DW4Mwg73DHINMA4GDFEMeQ1xDcYOsQzPDqYMUw6WD4UOBA1VDhkMtgyvDuEL6g47DzsMWhCbDZYMIg4ZEOcOiRCSD9UPfA9hEBMRdg4AAHwOowyADi4NPgwADQQO5AzpDHgMMg1DD+4NOA67DDcOJwzEDEINzA3ADY4Mkg1oDEoNhQ2UDV4NMA6FD7MNPA+fDWUOYQ4QEOINlg33DAsOXgnPDEgMKQ0GDJ0M9Aq0CgsNVg36DBIMyAvNDmsNUgwlDuQM6Q6NDrMP3g2xDBQQyAoyDTgQdgyYDYQN9A3EDj4PLA3pDAAAAAB3DXkNtg5UDOwPZw7SELkOag4sCsUOfg9TDlUPBg2xDxIOSA+1DHAODAzcDo0NKw0AAJgNxw58DUUMDg18Ci0Odg4LD40PsA4tDqkPdg9HDaIM2wxQDUQMqg1lDjsLeQy3Ck0O9g1yDH8M2QzBDIAKvAxKDUgNVA4BD5gN0As6D8wPww0lDaEOyw5jDuYN0Q2kDukP+g7gDAAA/Q3/DcgOKxGXDd0NoQvzC0UNJQw8CvILDA54DO8QjhAkEHAORQ/dD6ERSxF5EVgTIBO4GcAKyQ/TDj0O9Q6eDRMQhxCADe0QbBDdEEQQAABSEpwPFw7iDa4OuwwmDQMOeAzPDUoNcA9bDQAN6wwoCT0MNA3vEBsP/g2WDDcN8Ae5DCYLBw2mDH0NpRCJDsMOTg0lEPgP3BGOENEP+hGTEfUM1QzkCqAPZwoMDUYMMA3rDrAOiQ4cEasQ4xCDERMTBBPrDE4NHhCIDgAAOhI8DqAM+gxfDV4M8g1ECqMLfA0IDAINEA03Dz4PTAyyDVEN6w2bDA0NAA8TDtwMhw0RC48Pyw/1DtAPTA9QEEwQARBWECkQNBAiDvwM1w68C0MMQgxEC4gMuQ6SC3AKDQ51CIcMdwyZCCYGgwkkDS8KLw2jDkIPgBBEDxINmQ7jDC0KOQwdDDgIQw7NDtAMOQwSCrcMAAAAAAAAAACyDt4NNxD4DbgPBg3AEDANsBHlDF0ODxBVDo0OIQ8eEWcNHw9LDF4JRAjzCmsNTAxeDhIQPQ3bDEQPtxAAEGkM2A/3ELcQgxCSEBEQcwxKD8MMSwxiD6UQ5A97CL0HAguiC3AJVQpmCukLuQmMCOYLlA0ED9kLNQglDHgNZA2PDJYKAQ+KDBMNEgo+Dh4PxBAfD64IyQ4XDcINPA9fDsoOhg4+DGcMdwxlC/0JogyjCn0MLRDUDjkRAhEkEcEQlRAYD8AQjREuEZAb8Q5REVYOyw8oD6APjw+oENYPYRAUDwAANxAAAKUQLxCzEBQQDxCJDXgNfw/MDdMMTRDgD1kMwwwPDaQMXAjMDCEQyQ2aDUgNXQ0gDtsM0g3CDGEOBQ01D94Ovw9WDskMJQ+8EMMQNBB3D8wOBg7+DvoLXg0wC18MxQ01D7YPYA4pD24O9Q/mD40QGRGwEJ4O5BDBEDMQAABoDvQJtgmoCmEMfgykCnIKeA5qDIQOVQ01DaENYgwtDBoMrwoMDRsLJA+BDzUNKA2IDd8Nfg9VELYOkw5RDeIOMg/UD4UQOQ/kDrsQWg34DU4NTw86DB4Owg0hDs8MIwyrDFsMJAzkCrsIIw3uChUNfg0rDf0P9w9xD48OaA0AD2YNkw9tDhwN6A6bDnkPMA4cDRgP2ggAAAAAAAAAABUQZxAFEMQNWg4qDUgN1AxLD0EPihBnD+QNYBAzEP4PMw5lDUINBQ3CDaMOkw2GDSQPZw6fDQsM8Q66DggQtw/JDXUOYRCxD6APdRDXDT8N0w6dDGUOoA45EJIN+AmWDJUMMAsbC30MAg1IDEMNYgrICsYN9Q1KDUgNwgw9DVwJZwyMDosN4A0+DpoPyA9xEAgOOQs8EV8OBRBADo0OqQ6kDY4NPwsgCq4MZAubCUIKBQ46EHsPrBIMEeYRYRC2D/YPjA9+EMoSrSDKD1oPfg/CELIQRxBEDqcQBxC0EM8OvA4fDaET0A+TD4gQWQ4vD+cP8wytDWYPrA22DY4N2wnXDmsN1gyKDN4LwQ7VDnINWQ4nDOQNFhA4DxsNjwo5DHQMlww1DroNNQ4TDxUPQhDLD2cOyQ/DDtEO9w1ADrYN+A8qDgQN0g4MEDsO6Q8VD24QPhDID4MPlhCKECQRihMAADAPZQ3xDpgMWA2LDUYOJQ5oEPQMpQ4pDd4MmAy8DTwN6w0XDH8NUQ3sCjsOrAxuEFgN9g2pDj0OhAzgDUQNuQ/MDsgOeQ+0DXYNwA1BD3UO8wkaEGgMMgydDjsMrgoGEK8LLA5jDigNpQwYD00N7w19DhsPDQ1IDE8O2Q4nD2APow2VDnsMtg8zDo8NiA3HD04OAQ8VEAAAAAAAAAAAAgwOD2kO2A1sD1kO7Q3ZECkPJA9YDuQOnQ65D2IN7A88DjEN2A56DmMP4A5BDiEPAhCGDRkPVg3LEHENJhAoEFgPZw7YDhwQNRB3DnAPDA1rEKcPOw1CD48OYw30CeINgg1GDlEMBwwKELQN3wvXDkUNCQ3mDnMQhQ8RDjkOIgxSDeEJQQziDLUPlQ+IDwwPZxINENUPpA6/DIQPWg8yDrMPWA/HDjAKmAzpCggNeAyhDegQtQ05D5kRsw86D90ObQ6yDYsPcRWJKDMVrA0aDH4NDBDZEOgNlA3TD4APDwoAADUQAACQBXkKaA9WECUNhgzRDb0GMQwhEMkNYhBYEJEMMw72DXYLogofEGMKtw5ZD28MXwx1DVgLlA2TDeQOaAyxDbwM6g9GDloRlg3nEHoN0wwXEfIOEQ5PEc0OYg27DWkPLw2ODwgOkQ7uDe8MYQ0yEFUPIg9gFOsRshFMFwAAAAAtEmoOqw9HDrYMVQtPEYoJaQwkDLcPMw0iCkANXA9ECesNARAZB8gLMQtZDVkMcgqKC8YKCQ1BDu8OsAzzDQ0Klg79DM8M3AzoDVgNYw7yC3kNPQsgD5IKpw0eEHAPhw7eCm0NKQluDLwIqQl8DVUQ/g6HDvYOdgxBEQ0QoAsiDmUQrwzPDW8Q7Q4xDmsLXBGnD0QPAAAAAAwUDRQdDokPHxB2DI0RKQ0fD/MQdw3hDHUMqQ34Dk4P/g0vD/gOnwxpDIMQkwoAAAAA9wx1DGEPuQ/9DP8PsAwMD0ALBQ1NDhgOwQw8D2YLrQnXDcQR0Q/9DYoNWg3BCfML3w7iDfwJNQ58DCgR0AxhEPUKsQxTDJsIChEWDokQYg8EDRcO4w1fDDcNXQwNDHwN+BEqEVAWOxWAD3AMdhCBD9AOHQxZDW4JVwuBA4kJCg3SBtEPAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAiGQAAPRwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAP8WAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAthwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAGMYAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABYHWoYAAAAAAAAshoAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABuGwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAJMYAAAAAAAAAAAAAAAAAAADGAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABYaAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHocAAAAAAAAAAAAAD4RAAAlFXARAAAAAAAAAAAAAAAAAAAAAAAAqhAAAAAAAAAAAAAAAAAAAAAAAAAAAJ8VAAAAAFIQRxE2EAARAACmEQASAAAAAAAAAACkFAAAAAAAAAAA6BGCEQAAAAAAAOwRAADcFOgUAAAAAAAAAAAAAAAAoxEAALoSAAAAAAAAAAAAAAAA5REAAAAAAAAAAGYUAAAAAAAArRQAAAAAAAB7EAAAwxAAAAAAiREAAAAAAAAAAAAAAAAAAAAAAAAoFWYT2hMAAAAAAAAAAAAA3RMAAN0QAAAAADIRAADDECkRAAC3FgAAAADGEMwPAhUAAAAAAAAAAAAAVBAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAFMWAAAAAAAAAAAAAAAASxXiEC8SAAD1FQAAAACkEQAASREAAAAAAAAAAAAAAAAAAAAAAAAAAAcUxhIAAAAA4BM6FAAAAAAAAMQQAAD3EAAAAAAAAAAAKxIAAAAAZBQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA0Q8AAAAAAAAAAJoRbBIAAAAAAAAAAAAAAAAAAGUSAAAAAA
2018-04-12 15:32:43 DEBUG    httpproxy.views:170 get_response() - Response body:
{"a5AQbI84Tw-yBSTQLsnOHg.3.0.2": {"min_value": 0.0, "max_value": 0.010028749704360962, "dense": "AADUCGsJjwmAC2ELEglGDJ4KNwmdCmcHMgssCVwHlQiYClYHEAouDFEIRgwIDkgMWAuCCz0KyQgOCqkJygVNC6sGHAjgC2gM2AtUDEYIiQx+DI8J1wihDIcLlgq2C3oLcwpSCzIMcQ5wCxcKpgvRDLUNng19DJkKeQs9C7EMMgkiCeYGTAxlCNYISwVzCWkLtwnKCFcIjgn/CcwKTw28CncNlQzvDPYLrQ03DEUMtQk+DIIG+gn7CA0KvgocCtYK2g0rEHoMBw2RDA8MtQ23Di4MMg3tDaQOOA4sDTYM3A1tDl4OSRBwDrMOEQ9/D2YQ3w67D+oP4Q37DJgMcgsgCyMNFAlCCpUMXgxfC9EKvAleDfoKIgtTCkUNQA0tDF4NuAvECUgL3wj/CIUNQQlIDDkMugumCwULmAhyCjkM3gmiCg4MCQh5CoUKIAwxDA8MpgyiDZEM/AmbDOQJZg24C0EK7QpTCfAJTwxwCikM8gmsCAQIRQosCi0JqQx3DI0MSAwEDCcH0gtaCQkJZww8DPEImwylCe8NaQpUDNoNVQx7DBEM+w0BDBINkg9kDT8NXw3SDNgM2A0RDpYMVxACD88OuAw7D1sOpQ0yEWsPMxDWDwAAYxFVEBwPAAAzChEMpQnNCTYMZgjhB80JLgzJCnMKrApOCVoMXAbSCeAJmAwAAAoMiAwEDHsIoAhNCVwJqQy5CL0EwAmpCVMIeAmpCwQHFwyXCuIJfwpXDJ8MpwwwDYoKFgpuBxAFrgVDB6IF1Am5CZUKrQtHCN8MDA42DYMLFArYCaMK3gruC48LsAyKDDoMyAyFCsMKowtlCikNjgzxDfAMTA1ACowJRgZPCT8JLgpLCRELcwhBDfkH9wdYCHMI1ggQDJwKrwnYCJsIFwwTC2UJ7wwvECIOlQ4iCoMLiAqLCPEKnwlDBokJoQyiDDQMfAsnDfANygqGC+kNlAxWDUENmgt3DhgMvg1oDUQOyBD2EGgQfg7ADvYNXA8jELoOpw02DAAACQarCjYJlQn2B4oK7AkPCG4MgAjSC/wJZAtgCi8NYw0VDRENAw10DY4M6wqUChsNxgtVCSkLvAngCckJQQzwCWQJnwp0CWULMAofC4ULdAkMDMAKKQ3lDe4LeA2ICQgLlAc/DNcJOwU2CRcJSgqyDMEJ0wv2BpQJXwV9CxsMHAr/CQAL4gr1DKQJPAk+Ch8OlQv/DBEKsgz4C/MOcg2zDlENOA06DHkKZgp8DFINgg1kDGIOqwx+DK0NNA1LEK0Pvg5/D/4NCw6UD0cPSA89ELcQpxDDEOoQkxCTEU0R8g0oCTYMGgkxCQsMCQohBrAHUAz+BHUMEwlwB2UL2gb8CGsLOQtvDAAAmAi7CzYLEwwFBwgKgQqeC18JvQoIBXEGZQzVCf0HtAVUAnYDmggbBKUMsQqDCg8JjArqBzoIwwmKCyEIzwihClcICwzwCWEMOg05DAsMzAmyCQYBCQncCcQKrgzWClwM9wq3CCAKnQhYDHIMTA0SDDIPSA3GDZoKTwrMCfEIEgywCLsIygmFCY0HiQgHBHAJEwrQCAUN7QirCIIFqAgdC7UKOAv3Dx4QdBBUDacMRgxuC6cIUgZ6CNYL2AiTCUYNpgxCDD4NuQsLDB0Mtw2UDqkMLAvSDbcOJAwBDDANTQ+BEVQRPQ8zDGYOhQy5DjsRAQ4FDTAPAAAtC84MoAfdCT8LuAMnCRQJ0gjNCQcK5QyjCkMNfw6SDgwOJg46DU4MngtaCdcHOQiiCrYJVgoGBlgIPAoCCl8MzguJC8ELNQwKC4sJRwyeCnsKBg23DUcNgwxbDTELGAosCXAJPwkaBZ0ItAwGCrcNZAvvCTkKFAmVCNEKeQszDMwJgwwfDEUG8gfOCmwK9Qz8DyIOSQtYDf0MAw+bDv0MJQ3KCycNXQs8DRUOYQ1FDnANaQ1ZDKUMzA7DD6oP9RBdEDoNhQ6NDbsO1xCJEO4PchASEawQIA01FYwQuRG9EQAA5gkoCUsEXAsBDKsLfwfnCD8K/glVCDsIKwrYBeoIEQpGCDYNAAA3DVAL6QhcCqYMwgtiDGQKsQwaBSQJkgj6B70JHwx5CkoMPAktCiEJxwyzCdoJ7QsUCnwLPww7DVYLsAQnDdoLZAvICfIJQgroC50MggmLCRQNWwxJDHcOZgsGDR8NVAt1DS0KfQlVCwAA9w2NDBAMrwp3DGgLnAiJDRMKrAwoCuoKIA2QDC0MJQueCtsK2gsFDD8G5QqnCxQMeQ1MDPgMkQzqDCYMrQwLDHcM8Q1QDFMKCgthCVsMeAqsDDgM/QxcDkMObQ01DqINOw0aENgOFw5eDzoOQw3gDfAOQRCHEPANDA4OEQQRcxDvEPIPaw/LEc4Q/RBhBRMMBQixC7cMMgsDDKYLVgu4DQkJZgxQCQgL9wumC3ANVAyPC7gMBgzrCt0LeQrLCGQMdQjhDPoL9wvmCfUL3QkQCfsL7gh1DZEMUw0UDAoLOg1cDEINXwz2DEINFQwwDSMNgw6LCxoMPAnACdcKxQqDCQ0N0AztC1YMGAuLDBgNAA6FDPQL0wspDBoMwg1CDM0JzwuSDE4O1Q0YCpAOMQ1hDF4PCw5hDmwNVxAWDR0PgQ/zDggOEBLJDd8KHg0zDBoNOxF1ECAQsQw6C3oMyQ6mEMEQJg5YDgAApxBDEL8OAADbCZYK7AuoBzIJign2CJEJcgpHCmALzQgXCTALlAn8CVgOfQoAAIsJOAr2CiUK8AybDNgHywj6C/YJPwjTDTULoQnRCNAINQj0ChELcwn0CckLFwymCHYKLQx+CaQLJQtmDXEGpgcwDOEMsA1vC6QLtwyNDCcJXAsrCk0MUQw7DLoMHwwCCzMPAAvHC64K/AbTCgINnQwUDL4M9wrmDOsKkwyfCkEMhghDC14LxAtQCUAK2wgsCBsGkAwCC9ELMw1DDbkLnA3gCocOWg2gDHIKWQvHDnENdwsNDLcJPgwoCW0KqwzUDW4O2A41DfwOig9EELQPjBDFDAEPag2hDJ8NkxCjEBgQpQ09Dl0QGxDzDp0QIhG6D2wPGxG+DkkM5wtoCpkMtwx2Ce8MTQzLC2YLaQsnDF0M3wlWB84JKgxIDGoLBAoGC2kM+gsoCaQLVgsHBuoM3wtzCjkNcQqODA0K2Qx9CIgMcwxfDL8M/QjRC/MOeAz+DEAMVQsoDCIMuA1JC8AKYg32DdsL/wgYDV4KXg0YDJgMcwocDNQM8AkkDQsL+QyJDGwMGgweDn4NxwnwC3oK/Q/VDfANpQ1aDoQOJQ3NDQMPlg5WD3YQEhDpDgkQRRBTER8N7wu2DOQNQw07EKERbBGyDSYMcAynD8QQaBElEcYQAABWEn4N8gwAAHII/wrFCy0J1QzjCcgKzgzZCskLGApWDA8MDQvJDFwKwgiqDAAAswhlDGUKQQYuB+8MEgtwClgKXgw1Bx4I6QfGBDMJqwjzBRQHkQl2CQIMKAz/DKcK3wRmBagJEgdoCO4IpgdPCVQJIwzKCvgKqw3/C7kLhwzTCTwJAgZUCr4Mww91De8Kcgp0CJ4I7QpyDPYM4g7MDHcNhQ09DLcJ3gVNB70G/whRCGYE5Qc8CHAKSgsmCRQHqQg5CPINXgqdBhsKEAlwCqEMBAsCEOIRNxMAEV0IzQpVClgKdwaVCBoJ8AY6DQwQmA1bDPEOmQm0DAoM6QtPDlUMSgpsDAIQPQy+DEsKFA7uEBcUyQ98DT0NOg56DSUQmwwPC+cI2QwRCkUMMwn/B30KFwp9BrkILAtJCisKqQlXCoENIw+zDUMPDBHcENQNhQwWCzEMkAeoCQQIewafCBoM9gTOCeEI2AokDaoKsA3VCTwFswvFClQL8w6jDhwN7AqzC/gLdAmeCvIIgws4CbgIcQjHDYAPigwPBV0KfgawCPkKxwwJDUgK4wgCDGoGAQkXBpIKlAquDyEQUQxvC0IJCxBdEIkQ7A10CgoMJwlzCwUKzAtGCIQLaQ5FCUcIIA/KEYkQ0BLwEOkNqwo5DkATfhOIE94RxxHKEYoRPg0AAOYNrhNlEgAAigeeCCQLZBBLDEcIwAmHB6UIjAdiC8oHBAh8CEEIvwoOBnkMAAAAAOAIcwrWCgkKOQr4CxoMrQdrC8AK0Qm3CJ4Jvwa1CAEK8wXpBqkIZQ2GD9wOuQwlDHIJtgpaCY8FrwWdCvMJCAzLDLAMhw4HDwAN1w0YDCMLwAgPBmAJdw4cEBQNqgopCj4FqArwC/oJ1w2AErYQchCCEUEPMQiNA4YFGQm4BtYEcAilCCwKMgWwBjwGGgm/Ca4F3A05C9IIsAkbCYUIJwz2CfwQGBVeFHQQMwnXB2MGdwdvCewHaQj/CKwLOxBcDdsNTAymCBIOmw36CxUOXwxoDKIMQg9pChcKjglhDHgRxhQeEGIN7QuJC2QMEg6TDYAJTANWCwcMsww/CZALOwqtC4AInweGCGcMHAsHBwYNHw9nEQ8REhR2FOgT1w8wDQkNWwvoCEcKbAibCbMJfAgnC9sJJwoHCKoIXAuVDTEJ1wjEB2cKSAkeFDcSRRAYDcgMHQuBCTsKXQhqCSQGcAm5CsYLxRE7DZ0JFgg+BScDkglfDVUMOQhiCEEIlgJcCWMMPAmzDAcRQBIlCWkI3QoBEeYQARAbDr4JBwozBzoISgvdC/YLwwnhCaUKywTJDZUUvxLEFXsSHA3lDPUO7BPiFQ4X4RR8EigRTxECEl8RORGCFCIWDQpTCfwLwge/DIMJCQmDCFMKRgkjC3MLPQmhCIYIogq3CRIMNA0AAAAA1wn9C54MTQlKC8EKggv/CBwKUgl5CW8L4AmVCuQKhAM6CbAJCAzMDscNLQ+7DL4K+QrqBDkFsg09CgkI1gs3C1YKwAwlDS0PMgmvDNcMXwtADIkC4gpbDGUQ9QxdCYMKrAckDAsKkA5/DRURCREJECEPvQyaCBEKCwj6COsIqgkyCVAKjwlHDNQHFQZ7BFoJRgvJDUQKmQfwCrgJBwsXC3QJiRAVEyYUmhCdC04IjAo/Bl0K3QNOCNoJfAsUD8AMSQwgCSIMjgl3CpYNNgpmDq0Jhw0KDtILpAulCBoM0hEWFMQQmAyYDE4K+gzqDqsN+Qw9CwAAiwsdCikMKw1jCn8IYwmrC58KrAn8DBILAwyZD7oQIhEmE4wTCRQUD+cOIwt2CvoHmAxQCMkMEQpKDDAIDQo0B5QLUAmvCHINkglxCsEL3wonCT4VBBHCDx0JzwxNDC4Jgwp6CdAJIwn6CF8IZw1pEAULfwEgCcEFxQkCCjwMcAurDQIM2gq8CHYGEQtXDEkLMBEVEQ4M0AkcDPwO/RAhEXENCwm/CmQMeAwmCGsFUAqMC00M3QzqDKwOEBJEE+sUZBIoDNINMQ1QE0EWUBVcElUR4hGGEe0Q3ROEETQV4BQAAJEJ3wh9CisM5gw2CIwH8g4lCZIN/wmrB+4JJwnaCmAK8gdZDgAAAADoC+YJJQcgBaMJewsZCm8JFgfvB6UF7QR/B7oJ/wMUBBAKCQgeCp0KOg9NDgMMCgsRCmoJiwi4CAgGJgmMDP0LgwkiDB8MBgxNClENwgtqCiwHaQi0CRAMGRDoC+wLrwpnCRYJHwlYDpsMBhJFEDQQPBB7DzsKiQoJB7wGLgqECGgKyghvCCsL8wdpCKEJUgXSCKwNRQsJCaIFmAh5CmgIvgwQEFgTIRTIERgJPgptCT8HJwgLCd4KqwgSDEoO6QyqDZgMfApaDA4K9AtJDbEN/Am0CkkPTwxHCmkLgQ2GEDYT+A9xDAcOuQxUDBoPhg0lC+YMJwgaCFINrgq2CBYKEwyECOAKJQr7CUoKOAnCC3wPbhETESoSTBPEElQQegu6C48KjQnAB0EGXQhmCK8IxQlwCpIJ1wpnCCoJIg13CjsJMgdZCJMKHg9xDyAPUA1QDE0KAgqJCEsJDAs3BVoK+gpDC68Q1A0zCXEKkAWyBvcLQAtQDKIMMAg6B2UIBAh5CKEJKgkKEdkQCAxeCm4NLxDTEHcRjg+ACmsJdAphCoAJgAkWDKoJPQqkCUYMsw6TEuoS2RMVEhAN+Q1CDkwThRRNFZcTEBJ5EcMSThGQEB4SsBV0EyQJwAeFCP8LbwyvCbgK7QjpCXkJdwptCKwK3QcfBU4IFQ2cDGsMAAA8Dw8IzAz0CiQKVQpnCrIMpA2pCd4KfglhC34IPgvKBEYIbAImBuYGkgzHCtYOjQw3DIoJHwkMBoYIyAm7Cg0K1go/DDwKHA39DMIMGA0WCmgK+wovDCAJZQ3pDAYNbA1QC3kKFAkbCxUO8QkRD/8O3g65DlkMzAxsCScJwwmyCesIOQsTChsMoAhxCVcGvwr4Cw4JBQwVDKgJRAq9DAYMwAwzC5QOQBBdEO4Muwt9DMwHbQmHCQsKfQigCy0OJg6aDvQODAwODPMKLw2WDBoQaQ/+DigPuA+GDmsODg5OD40QLxFDEXQPjQ4cDTIOdhAtEG8NCA35BDkK6Qw6Ct8MVwoUCf4IyQZxChcKXgyxDIEKCQ0wDnwPfA5lD/sPMBAUDvMJHgwZCs4FvwqaB1MK9wldDLYLYQxwCjgIgwoMDWQL7QgvCxQKcgoeDc8OsQ0uDY8M3QuZC8EM+go+CukI6wpRCbkKeg2MDLYJKAhVCTgLXgmaDd0MHgzNC8MIPQvQCCINnQykDN4NARA8DWIN6wxTDcgN9w1ZDgILxQ3iDLAKUwxaDuUMLA77DWgN1A7jDbwPkw/5EIYOgg82ESoQdBE3EYURghHREMkRZxEdEgAA7RJdE6AQAAC9BxcKggwVCwEMHwnnCbkLxQhCC5QHOAoQC6AJKQknDAENAA4AANkJeAnmCDAJcgsqB7EJlwtQCrgJZwawC0MK8QVlB0ILcQSbCkMMwAlADHoNQQvoDIsMdQkJC94J2wZCCcIIgQkUCpQIyAn0DDwMxwjdCtQJJAyDCuAJ+QznDUcORAocDDkMiwmRCkQKYAehDGEOcQ44D94PJwwsCdgFPwrkCDcKKQdhCLsJGQg2C/MMkAIICdUMIQo2DSAMBwkDC+4FUQt7DFoLJQ9UEqsRTQ7vDAsK4QvlBwgKswlxCSoGuwyYDSIP0w4fDE0M2g3eDccN+w7UDqcNWQ5gEJcK5w0jDWcN6BBIEzMRdA7JEPoNdw59D7AOZA0RDAgOJgsVCwILdwvrCqEGCQh3Cs8LFQrADOAMEgl4DDYQhA8EED0QAhB7DdEJzwxaCrkKoQn/CMMJqwfOCtoMCgyODGMKwQmACx8O1AqECvgKygnRC6QN4Q7nDmIN+QyhC8kKlgrsClcFKgKdDBENwQ3NDk0MjgsKCdkI1QlsCZYMPQzRC7sI/AwwDUcIfwqYDYALTQ7aDl8N5QgyDZgPjhC0DuEOjguPDu8IDgz+B9oMtQzdDFgOVA0HDVoPMRFOEB8SDRHHDYsPvg8qEj4SyBOuEmQRbREeEqIRAAAJEk8UCxMAAJEK7Ah4CRMNVwnGAWAMrwjICXQK+Ql2Cv8IMwa8CrgL3An+CQAAAAA+C5AMcQaMCccKfwyWCToJgAxsC6MHcghUCTYIaAkhBhQGSwntBuUMqQxKDTAMZAiECCsKpAPQBhQJwQbmB34JUQonCXIMVw7oB0AKZgvcBigCEwfcCsEM7g4LDcgJMgi8CRkJIgsbDqsLTRAGD40OzBAWDloJDAodCskJSgrnC3AJvgjCB04N1AZOCTUJeAdAB4QLGwsOCsYGTAgOCj0MOgu0D+oRbhOuEFoLgQmGClsM/whkCYILFQg+DFIQww1QDrYMdgttDfAKbQyoD4wOjwtPDrgQGAvGCawJcQ4BER8VshB7D7kMlQzGDE0PHg+aC7wG/gRACjILsgqiDF0KRgpEClMKAQqOC/UKfAh2B0YPGg+hD10Q/xDiDlcPEQv3CyIMAgnTBaQJxQhZCpQIegmhCbMKswhpBzsJ6AzyCL4KsAhcDHgKcw1QEMQOmwweDPIKOQuLCHUIRApgCdwIkwjDDPoQHg1MBywMXAttDLAJNA23DmQJJwqECkkJegm5CnwLxgplEjMREA6oCyQMUxDuECoQDQ9BCrUMiQqsB2AJBAxwCr4LDQyVDF0Ksw9FE
```

hackdna: [Log level config for httproxy](https://github.com/refinery-platform/refinery-platform/blob/99658a792f3a81971e2b6c2be8f406360df32660/refinery/config/settings/base.py#L286)

refinery/tool_manager/tasks.py
```diff
@@ -14,6 +14,6 @@ def django_docker_cleanup():
 
 @task()
 def start_container(visualization_tool):
-    visualization_tool.django_docker_client.run(
+    return visualization_tool.django_docker_client.run(
```
hackdna: It's best to avoid returning a result in a task: http://docs.celeryproject.org/en/3.1/userguide/tasks.html#ignore-results-you-don-t-want

scottx611x:
@mccalluc Pinging you for review on this since I can't assign you.

deployment/refinery-modules/refinery/manifests/docker.pp
```diff
@@ -1,5 +1,15 @@
 class refinery::docker {
-  class { '::docker':
-    docker_users => [$::app_user]
+  if $::deployment_platform == 'aws' {
+    # Add env var pointing to remote docker host
+    file_line { "docker_host":
+      path => "/home/${app_user}/.profile",
+      line => "export DOCKER_HOST=${docker_host}",
+    }
```
hackdna: Is setting DOCKER_HOST in ~/.profile the best practice for production environments?

scottx611x: The current implementation requires someone to run the `load_tools` mgmt. command to add new `ToolDefinitions` representative of `VisualizationTools` into the system. This pulls down docker images under the hood. I don't know of a better approach currently, but somehow this user needs the `DOCKER_HOST` env var set.

deployment/aws-config/config.yaml.erb
```diff
@@ -38,6 +38,7 @@ COGNITO_IDENTITY_POOL_ID: "<%= tf.identity_pool_id %>"
 VPC_ID: "<%= tf.vpc_id %>"
 PUBLIC_SUBNET_ID: "<%= tf.public_subnet_id %>"
 RDS_DB_SUBNET_GROUP_NAME: "<%= tf.rds_db_subnet_group_name %>"
+DOCKER_HOST: ""<%= tf.docker_hostname %>""
```
mccalluc: Is the double-double quote correct?

deployment/manifests/aws.pp
```diff
@@ -31,8 +31,8 @@
 # See code in refinery-modules/refinery/...
 include refinery
 include refinery::apache2
-include refinery::aws
 include refinery::docker
+include refinery::aws
```
mccalluc: No longer alphabetical? (I don't really care.)

requirements.txt
```diff
@@ -16,7 +16,7 @@ django-auth-ldap==1.2.6
 django-celery==3.1.17
 django-chunked-upload==1.1.0
 django-debug-toolbar==1.4
-django-docker-engine==0.0.42
+django-docker-engine==0.0.42.2
```
mccalluc: @scottx611x : Could you file a bug against django_docker_engine for the error you hit earlier, that kept us from using the latest?

deployment/stack.py
```diff
@@ -105,6 +105,8 @@ def make_template(config, config_yaml):
         "DEFAULT_FROM_EMAIL=", config['DEFAULT_FROM_EMAIL'], "\n",
         "SERVER_EMAIL=", config['SERVER_EMAIL'], "\n",
         "IAM_SMTP_USER=", functions.ref('RefinerySMTPUser'), "\n",
+        "export DOCKER_HOST=", config["DOCKER_HOST"], "\n",
+        "export FACTER_DOCKER_HOST=", config["DOCKER_HOST"], "\n",
```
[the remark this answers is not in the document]
scottx611x: From what I understood the first `export DOCKER_HOST` made the env var available to processes managed by `supervisord` (celery tasks specifically) without having to manually set it for every config in `supervisord.conf`. While the second `export FACTER_DOCKER_HOST` allowed me to reference the value like: `${docker_host}` in puppet manifests

deployment/refinery-modules/refinery/manifests/docker.pp
```diff
@@ -1,5 +1,15 @@
 class refinery::docker {
-  class { '::docker':
-    docker_users => [$::app_user]
+  if $::deployment_platform == 'aws' {
+    # Add env var pointing to remote docker host
+    file_line { "docker_host":
+      path => "/home/${app_user}/.profile",
```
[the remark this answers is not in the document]
scottx611x: What the the `::` do for us?

deployment/terraform/modules/docker_host/main.tf
```diff
@@ -0,0 +1,60 @@
...
+  }
+  tags {
+    Name = "${var.security_group_name}"
+  }
+}
+
+resource "aws_instance" "docker_host" {
+  ami                    = "ami-2757f631"
+  subnet_id              = "${var.private_subnet_id}"
+  instance_type          = "t2.medium"
```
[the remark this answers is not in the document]
scottx611x: The single cpu t2 instance types had poor performance while testing

deployment/terraform/modules/docker_host/main.tf
```diff
@@ -0,0 +1,60 @@
+resource "aws_security_group" "allow_docker" {
+  name        = "${var.security_group_name}"
+  description = "Allow connection to docker engine from within VPC"
+  vpc_id      = "${var.vpc_id}"
+
+  # Explicitly set rule for port 2376 since this is the conventional port used
+  # for interacting with docker over tcp
+  ingress {
```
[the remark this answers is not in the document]
scottx611x: I thought it was better to leave it out since port 2376 is special? Thats what we're interacting with the docker machine over.

deployment/terraform/modules/docker_host/main.tf
```diff
@@ -0,0 +1,60 @@
...
+#!/bin/bash
+set -o errexit
+set -o verbose
+sudo apt-get update
+sudo apt -yq install docker.io
+sudo mkdir /lib/systemd/system/docker.service.d
+sudo echo -e '[Service]\nExecStart=\nExecStart=/usr/bin/dockerd -H fd:// -H tcp://0.0.0.0:2376' > /lib/systemd/system/docker.service.d/startup_options.conf
+sudo systemctl daemon-reload
+sudo service docker restart
+EOF
```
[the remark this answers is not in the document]
scottx611x: Would that become a headache if we just want this one thing?

Its easy to determine if this machine has docker installed from the Refinery instance (use docker-py client, launch a tool, etc.), but it would be more difficult to determine whether a puppet manifest applied properly. We'd have to configure ssh access from the refinery host to the docker one to be able to tail logs etc. which I don't think we need to do.

Maybe less is more in this case?

deployment/refinery-modules/refinery/manifests/docker.pp
```diff
@@ -1,5 +1,19 @@
 class refinery::docker {
-  class { '::docker':
-    docker_users => [$::app_user]
+  $_docker_host = "unix:///var/run/docker.sock"
```
hackdna: The reason to set `DOCKER_HOST` on both AWS and in Vagrant is to enable communication using TCP in both cases. The only difference is that on AWS it's a separate box and in Vagrant it just happens to be localhost.

deployment/refinery-modules/refinery/manifests/docker.pp
```diff
@@ -1,5 +1,19 @@
...
+
+  if $::deployment_platform == 'aws' {
+    # Get docker host value from FACTER_DOCKER_HOST
+    $_docker_host = "${docker_host}"
```
hackdna: `${::docker_host}` then no need for an underscore in `$_docker_host`

deployment/stack.py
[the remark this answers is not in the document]
hackdna: Puppet adds `DOCKER_HOST` to both `~/.profile` and Django settings. Celery workers should be able to obtain that value from either of those places.

deployment/terraform/modules/docker_host/main.tf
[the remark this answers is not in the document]
hackdna: What were the testing conditions?

deployment/terraform/modules/docker_host/main.tf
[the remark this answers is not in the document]
hackdna: You can simply add that info to the comments of the other block.

deployment/terraform/modules/s3/outputs.tf
```diff
@@ -1,3 +1,7 @@
 output "upload_bucket_name" {
   value = "${aws_s3_bucket.uploaded_files.id}"
 }
+
+output "bucket_name_base" {
```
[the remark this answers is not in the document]
hackdna: This value is already output by the main module.

deployment/terraform/modules/docker_host/main.tf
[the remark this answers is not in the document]
hackdna: The problem is that there are already two different ways to provision the environment: Puppet on AWS/Vagrant and a custom script on Travis CI. Now Docker configuration is applied with Puppet on Vagrant and with a custom script on AWS. This leads to code duplication and can also introduce subtle differences in configuration that could make troubleshooting very difficult.

Also, what makes you think that a shell script is more reliable that applying a Puppet manifest? If Docker provisioning fails then you still have to examine the logs in either case.

deployment/terraform/modules/docker_host/main.tf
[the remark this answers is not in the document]
scottx611x: I wanted to at the very least hit the max running container limit of 10 while having at least two instances of each of the tools to be deployed in production running. While trying that with `t2.micro` I was unable to launch more than 3 concurrently. I then bumped up to `t2.medium` and had success. I could always try with `t2.small`

deployment/terraform/modules/docker_host/main.tf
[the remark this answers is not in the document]
scottx611x: I was just thinking that we're trying to do something very specific with this 2nd host and the time investment to configure and provision it with puppet may not be worth it at the moment if we have growing pains with this approach and need to move to a more scalable solution.

deployment/terraform/modules/docker_host/main.tf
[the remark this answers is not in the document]
hackdna: That would be useful. Over-provisioning becomes expensive very quickly: t2.small is twice the cost of t2.micro and t2.medium is twice the cost of t2.small. Scaling horizontally instead of vertically is key.

deployment/refinery-modules/refinery/manifests/init.pp
```diff
@@ -166,7 +166,10 @@
 ->
 exec { "supervisord":
   command     => "${virtualenv}/bin/supervisord",
-  environment => ["DJANGO_SETTINGS_MODULE=${django_settings_module}"],
+  environment => [
+    "DJANGO_SETTINGS_MODULE=${django_settings_module}",
+    "DOCKER_HOST=${docker_host}"
```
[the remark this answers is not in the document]
scottx611x: From what I understand, subprocesses managed by supervisor have accesses to the environment variables of the shell used to start supervisord as well as ones set through: `environment =>`. I don't believe the `user` entry gives access to anything within `/home/${::app_user}/.profile` unless you explicitly set it.

http://supervisord.org/subprocess.html#subprocess-environment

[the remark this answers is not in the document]
hackdna: OK. Then lets just reference the globals explicitly: `${::docker_host}` and `${::django_settings_module}`

Issue #2352, opened 2017-11-09
User/Files Filtering Behavior


jkmarx:
* Specific code commit: 09a2aa2e7e739fb286f370258e899dfe27f5d9c5
* Version of the web browser and OS: Mac OSX
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): 
 Test site

### Steps to reproduce
Please list all the actions and the input data used:

1. User/Files
2. Search, selected fields, etc

### Observed behavior
![user-attribute-bug](https://user-images.githubusercontent.com/9956764/32624737-23c1a820-c558-11e7-8dcc-3cbf61ab9bbe.gif)

### Expected behavior
Ref #2263

PR #2604, opened 2018-02-17
Jkmarx/hide user files column

### LINKED ISSUE #2350: User/files Empty Columns
* Specific code commit: 09a2aa2e7e739fb286f370258e899dfe27f5d9c5
* Version of the web browser and OS: Mac OSX
 Test site

### Steps to reproduce

1.  Open the user files page
2. Sort Date or Experimenter

### Observed behavior
column-empty

### Expected behavior
* If columns are empty, should they even be shown?

### LINKED ISSUE #2458: Add columns/facets to cross filebrowser to meet Shivdasani lab needs

jkmarx:
-This is a temp solution to not create columns for attributes which do not have any fields. There is another bug which is causing empty columns. Specifically, attributes which have fields but the fields all return 0. Wondering if it's a local issue? Ref #2352
Generally resolves #2350 
Resolves #2458 

- [x] Fix columns which don't have facet counts


refinery/ui/source/js/user-file-browser/services/factory.js
```diff
@@ -68,8 +68,13 @@
                   '</a>' +
                 '</div>',
             width: 30 }];
+      // temp solution to handle empty columns until backend is updated to
+      // avoid sending attributes with no fields
+      var validateAttributeStr = culledAttributes.join('');
       settings.djangoApp.userFilesColumns.forEach(function (column) {
-        defs.push({ field: column });
+        if (validateAttributeStr.indexOf(column) > -1) {
```
mccalluc: I'd worry about getting false positives from chance concatenations... Could you use the array instead?
```javascript
_.contains(culledAttributes, column)
```
(or a built-in, if the JS version we're targetting supports it.)

jkmarx: @mccalluc I need to make a few other changes also, but yes your suggestion makes sense.

jkmarx:
@mccalluc Sorry, I actually wasn't ready for this pull request to be reviewed. I added & then removed you as a reviewer.

PR #2608, opened 2018-02-21
Jkmarx/visual no display

### LINKED ISSUE #2552: Visualization Tab - loading bug
* Specific code commit: e859c8eb301a7aeb95222d36629e354d1bc4bb45
* Version of the web browser and OS: Safari
Beta site

### Steps to reproduce

1. Don't login
2. Navigate to a public data set which has visualizations, view the visualization tab.

### Observed behavior
Spinner keeps spinning

### Expected behavior
For it to time out with something informative or show visualizations.

jkmarx:
Resolves #2552


refinery/tool_manager/tests.py
```diff
@@ -2668,12 +2668,6 @@ def test_get_request_authenticated(self):
         self.create_tool(ToolDefinition.VISUALIZATION)
         self.assertIsNotNone(self.get_response)
 
-    def test_get_request_no_auth(self):
```
jkmarx: We check for obj level perms in the API therefore we do not need the api level check. Units tests were already written for the obj level perms, so these are unnecessary now.

PR #2630, opened 2018-02-27
Jkmarx/fix facet query user files

### LINKED ISSUE #2178: Facet counts behave differently between cross-dataset and within-dataset
* Specific code commit: 2a3c825183c423d388c391f52a18e13d5caa2925

### Steps to reproduce

On cross dataset, the counts seem to be the number records with that value in the current found set, while it within dataset, it's more like the number that would be added?

On cross dataset, selecting only a zero valued facet will bring the others to zero, while within dataset, there may be to results, but the other facets will retain their counts.

I'm not clear on what the desired behavior is. For context: #2088 and #2069

jkmarx:
- Resolves #2178 
- Syncs filter order w/ column order

user-files-fix


refinery/data_set_manager/search_indexes.py
```diff
@@ -112,50 +112,6 @@ def prepare(self, node):
 
         id_suffix = "_" + id_suffix + "_s"
 
-        # User files gets pre-defined attributes for now.
-        data['filename_Characteristics' + NodeIndex.GENERIC_SUFFIX] = node.name
```
jkmarx: removed filename field, not necessary since the name field is the same thing

refinery/data_set_manager/search_indexes.py
```diff
@@ -112,50 +112,6 @@ def prepare(self, node):
...
-
-        data['filetype_Characteristics' + NodeIndex.GENERIC_SUFFIX] = node.type
-
-        data.update(self._assay_data(node))
-
-        # create dynamic fields for each attribute
-        for annotation in annotations:
```
jkmarx: shifted the order, did not delete any code. We grab file_store_item first so we can set the user/files index before we create dynamic fields

refinery/data_set_manager/search_indexes.py
```diff
@@ -166,6 +122,10 @@ def prepare(self, node):
             file_store_item = None
             download_url = NOT_AVAILABLE
         else:
+            data['filetype_Characteristics' + NodeIndex.GENERIC_SUFFIX] = \
```
jkmarx: once fileStoreItem confirmed, we set filetype. It's the same filetype as in the file browser

refinery/data_set_manager/tests.py
```diff
@@ -2062,14 +2059,6 @@ def test_prepare_node_no_file_import_task_id_yet(self):
             expected_download_url=PENDING
         )
 
-    def test_prepare_node_no_file_store_item(self):
```
jkmarx: Need to follow up on this unit test which was failing. Waiting for the fileStore pull request to merge also.

refinery/data_set_manager/utils.py
```diff
@@ -668,15 +667,21 @@ def generate_solr_params(
...
-        facet_template = '&facet.field={0}_Characteristics{1}' + \
-                   '&facet.field={0}_Factor_Value{1}'
-        solr_params += ''.join(
-            [facet_template.format(s, NodeIndex.GENERIC_SUFFIX) for s
-             in settings.USER_FILES_FACETS.split(",")]
-        )
-        solr_params += '&fl=*{},name,*_uuid,type,django_id,{}'.format(
-            NodeIndex.GENERIC_SUFFIX, NodeIndex.DOWNLOAD_URL
-        )
+        facet_arr = []
```
jkmarx: to utilize the helper methods, I needed to create the facet list first and then create the query. This is the main part of the solution to sync up the file browser filtering with the user files filtering

refinery/ui/source/js/user-file-browser/ctrls/user-file-browser-files-ctrl.js
```diff
@@ -50,7 +50,7 @@
         }
       });
 
-      return _.keys(facetCountObj).concat('date_submitted', 'sample_name', 'filename');
+      return _.keys(facetCountObj).concat('date_submitted', 'sample_name', 'name');
```
jkmarx: use already indexed name

refinery/data_set_manager/utils.py
```diff
@@ -668,16 +667,15 @@ def generate_solr_params(
 
     if facets_from_config:
         # Twice as many facets as necessary, but easier than the alternative.
-        facet_template = '&facet.field={0}_Characteristics{1}' + \
-                   '&facet.field={0}_Factor_Value{1}'
-        solr_params += ''.join(
+        facet_template = '{0}_Characteristics{1},{0}_Factor_Value{1}'
```
jkmarx: modified to use helper methods

refinery/ui/source/js/user-file-browser/ctrls/user-file-browser-files-ctrl.js
```diff
@@ -95,13 +97,25 @@
...
+        + $location.host() + ':' + $location.port()
+        + vm.downloadCsvPath();
     };
     vm.gridOptions = gridOptionsService;
     vm.gridOptions.onRegisterApi = function (api) {
       api.core.on.sortChanged(null, vm.sortChanged);
     };
+
+    // helper method to create facet query for solr
+    function createFacetQuery () {
```
jkmarx: shifted logic in service to use filter_attribute field which meant I had to separately make the fq field here

refinery/ui/source/js/user-file-browser/partials/user-file-browser-filters.html
```diff
@@ -48,8 +48,8 @@ <h4 class="panel-title">
                         <input
                           type="checkbox"
                           name="optionsCheckboxes"
-                          ng-click="$ctrl.filterUpdate(attribute, field.name)"
-                          ng-checked="$ctrl.filterIsSet(attribute, field.name)"
+                          ng-click="$ctrl.filterUpdate(field.assocAttribute, field.name)"
```
jkmarx: associate internalName with field for filtering

refinery/ui/source/js/user-file-browser/services/factory.js
```diff
@@ -22,11 +22,7 @@
       createColumnDefs: createColumnDefs,
       createData: createData,
       createFilters: createFilters,
-      getUserFiles: getUserFiles,
-      _mergeAndAddObject: _mergeAndAddObject,
```
jkmarx: remove unused helper methods since I shifted the logic for the attributeFilter. Track the different attributes and keep duplicate fields separate.

refinery/ui/source/js/user-file-browser/services/factory.js
```diff
@@ -102,64 +98,32 @@
...
             return nameCount.name.toLowerCase();
           }).join(' ');
           // "foo_Characteristic" and "foo_Factor_Value" both map to "foo".
-          var display = mapInternalToDisplay(key);
+          var display = mapInternalToDisplay(attributeName);
           if (!angular.isDefined(filters[display])) {
             filters[display] = {
               facetObj: [],
-              lowerCaseNames: ''
+              lowerCaseNames: '',
```
mccalluc: Why the trailing comma?

refinery/ui/source/js/user-file-browser/services/factory.spec.js
```diff
@@ -94,9 +58,10 @@
           {
             organism: {
               facetObj: [
-                { name: 'mouse', count: 4 },
-                { name: 'cat', count: 2 },
-                { name: 'dog', count: 4 }
+                { name: 'mouse', count: 1, assocAttribute: 'organism_Characteristics_generic_s' },
+                { name: 'cat', count: 2, assocAttribute: 'organism_Characteristics_generic_s' },
+                { name: 'mouse', count: 3, assocAttribute: 'organism_Factor_Value_generic_s' },
+                { name: 'dog', count: 4, assocAttribute: 'organism_Factor_Value_generic_s' }
```
mccalluc: I'm trying to remember, but are the facet names changing here? Did the `generic` replace just the distinguishing IDs, or `Characteristics` vs `Factor_Value` as well?

mccalluc: Never mind: This isn't a change, I don't think.

jkmarx: Correct, it's not a change. It's tracking it's internal name vs display name.

refinery/ui/source/js/user-file-browser/services/factory.js
jkmarx: Typo

PR #2633, opened 2018-02-27
Scottx611x/remove workflow manager

### LINKED ISSUE #2463: Remove workflow_manager app
Look into the potential removal of this.
(the workflow visualization may be the only thing that needs to stick around)

See also #2583 & #2466

scottx611x:
Resolves #2463


refinery/core/tests.py
```diff
@@ -1593,9 +1589,23 @@ def setUp(self):
...
 
-        self.url_root = '/api/v2/data_sets/'
+        self.client.login(username=self.username, password=self.password)
+
+
+class DataSetApiV2Tests(APIV2TestCase):
+    def create_rand_str(self, count):
+        return ''.join(
+            random.choice(string.ascii_lowercase) for _ in xrange(count)
+        )
```
mccalluc: Probably better not to introduce non-deterministic behavior?

refinery/templates/core/workflow.html
```diff
@@ -74,12 +74,14 @@ <h4>Layout</h4>
 		<p>
 			<label class="radio">
                 <input type="radio" name="optionsRadios" id="cb_layout_kind_refinery" value="1"
-                       onclick="workflowVisualizationModule.reload_workflow('{% url 'download_workflow' workflow.uuid %}')"
+                       onclick="workflowVisualizationModule.reload_workflow(
+                         '{% url 'workflow-download-workflow' workflow.uuid %}')"
```
mccalluc: Sorry, but I don't see where this is coming from: The name seems odd, but maybe it's the best choice in context?

scottx611x: Ahh so Django Rest framework provides human readable names for each url within a Router.
In this case for a detail route in `WorkflowViewSet` its: `{basename}-{methodname}`

This link was helpful to me: http://www.django-rest-framework.org/api-guide/routers/#defaultrouter

`{% url 'workflow-download-workflow' workflow.uuid %}` ends up becoming:
`/api/v2/workflows/<workflow_uuid>/download_workflow`

Looking at this again, I think `/api/v2/workflows/<workflow_uuid>/download` would be better

refinery/core/tests.py
scottx611x: I think this just moved up/down in the diff. I didn't introduce it. I'll look into it though now that I'm around it.

Issue #2628, opened 2018-02-27
Django Tests - Refactor into seperate files


jkmarx:
Organizing Django Tests - 2 Scoops Guide [JM]
Break up the monolithic test.py into test_views, test_models, test_utils……

![screen shot 2018-02-21 at 7 22 02 am](https://user-images.githubusercontent.com/9956764/36738868-9405a06e-1bac-11e8-8a6c-cb37b08056ff.png)

Issue #2563, opened 2018-02-06
Refactor FileStoreItem.get_datafile_url()


hackdna:
[File URLs can be provided by Django](https://docs.djangoproject.com/en/1.7/topics/files/) if [MEDIA_URL](https://docs.djangoproject.com/en/1.7/ref/settings/#media-url
) is set correctly.

hackdna:
Closed in #2634

Issue #2515, opened 2018-01-19
FileStoreItem.get_file_extension() returns None for certain FileTypes


hackdna:
* Specific code commit: b9a3e9a
* Version of the web browser and OS: N/A
* Environment where the error occurred: any

### Steps to reproduce
Make sure initial data is loaded and open Django shell:
```bash
manage.py migrate
manage.py shell
```
Create a `FileStoreItem` with a `FileType` referenced by more than one extension, for example:
```python
>>> import django
>>> django.setup()
>>> from file_store.models import FileStoreItem
>>> f = FileStoreItem(source='/path/test.tdf')
>>> f.set_filetype('tdf')
True
>>> f.get_file_extension()
```

### Observed behavior
```
2018-01-19 16:38:14 ERROR    file_store.models:273 get_file_extension() - Error while trying to fetch FileExtension get() returned more than one FileExtension -- it returned 2!
```

### Expected behavior
`FileStoreItem.get_file_extension()` should return extension of the file on disk or from the source as stated in its docstring

hackdna:
Closed in #2634

PR #2634, opened 2018-02-27
Hackdna/refactor file store

hackdna:
Solve #2515 and #2563
Toward #2224, #2242, #2508, #2531 and #2628


refinery/core/tests.py
```diff
@@ -2295,3 +2296,49 @@ def test_create_workflow_engine_bad_group_name(self):
...
+        self.assertFalse(is_absolute_url('/path/to/file'))
+        self.assertFalse(is_absolute_url('example.org'))
+
+    def test_is_absolute_url_with_empty_input(self):
+        self.assertFalse(is_absolute_url(None))
+        self.assertFalse(is_absolute_url(''))
+
+    def test_get_absolute_url_with_absolute_url(self):
+        self.assertEqual(get_absolute_url('http://example.org'),
+                         'http://example.org')
```
mccalluc: Move this test into the next class?

refinery/core/tests.py
```diff
@@ -2295,3 +2296,49 @@ def test_create_workflow_engine_bad_group_name(self):
...
+
+    def test_get_absolute_url_with_absolute_url(self):
+        with override_settings(SITE_ID=self.current_site.id):
+            self.assertEqual(get_absolute_url('http://example.org'),
+                             'http://example.org')
+
+    def test_get_absolute_url_with_empty_input(self):
+        with override_settings(SITE_ID=self.current_site.id):
+            self.assertEqual(get_absolute_url(None), None)
+            self.assertEqual(get_absolute_url(''), '')
```
mccalluc: Is this the behavior you want? I worry that folks could be surprised when the result of `get_absolute_url` does not pass `is_absolute_url`. Maybe it should throw an exception?

refinery/core/utils.py
```diff
@@ -853,13 +849,14 @@ def get_full_url(relative_url):
...
-        settings.REFINERY_URL_SCHEME, current_site.domain, relative_url
+        settings.REFINERY_URL_SCHEME, current_site.domain, string
     )
 
 
-def is_url(string):
-    """Check if a given string is a URL"""
-    return urlparse(string).scheme != ""
+def is_absolute_url(string):
+    if string and '://' in string:
```
mccalluc: - I think you want a stricter regex: I don't think anything would preclude `://` from showing up in the path or query
- just return the result of the expression? `if` is redundant.

refinery/data_set_manager/models.py
```diff
@@ -595,17 +595,16 @@ def run_generate_auxiliary_node_task(self):
         file_store_item = self.get_file_store_item()
 
         # Check if we pass the logic to generate aux. Files/Nodes
-        if (file_store_item.filetype.used_for_visualization and
-            file_store_item.is_local() and
+        if (file_store_item and file_store_item.filetype and
```
jkmarx: very lengthy, helper method?

refinery/data_set_manager/tests.py
```diff
@@ -1540,9 +1540,9 @@ def test_update_attribute_order_ranks_invalid(self):
         attribute_list = AttributeOrder.objects.filter(assay=self.assay)
         self.assertItemsEqual(old_attribute_list, attribute_list)
 
-    @mock.patch("data_set_manager.utils.core.utils.get_full_url",
-                return_value="test_file_a.txt")
+    @mock.patch("data_set_manager.utils.core.utils.get_absolute_url")
```
jkmarx: Don't see a reason to change the previous format.

refinery/file_store/tests.py
```diff
@@ -254,39 +213,53 @@ def setUp(self):
...
+            identity_id=identity_id
+        )
+        self.assertEqual(
+            translate_file_source(self.rel_path_source),
+            "s3://{}/{}/{}".format(
+                settings.UPLOAD_BUCKET, identity_id, self.rel_path_source
+            )
+        )
 
-    def test_translate_from_relative_path_without_username_or_base_path(self):
```
jkmarx: prefer this name

refinery/file_store/models.py
```diff
@@ -280,98 +238,33 @@ def get_file_size(self, report_symlinks=False):
             return 0
 
     def get_file_extension(self):
-        '''Return extension of the file on disk or from the source.
```
jkmarx: these comments are clearer

refinery/core/tests.py
hackdna: Which one is the next class? TestGetAbsoluteURL class is the last one in that file.

refinery/core/tests.py
hackdna: Good question. I would have done this but since I was refactoring, I've kept the function signature as is. Raising an exception requires exception handling in every use case. I'll open an issue for this.

hackdna: See #2636

refinery/core/utils.py
hackdna: * Keep in mind that it is not trying to validate a URL. I figured if it's good enough for Django then it's good enough for Refinery: https://github.com/django/django/blob/5cc28dc752c3ae78456bb835c3ba195489fc26d7/django/conf/urls/static.py#L22.
* I'll refactor to simplify.

refinery/core/utils.py
```diff
@@ -834,16 +834,12 @@ def invalidate_cached_object(instance, is_test=False):
         return mc
 
 
-def get_full_url(relative_url):
+def get_absolute_url(string):
+    """Creates an absolute URL from a relative URL using the current Site
+    domain and REFINERY_URL_SCHEME Django setting
     """
-    Creates a full URL (including hostname) from a given relative URL
```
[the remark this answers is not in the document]
hackdna: This is the [Sphinx format](http://www.sphinx-doc.org/). We moved away from Sphinx a while ago because it was more trouble than it's worth.

refinery/data_set_manager/models.py
hackdna: I think the whole `run_generate_auxiliary_node_task()` might need refactoring or rewriting. This is just a quick fix to make it more robust.

refinery/data_set_manager/tests.py
hackdna: Format?

refinery/file_store/tests.py
hackdna: Changed the name to make it more explicit.

refinery/core/tests.py
mccalluc: This is in `TestIsAbsoluteURL` right now: `TestGetAbsoluteURL` begins just below.

refinery/file_store/models.py
hackdna: Perhaps. However, old comments don't match the new functionality.

PR #2638, opened 2018-02-28
Scottx611x/persistent visualization urls

scottx611x:
I noticed this morning that although we have the ability to relaunch `VisualizationTools` from the UI, if one was to share the url of a running `VisualizationTool` with someone else who accesses said url after said `VisualizationTool's` container had been purged then they would get the "Please Wait" page indefinitely.

This PR relaunches `VisualizationTools` if they aren't already running when their url is accessed
feb-28-2018 14-17-20

### DIFF
--- a/refinery/tool_manager/tests.py
+++ b/refinery/tool_manager/tests.py
@@ -61,8 +61,6 @@ from factory_boy.django_model_factories import (AnnotatedNodeFactory,
                                                 ToolFactory)
 from factory_boy.utils import create_dataset_with_necessary_models
 from file_store.models import FileStoreItem, FileType
-from selenium_testing.utils import (MAX_WAIT, SeleniumTestBaseGeneric,
-                                    wait_until_class_visible)
 from tool_manager.management.commands.load_tools import \
     Command as LoadToolsCommand
 from tool_manager.tasks import django_docker_cleanup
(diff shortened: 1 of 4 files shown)


refinery/tool_manager/tests.py
```diff
@@ -2979,6 +2979,38 @@ def test_get_with_invalid_tool_type_request_param(self):
...
 
+    def _test_launch_vis_container(self, user_has_permission=True):
+        self.create_tool(ToolDefinition.VISUALIZATION)
+        self.assertFalse(self.tool.is_running())
+
+        if user_has_permission:
+            assign_perm('core.read_dataset', self.user, self.tool.dataset)
+
+        # Need to set_password() to be able to login. Otherwise
+        # usr.password in the hash representation which is not what the
```
mccalluc: "in the hash" -> "is the hash"?

refinery/tool_manager/tests.py
```diff
@@ -2979,6 +2979,38 @@ def test_get_with_invalid_tool_type_request_param(self):
...
+        # login() expects
+        temp_password = "password"
+        self.user.set_password(temp_password)
+        self.user.save()
+        self.client.login(username=self.user.username,
+                          password=temp_password)
+
+        with mock.patch.object(VisualizationTool, "launch") as launch_mock:
+            get_response = self.client.get(
+                "{}/".format(self.tool.get_relative_container_url())
```
mccalluc: Is the trailing slash required?

refinery/tool_manager/views.py
```diff
@@ -164,3 +168,59 @@ def relaunch(self, request, *args, **kwargs):
         except Exception as e:
             logger.error(e)
             return HttpResponseBadRequest(e)
+
+
+class VisualizationToolProxy(Proxy, object):
```
mccalluc: It was already a visualization-tool-proxy: Perhaps rename to reflect the difference? `AutoRelaunchProxy` or something like that.

refinery/tool_manager/utils.py
```diff
@@ -565,3 +565,8 @@ def create_expanded_workflow_graph(galaxy_workflow_dict):
             )
             graph[parent_node_id][current_node_id]['input_id'] = edge_input_id
     return graph
+
+
+def user_has_access_to_tool(user, tool):
+    return \
+        False if not user.has_perm('core.read_dataset', tool.dataset) else True
```
mccalluc: just `return user.has_perm('core.read_dataset', tool.dataset)`?

refinery/tool_manager/views.py
mccalluc: Would this be better left in the file where it's used? Or would that be weird?

refinery/tool_manager/views.py
```diff
@@ -164,3 +168,59 @@ def relaunch(self, request, *args, **kwargs):
...
+    def _proxy_view(self, request, container_name, url):
+        visualization_tool = get_object_or_404(
+            VisualizationTool,
+            container_name=container_name
+        )
+        if not user_has_access_to_tool(request.user, visualization_tool):
+            return HttpResponseForbidden(
+                "Requesting User does not have sufficient permissions to "
+                "view Tool with uuid: {}".format(visualization_tool.uuid)
+            )
```
mccalluc: Could be more concise? "User does not have permission to view Tool {}"

mccalluc:
(But test failures could be related?)

refinery/tool_manager/tests.py
scottx611x: It is unfortunately, Normal django routing middleware tacks this on automatically. The `format()` is probably overkill for this though

refinery/tool_manager/views.py
scottx611x: Agree with the renaming, I'd push to keep it in `views.py` since we're wrapping a view now

Issue #1187, opened 2016-06-06
Handle bulk deletion of Workflows, Analyses, and Datasets properly


scottx611x:
Commit: fbc5a7b0d437d2117029e1d04baea8f654bc67b9

### Steps to reproduce
* Perform a `QuerySet delete()` upon one of the models mentioned.

### Observed behavior
* If you debug the deletion process you can see that to achieve the proper deletion checking, the `QuerySet` is iterated over and each object is individual checked if it can be deleted. Then each object that passes these checks is deleted using the Model's `delete()` method.

### Expected behavior
* The checks that are in place to see if Workflows, Analyses, and Datasets can be deleted should remain, but there should be some logic introduced to perform a bulk deletion on the objects that pass the checks as individual deletions could become costly in the long run.

hackdna:
see also: #1425

hackdna:
Simply use [pre-delete](https://docs.djangoproject.com/en/1.8/ref/signals/#pre-delete) or [post-delete](https://docs.djangoproject.com/en/1.8/ref/signals/#post-delete) signal handlers.

PR #2646, opened 2018-03-02
Scottx611x/delete related Dataset & Analysis objects

### LINKED ISSUE #1428: Delete related objects when analysis is deleted
- Specific code commit: 738eba5c
- Version of the web browser and OS: N/A
- Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): Vagrant VM, dev mode
### Steps to reproduce
1. Import a Galaxy workflow and a data set
2. Run the workflow on the data set
3. Delete the resulting analysis in the Django admin (not a bulk delete)
### Observed behavior

Related objects like WorkflowDataInputMap, WorkflowFilesDL (others?) are not deleted
### Expected behavior

Related objects should be deleted

### LINKED ISSUE #1580: DataSet deletion can break half-way

### LINKED ISSUE #2632: Remove related objects when deleting a data set
* Specific code commit: 64ef52b
* Version of the web browser and OS: Chrome 64 on OS X 10.12.6
* Environment where the error occurred: Vagrant

### Steps to reproduce
1. Import a [data set](http://data.cloud.refinery-platform.org.s3.amazonaws.com/data/sample-data/human-chipseq/hg19-metadata-local.txt)
1. Navigate to the launch pad and delete this data set

### Observed behavior
The following instances remain in the database:
* 4 FileStoreItem
* 16 Node
* 1 Investigation
* 1 Study
* 1 Assay
* 2 NodeCollection

Also, It is impossible to delete some of the remaining objects manually (e.g., Investigation)

### Expected behavior
* All objects related to the data set should be deleted
* It should be possible to delete remaining objects manually

### Notes
Refinery log:
```
2018-02-27 17:17:45 DEBUG    core.models:741 get_isa_archive() - Couldn't fetch FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:558 delete() - DataSet has no isa_archive to delete: 'NoneType' object has no attribute 'delete'
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.models:2185 _nodecollection_delete() - Could not delete FileStoreItem:FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:572 delete() - Couldn't delete NodeCollection: Cannot query "None: hg19-metadata-local: 52": Must be "Study" instance.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 ERROR    core.models:820 get_file_store_items() - Error while fetching FileStoreItem: FileStoreItem matching query does not exist.
2018-02-27 17:17:45 DEBUG    core.utils:374 delete_data_set_index() - Deleted data set (uuid: 0acdf4c4-4e85-43b6-bb8f-64a3f2599fe0) index
2018-02-27 17:17:45 INFO     pysolr:325 _send_request() - Finished 'http://localhost:8983/solr/core/update/?commit=true' (post) with body 'u'<delete>' in 0.160 seconds.
2018-02-27 17:17:45 DEBUG    core.utils:390 delete_data_set_neo4j() - Deleted data set (uuid: 0acdf4c4-4e85-43b6-bb8f-64a3f2599fe0) in Neo4J
```

Also, when trying to delete some of the remaining objects manually:
```
2018-02-27 17:33:36 ERROR    django.request:256 handle_uncaught_exception() - Internal Server Error: /admin/data_set_manager/investigation/52/delete/
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/contrib/admin/options.py", line 618, in wrapper
    return self.admin_site.admin_view(view)(*args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/utils/decorators.py", line 110, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/views/decorators/cache.py", line 57, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/contrib/admin/sites.py", line 233, in inner
    return view(request, *args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/utils/decorators.py", line 34, in _wrapper
    return bound_func(*args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/utils/decorators.py", line 110, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/utils/decorators.py", line 30, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/utils/decorators.py", line 145, in inner
    return func(*args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/contrib/admin/options.py", line 1718, in delete_view
    self.delete_model(request, obj)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/contrib/admin/options.py", line 1086, in delete_model
    obj.delete()
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/db/models/base.py", line 896, in delete
    collector.delete()
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/db/models/deletion.py", line 287, in delete
    sender=model, instance=obj, using=self.using
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/dispatch/dispatcher.py", line 189, in send
    response = receiver(signal=self, sender=sender, **named)
  File "/vagrant/refinery/core/models.py", line 2180, in _nodecollection_delete
    nodes = Node.objects.filter(study=instance)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/db/models/query.py", line 679, in filter
    return self._filter_or_exclude(False, *args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/db/models/query.py", line 697, in _filter_or_exclude
    clone.query.add_q(Q(*args, **kwargs))
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/db/models/sql/query.py", line 1310, in add_q
    clause, require_inner = self._add_q(where_part, self.used_aliases)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/db/models/sql/query.py", line 1338, in _add_q
    allow_joins=allow_joins, split_subq=split_subq,
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/db/models/sql/query.py", line 1179, in build_filter
    self.check_related_objects(field, value, opts)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/db/models/sql/query.py", line 1074, in check_related_objects
    self.check_query_object_type(value, opts)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/db/models/sql/query.py", line 1058, in check_query_object_type
    (value, opts.object_name))
ValueError: Cannot query "None: hg19-metadata-local: 52": Must be "Study" instance.
[27/Feb/2018 17:33:36] "POST /admin/data_set_manager/investigation/52/delete/ HTTP/1.1" 500 224870
```

scottx611x:
Resolves #2632 
Resolves #1580 
Resolves #1428 
Towards #1187

### DIFF
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -518,15 +518,17 @@ class DataSetResource(SharableResourceAPIInterface, ModelResource):
             except:
                 pass
 
-        isa_archive = bundle.obj.get_isa_archive()
-        pre_isa_archive = bundle.obj.get_pre_isa_archive()
+        investigation = bundle.obj.get_investigation()
+        if investigation is not None:
+            metadata_file_store_item = \
+                bundle.obj.get_metadata_as_file_store_item()
 
-        if isa_archive:
-            bundle.data["isa_archive"] = isa_archive.uuid
-            bundle.data["isa_archive_url"] = isa_archive.get_datafile_url()
-
-        if pre_isa_archive:
-            bundle.data["pre_isa_archive"] = pre_isa_archive.uuid
+            if investigation.is_isatab_based:
+                bundle.data["isa_archive"] = metadata_file_store_item.uuid
+                bundle.data["isa_archive_url"] = \
+                    metadata_file_store_item.get_datafile_url()
+            else:
+                bundle.data["pre_isa_archive"] = metadata_file_store_item.uuid
 
         analyses = []
         for analysis in bundle.obj.get_analyses():
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -518,15 +518,15 @@ class DataSetResource(SharableResourceAPIInterface, ModelResource):
             except:
                 pass
 
-            metadata_file_store_item = \
-                bundle.obj.get_metadata_as_file_store_item()
+        metadata_file_store_item = \
+            bundle.obj.get_metadata_as_file_store_item()
 
-            if bundle.obj.is_isatab_based:
-                bundle.data["isa_archive"] = metadata_file_store_item.uuid
-                bundle.data["isa_archive_url"] = \
-                    metadata_file_store_item.get_datafile_url()
-            else:
-                bundle.data["pre_isa_archive"] = metadata_file_store_item.uuid
+        if bundle.obj.is_isatab_based:
+            bundle.data["isa_archive"] = metadata_file_store_item.uuid
+            bundle.data["isa_archive_url"] = \
+                metadata_file_store_item.get_datafile_url()
+        else:
+            bundle.data["pre_isa_archive"] = metadata_file_store_item.uuid
 
         analyses = []
         for analysis in bundle.obj.get_analyses():
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -518,12 +518,10 @@ class DataSetResource(SharableResourceAPIInterface, ModelResource):
             except:
                 pass
 
-        investigation = bundle.obj.get_investigation()
-        if investigation is not None:
             metadata_file_store_item = \
                 bundle.obj.get_metadata_as_file_store_item()
 
-            if investigation.is_isatab_based:
+            if bundle.obj.is_isatab_based:
                 bundle.data["isa_archive"] = metadata_file_store_item.uuid
                 bundle.data["isa_archive_url"] = \
                     metadata_file_store_item.get_datafile_url()
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -56,12 +56,12 @@ from galaxy_connector.models import Instance
 import tool_manager
 
 from .utils import (add_or_update_user_to_neo4j, add_read_access_in_neo4j,
-                    async_update_annotation_sets_neo4j, delete_analysis_index,
-                    delete_data_set_index, delete_data_set_neo4j,
-                    delete_ontology_from_neo4j, delete_user_in_neo4j,
-                    email_admin, invalidate_cached_object,
-                    remove_read_access_in_neo4j, skip_if_test_run,
-                    sync_update_annotation_sets_neo4j, update_data_set_index)
+                    async_update_annotation_sets_neo4j, delete_data_set_index,
+                    delete_data_set_neo4j, delete_ontology_from_neo4j,
+                    delete_user_in_neo4j, email_admin,
+                    invalidate_cached_object, remove_read_access_in_neo4j,
+                    skip_if_test_run, sync_update_annotation_sets_neo4j,
+                    update_data_set_index)
 
 logger = logging.getLogger(__name__)
 
@@ -987,17 +987,6 @@ class AnalysisResult(models.Model):
         )
 
 
-class AnalysisQuerySet(models.query.QuerySet):
-    def delete(self):
-        for instance in self:
-            instance.delete()
-
-
-class AnalysisManager(models.Manager):
-    def get_queryset(self):
-        return AnalysisQuerySet(self.model, using=self._db)
-
-
 class Analysis(OwnableResource):
     SUCCESS_STATUS = "SUCCESS"
     FAILURE_STATUS = "FAILURE"
@@ -1030,8 +1019,6 @@ class Analysis(OwnableResource):
     # output_nodes = models.ManyToManyField(Nodes, blank=True)
     # protocol = i.e. protocol node created when the analysis is created
 
-    objects = AnalysisManager()
-
     def __str__(self):
         return "{} - {} - {}".format(
             self.name,
@@ -1052,6 +1039,20 @@ class Analysis(OwnableResource):
             ast.literal_eval(self.workflow_copy)
         )
 
+    def has_nodes_used_in_downstream_analyses(self):
+        """
+        Check if any Nodes that are part of an Analysis have been Analyzed
+        further down the line.
+        :return: Boolean
+        """
+        for node in self.get_nodes():
+            analysis_node_connections_for_node = \
+                node.get_analysis_node_connections()
+            for analysis_node_connection in analysis_node_connections_for_node:
+                if analysis_node_connection.direction == 'in':
+                    return True
+        return False
+
     def delete(self, **kwargs):
         """
         Overrides the Analysis model's delete method and checks if
@@ -1066,59 +1067,14 @@ class Analysis(OwnableResource):
         5. Continue on to delete the Analysis, AnalysisNodeConnections,
         and AnalysisStatus'
         """
-
-        delete = True
-        nodes = self.get_nodes()
-
-        for node in nodes:
-
-            analysis_node_connections_for_node = \
-                node.get_analysis_node_connections()
-
-            for analysis_node_connection in analysis_node_connections_for_node:
-                if analysis_node_connection.direction == 'in':
-                    delete = False
-
-        if delete:
-            self.cancel()
-
-            # Delete associated AnalysisResults
-            self.get_analysis_results().delete()
-
-            for node in nodes:
-                # Delete associated FileStoreItems
-                if node.file_uuid:
-                    node.get_file_store_item().delete()
-
-                # Remove Nodes from Solr's Index
-                try:
-                    delete_analysis_index(node)
-                except Exception as e:
-                    logger.debug("No NodeIndex exists in Solr with id "
-                                 "%s:  %s", node.id, e)
-
-            # Optimize Solr's index to get rid of any traces of the Analysis
-            self.optimize_solr_index()
-
-            # Delete Nodes Associated w/ the Analysis
-            nodes.delete()
-
-            super(Analysis, self).delete()
-
-            # Return a "truthy" value here so that the admin ui and
-            # front-end ui knows if the deletion succeeded or not as well as
-            # the proper message to display to the end user
-            return True, "Analysis: {} was deleted successfully".format(
-                self.name)
-
-        else:
-            # Prepare string to be displayed upon a failed deletion
-            deletion_error_message = "Cannot delete Analysis: {} because " \
-                                     "its results have been used to run " \
-                                     "further Analyses. Please delete all " \
-                                     "downstream Analyses before you delete " \
-                                     "this one".format(self.name)
-
+        if self.has_nodes_used_in_downstream_analyses():
+            deletion_error_message = (
+                "Cannot delete Analysis: {} because "
+                "its results have been used to run "
+                "further Analyses. Please delete all "
+                "downstream Analyses before you delete "
+                "this one".format(self.name)
+            )
             logger.error(deletion_error_message)
 
             # Return a "falsey" value here so that the admin ui and
@@ -1126,12 +1082,23 @@ class Analysis(OwnableResource):
             # the proper message to display to the end user
             return False, deletion_error_message
 
+        try:
+            super(Analysis, self).delete()
+        except Exception as exc:
+            return False, "Analysis {} could not be deleted: {}".format(
+                self.name, exc)
+        else:
+            # Return a "truthy" value here so that the admin ui and
+            # front-end ui knows if the deletion succeeded or not as well
+            # as the proper message to  display to the end user
+            return True, "Analysis: {} was deleted successfully".format(
+                self.name)
+
     def get_status(self):
         return self.status
 
     def get_nodes(self):
-        return Node.objects.filter(
-            analysis_uuid=self.uuid)
+        return Node.objects.filter(analysis_uuid=self.uuid)
 
     def get_analysis_results(self):
         return AnalysisResult.objects.filter(analysis_uuid=self.uuid)
@@ -1661,6 +1628,27 @@ class Analysis(OwnableResource):
             self.get_input_node_assay().uuid
         )
         self._prepare_annotated_nodes(node_uuids)
+
+
+@receiver(pre_delete, sender=Analysis)
+def _analysis_delete(sender, instance, *args, **kwargs):
+    """
+    Removes an Analyses's related objects upon deletion being triggered.
+    Having these extra checks is favored within a signal so that this logic
+    is picked up on bulk deletes as well.
+
+    See: https://docs.djangoproject.com/en/1.8/topics/db/models/
+    #overriding-model-methods
+    """
+    with transaction.atomic():
+        instance.cancel()
+        # Delete associated AnalysisResults
+        instance.get_analysis_results().delete()
+        # Delete Nodes Associated w/ the Analysis
+        instance.get_nodes().delete()
+
+    # Optimize Solr's index to get rid of any traces of the Analysis
+    instance.optimize_solr_index()
 
 
 #: Defining available relationship types
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -754,6 +754,7 @@ class DataSet(SharableResource):
                        FileStoreItem.MultipleObjectsReturned) as e:
                     logger.error("Error while fetching FileStoreItem: %s", e)
 
+        file_store_items.append(self.get_metadata_as_file_store_item())
         return file_store_items
 
     def is_valid(self):
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -523,34 +523,8 @@ class DataSet(SharableResource):
         super(DataSet, self).save(*args, **kwargs)
 
     def delete(self, **kwargs):
-        """
-        Overrides the DataSet model's delete method.
-
-        Deletes NodeCollection and related object based on uuid of
-        Investigations linked to the DataSet.
-        This deletes Studys, Assays and Investigations in
-        addition to the related objects detected by Django.
-
-        This method will also delete the isa_archive or
-        pre_isa_archive associated with the DataSet if one exists.
-        """
-
-        # terminate any running file import tasks
-        for file_store_item in self.get_file_store_items():
-            file_store_item.terminate_file_import_task()
-
-        related_investigation_links = self.get_investigation_links()
-
         try:
-            with transaction.atomic():
-                # delete FileStoreItem and datafile corresponding to the
-                # metadata file used to generate the DataSet
-                self.get_metadata_as_file_store_item().delete()
-
-                for investigation_link in related_investigation_links:
-                    investigation_link.get_node_collection().delete()
-
-                super(DataSet, self).delete()
+            super(DataSet, self).delete()
         except Exception as exc:
             return False, "DataSet {} could not be deleted: {}".format(
                 self.name, exc)
@@ -806,6 +780,29 @@ class DataSet(SharableResource):
 
 @receiver(pre_delete, sender=DataSet)
 def _dataset_delete(sender, instance, *args, **kwargs):
+    """
+    Removes a DataSet's related objects upon deletion being triggered.
+    Having these extra checks is favored within a signal so that this logic
+    is picked up on bulk deletes as well.
+
+    See: https://docs.djangoproject.com/en/1.8/topics/db/models/
+    #overriding-model-methods
+    """
+
+    # terminate any running file import tasks
+    for file_store_item in instance.get_file_store_items():
+        file_store_item.terminate_file_import_task()
+
+    related_investigation_links = instance.get_investigation_links()
+
+    with transaction.atomic():
+        # delete FileStoreItem and datafile corresponding to the
+        # metadata file used to generate the DataSet
+        instance.get_metadata_as_file_store_item().delete()
+
+        for investigation_link in related_investigation_links:
+            investigation_link.get_node_collection().delete()
+
     delete_data_set_index(instance)
     delete_data_set_neo4j(instance.uuid)
     async_update_annotation_sets_neo4j()
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -682,11 +682,10 @@ class DataSet(SharableResource):
         """Returns the FileStoreItem pointing to the isa/pre_isa archive that
         was used to create the DataSet"""
         investigation = self.get_investigation()
-        if investigation is not None:
-            return (
-                investigation.isa_archive if investigation.is_isatab_based
-                else investigation.pre_isa_archive
-            )
+        return (
+            investigation.isa_archive if self.is_isatab_based
+            else investigation.pre_isa_archive
+        )
 
     def share(self, group, readonly=True, readmetaonly=False):
         # change: !readonly & !readmetaonly, read: readonly & !readmetaonly
@@ -776,6 +775,10 @@ class DataSet(SharableResource):
 
     def get_node_uuids(self):
         return self.get_nodes().values_list('uuid', flat=True)
+
+    @property
+    def is_isatab_based(self):
+        return True if self.get_investigation().isa_archive else False
 
 
 @receiver(pre_delete, sender=DataSet)
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -2157,8 +2157,6 @@ def _nodecollection_delete(sender, instance, **kwargs):
     if type(instance) == Study:
         nodes = Node.objects.filter(study=instance)
         for node in nodes:
-            if node.file_uuid:
-                FileStoreItem.objects.get(uuid=node.file_uuid).delete()
             node.delete()
 
 
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -552,16 +552,6 @@ class DataSet(SharableResource):
         pre_isa_archive associated with the DataSet if one exists.
         """
 
-        try:
-            self.get_isa_archive().delete()
-        except AttributeError as e:
-            logger.debug("DataSet has no isa_archive to delete: %s", e)
-
-        try:
-            self.get_pre_isa_archive().delete()
-        except AttributeError as e:
-            logger.debug("DataSet has no pre_isa_archive to delete: %s", e)
-
         related_investigation_links = self.get_investigation_links()
 
         for investigation_link in related_investigation_links:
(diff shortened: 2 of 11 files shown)


refinery/core/api.py
```diff
@@ -518,15 +518,17 @@ def dehydrate(self, bundle):
...
-        isa_archive = bundle.obj.get_isa_archive()
-        pre_isa_archive = bundle.obj.get_pre_isa_archive()
-
-        if isa_archive:
-            bundle.data["isa_archive"] = isa_archive.uuid
-            bundle.data["isa_archive_url"] = isa_archive.get_datafile_url()
-
-        if pre_isa_archive:
-            bundle.data["pre_isa_archive"] = pre_isa_archive.uuid
+        investigation = bundle.obj.get_investigation()
```
scottx611x: This cuts ORM queries for FileStoreItems in half compared to the previous approach when GET-ing from api/v1/datasets/

refinery/core/models.py
```diff
@@ -2178,14 +2147,12 @@ def _nodecollection_delete(sender, instance, **kwargs):
...
     """
-    nodes = Node.objects.filter(study=instance)
-    for node in nodes:
-        try:
-            FileStoreItem.objects.get(uuid=node.file_uuid).delete()
-        except (FileStoreItem.DoesNotExist,
-                FileStoreItem.MultipleObjectsReturned) as exc:
-            logger.debug("Could not delete FileStoreItem with UUID '%s': %s",
-                         node.uuid, exc)
+    if type(instance) == Study:
```
scottx611x: Investigations are NodeCollections as well, but they get removed when a Study does due to the ForeignKey rel.

refinery/data_set_manager/models.py
```diff
@@ -154,6 +154,33 @@ class Investigation(NodeCollection):
...
+    def pre_isa_archive(self):
+        try:
+            return FileStoreItem.objects.get(uuid=self.pre_isarchive_file)
+        except (FileStoreItem.DoesNotExist,
+                FileStoreItem.MultipleObjectsReturned) as e:
+            logger.error(
+                "Couldn't fetch Investigation: %s's FileStoreItem from UUID: "
+                "%s %s", self, self.pre_isarchive_file, e
+            )
+
```
hackdna: Is there an actual need for both isarchive_file and pre_isarchive_file?

refinery/core/models.py
```diff
@@ -730,27 +716,15 @@ def get_file_size(self):
...
-        try:
-            return FileStoreItem.objects.get(uuid=investigation.isarchive_file)
-        except (FileStoreItem.DoesNotExist,
-                FileStoreItem.MultipleObjectsReturned,
-                AttributeError) as e:
-            logger.debug("Couldn't fetch FileStoreItem: %s", e)
-
-    def get_pre_isa_archive(self):
-        """Returns the pre_isa_archive that was used to create the DataSet"""
+    def get_metadata_as_file_store_item(self):
```
hackdna: This method appears to be unnecessary as this logic is repeated in the only place where it's used.

refinery/core/models.py
```diff
@@ -552,31 +552,17 @@ def delete(self, **kwargs):
...
-
-        related_investigation_links = self.get_investigation_links()
-
-        for investigation_link in related_investigation_links:
-            node_collection = investigation_link.get_node_collection()
-            try:
-                node_collection.delete()
-            except Exception as e:
-                logger.error("Couldn't delete NodeCollection: %s", e)
-
```
hackdna: DataSet.delete() should be replaced by pre_ and/or post_delete signal handlers to ensure this functionality is executed on bulk operations
https://docs.djangoproject.com/en/1.8/topics/db/models/#overriding-model-methods

refinery/core/models.py
```diff
@@ -552,31 +552,17 @@ def delete(self, **kwargs):
...
             file_store_item.terminate_file_import_task()
 
+        related_investigation_links = self.get_investigation_links()
+
         try:
-            super(DataSet, self).delete()
+            with transaction.atomic():
+                for investigation_link in related_investigation_links:
+                    investigation_link.get_node_collection().delete()
+                super(DataSet, self).delete()
```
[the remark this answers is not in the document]
scottx611x: Yes I believe it should stay. Imagine if someone extended parent's `.delete()` and that code now could cause a faulty child deletion

refinery/core/tests.py
```diff
@@ -611,87 +614,62 @@ def test_verify_deletion_if_workflow_not_used_in_analysis(self):
...
-            name="dataset_with_analysis")
-        self.dataset_without_analysis = \
-            DataSet.objects.create(name="dataset_without_analysis")
-        self.investigation_link = \
-            InvestigationLink.objects.create(
-                investigation=self.investigation,
-                data_set=self.dataset_without_analysis)
+    def _assert_related_objects_exist(self):
+        for model in self.models_to_be_removed:
+            self.assertNotEqual(model.objects.count(), 0)
```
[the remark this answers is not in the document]
scottx611x: What is wrong with having private test case methods that aid in asserting a certain state has been achieved? I've seen this pattern be helpful when making the same assertions many times.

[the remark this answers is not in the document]
mccalluc: I'm not sure I agree with that: The intention of what's being asserted here ("`related_objects_exist`") is more clear than if the two lines inside are copied and pasted where it's used.

refinery/data_set_manager/models.py
```diff
@@ -154,6 +154,33 @@ class Investigation(NodeCollection):
     pre_isarchive_file = UUIDField(blank=True, null=True, auto=False)
 
     """easily retrieves the proper NodeCollection fields"""
+
+    @property
+    def is_isatab_based(self):
+        return True if self.isa_archive else False
```
mccalluc: You don't need the ternary, and I'm not sure you need the method at all?

refinery/core/models.py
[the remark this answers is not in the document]
hackdna: This could cause hard to diagnose errors if exception handling is added to delete() in parent. "Avoid catching exceptions inside atomic" https://docs.djangoproject.com/en/1.8/topics/db/transactions/#controlling-transactions-explicitly

refinery/core/models.py
hackdna: I was referring to `if investigation.is_isatab_based` statement that is repeated both inside this method and in api.py which is a bit of a code smell.
As a matter of fact, there may not be a need for any of this logic if isa_archive and pre_isa_archive can be combined into a single field.

refinery/core/models.py
```diff
@@ -2178,14 +2147,14 @@ def _nodecollection_delete(sender, instance, **kwargs):
...
-                FileStoreItem.MultipleObjectsReturned) as exc:
-            logger.debug("Could not delete FileStoreItem with UUID '%s': %s",
-                         node.uuid, exc)
+    # Investigations are NodeCollections as well, but they get removed when
+    # a Study does due to the ForeignKey rel. between the two
+    if type(instance) == Study:
+        nodes = Node.objects.filter(study=instance)
+        for node in nodes:
+            if node.file_uuid:
+                FileStoreItem.objects.get(uuid=node.file_uuid).delete()
```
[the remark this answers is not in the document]
hackdna: Better yet to a pre_ or post_delete signal handler.

refinery/core/tests.py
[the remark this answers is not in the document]
hackdna: Yes, and it also makes it difficult to troubleshoot test failures. Two Scoops of Django 22.3.4: "Don't Repeat Yourself Doesn't Apply to Writing Tests"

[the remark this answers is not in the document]
scottx611x: I like two scoops a lot. Sometimes I find the suggestions to be opinionated, and I've learned to use it as a tool for guidance rather than as law. I think in this instance writing:
```
for model in self.models_to_be_removed:
    self.assertNotEqual(model.objects.count(), 0)
```
alone is less clear than calling `self._assert_related_objects_exist()`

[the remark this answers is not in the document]
hackdna: Tests should be written as simply as possible. Otherwise, troubleshooting failures becomes very difficult which defeats the purpose of writing tests.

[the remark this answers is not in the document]
scottx611x: Looking at this code again I believe this line `self.assertNotEqual(model.objects.count(), 0)` would mask which `model` failed the given assertion:
```
  File "/vagrant/refinery/core/tests.py", line 660, in test_dataset_deletion_removes_related_objects
    self._assert_related_objects_removed()
  File "/vagrant/refinery/core/tests.py", line 646, in _assert_related_objects_removed
    self.assertEqual(model.objects.count(), 0)
AssertionError: 1 != 0
```

Newer Django/Unittest have solved this type of issue with the use of [`subTest()`](https://docs.python.org/3/library/unittest.html#subtests) also see[here](https://www.caktusgroup.com/blog/2017/05/29/subtests-are-best/)

but for now we would have to do something like this for easy debugging of test failures:
```
    def _assert_related_objects_exist(self):
        self.assertGreater(Analysis.objects.count(), 0)
        self.assertGreater(AnnotatedNode.objects.count(), 0)
        self.assertGreater(Assay.objects.count(), 0)
        self.assertGreater(DataSet.objects.count(), 0)
        self.assertGreater(FileStoreItem.objects.count(), 0)
        self.assertGreater(Investigation.objects.count(), 0)
        self.assertGreater(InvestigationLink.objects.count(), 0)
        self.assertGreater(Node.objects.count(), 0)
        self.assertGreater(NodeCollection.objects.count(), 0)
        self.assertGreater(Study.objects.count(), 0)

    def _assert_related_objects_removed(self):
        self.assertEqual(Analysis.objects.count(), 0)
        self.assertEqual(AnnotatedNode.objects.count(), 0)
        self.assertEqual(Assay.objects.count(), 0)
        self.assertEqual(DataSet.objects.count(), 0)
        self.assertEqual(FileStoreItem.objects.count(), 0)
        self.assertEqual(Investigation.objects.count(), 0)
        self.assertEqual(InvestigationLink.objects.count(), 0)
        self.assertEqual(Node.objects.count(), 0)
        self.assertEqual(NodeCollection.objects.count(), 0)
        self.assertEqual(Study.objects.count(), 0)
```

Which would yield something like the following on an error:
```
Failure
  File "/vagrant/refinery/core/tests.py", line 660, in test_dataset_deletion_removes_related_objects
    self._assert_related_objects_removed()
  File "/vagrant/refinery/core/tests.py", line 645, in _assert_related_objects_removed
    self.assertEqual(FileStoreItem.objects.count(), 0)
AssertionError: 1 != 0
```

@hackdna I think that remaining DRY with this approach is favorable. One introducing more tests in this space in the future could easily miss one of the above assertions with a copy/paste approach rather than reusing `_assert_related_objects_exist` or `_assert_related_objects_removed`.

I'm not seeing the difficulty in troubleshooting a failure like the one above.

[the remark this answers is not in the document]
hackdna: The subTest() method may be useful in limited cases when you have small changes in test conditions, such as running the same test with a range of integers (as it is shown in the unittest docs).

Test cases in the Caktus Group article are too long and should be broken down into individual tests for clarity. Use of subTest() there is superfluous and only obfuscates the code.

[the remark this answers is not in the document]
scottx611x: I think the `subTest()` code would work fine for the problem I illustrated. Allowing for code like this:
```
for model in self.models_to_be_removed:
    with self.subTest(model=model):
        self.assertNotEqual(model.objects.count(), 0)
```

to yield an error like this:

```
Failure (model=FileStoreItem)
  File "/vagrant/refinery/core/tests.py", line 660, in test_dataset_deletion_removes_related_objects
    self._assert_related_objects_removed()
  File "/vagrant/refinery/core/tests.py", line 646, in _assert_related_objects_removed
    self.assertEqual(model.objects.count(), 0)
AssertionError: 1 != 0
```  allowing one to see which `model` had caused the assertion failure.

I linked the Caktus article just to show some usages of `subTest()`. It was my understanding that they chose particularly long tests in the article to better illustrate how it could consolidate testing code.

But thats all besides the fact because I wasn't advocating for the use of this feature since we don't have the ability to right now or in the near future.

---

My point out of my prior comment was that to have easy to debug test failures without succumbing to copy/pasting ~10 assertions every time that we want to check that a dataset's related objects have been removed/remain seems to naturally lead to writing something reusable like: `_assert_related_objects_exist()`.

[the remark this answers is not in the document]
Comment (hackdna) [reply refinery/core/tests.py]: There is no need to check if a data set was created properly in every test. The same goes for deletion. Right now a lot of tests are re-testing functionality covered by other tests. Test code will naturally become DRY when this is no longer the case.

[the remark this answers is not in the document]
Comment (scottx611x) [reply refinery/core/tests.py]: I'm confused, we aren't checking "if a data set was created properly in every test". Nor are we for `_assert_related_objects_removed`. Those are only used when they need to be, it just happens to be in  a few places in this PR.

When you say: "Right now a lot of tests are re-testing functionality covered by other tests", which tests are you talking about in the context of this pull request?

I think all of the tests within `DataSetDeletionTest` are distinct use cases that warrant separate testing.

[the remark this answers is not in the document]
Comment (hackdna) [reply refinery/core/tests.py]: For example, the functionality tested by`_assert_related_objects_exist()` is used a total of three times but it does not need to be present in `test_dataset_deletion_removes_related_objects()` and `test_dataset_bulk_deletion_removes_related_objects()` because it has already been tested in `test_transaction_rollback_on_delete_failure()`.

Functionality tested by `_assert_related_objects_removed()` is used a total of two times. Now, if I need to troubleshoot a failure in one of those tests, I'd need to understand not only the logic of the code under test but also, I'd need to look up whatever `_assert_related_objects_removed()` does and then what `self.models_to_be_removed` is. This adds unnecessary complexity and makes troubleshooting more difficult - just to save a few lines of code.

File: refinery/core/models.py [hunk 10]
```diff
@@ -2116,14 +2074,12 @@ def _nodecollection_delete(sender, instance, **kwargs):
...
-    for node in nodes:
-        try:
-            FileStoreItem.objects.get(uuid=node.file_uuid).delete()
-        except (FileStoreItem.DoesNotExist,
-                FileStoreItem.MultipleObjectsReturned) as exc:
-            logger.debug("Could not delete FileStoreItem with UUID '%s': %s",
-                         node.uuid, exc)
+    # Investigations are NodeCollections as well, but they get removed when
+    # a Study does due to the ForeignKey rel. between the two
+    if type(instance) == Study:
```
scottx611x: no reciever_subclasses just do for Study

scottx611x:
@hackdna I looked into the consolidation of `Investigation.isarchive_file` & `Investigation.pre_isarchive_file` into a single field and I think its too big of a change for the scope of this PR, but should still definitely be looked into further. I'll write up an issue

refinery/factory_boy/utils.py
```diff
@@ -1,11 +1,15 @@
-from datetime import datetime
 import uuid as uuid_builtin
 
-from core.models import Analysis, DataSet, Node
+from core.models import INPUT_CONNECTION, OUTPUT_CONNECTION, Analysis
+from data_set_manager.models import Node
 from factory_boy.django_model_factories import (AnalysisFactory,
+                                                AnalysisNodeConnectionFactory,
```
hackdna: Use the whole allowed width of the line

refinery/factory_boy/utils.py
```diff
@@ -15,15 +19,59 @@
...
 
-def make_datasets(number_to_create, user_instance):
-    """Create some minimal DataSets"""
-    while number_to_create:
-        create_dataset_with_necessary_models()
-        number_to_create -= 1
+def create_analysis(project, dataset, workflow, user_instance):
+    analysis_uuid = str(uuid_builtin.uuid4())
+    analysis = AnalysisFactory(
+        uuid=analysis_uuid,
```
hackdna: Use the whole allowed width of the line

refinery/factory_boy/utils.py
```diff
@@ -70,43 +105,40 @@ def create_dataset_with_necessary_models(
...
-        date=datetime.now()
-    )
-
-    study_uuid = str(uuid_builtin.uuid4())
-    study = StudyFactory(
-        uuid=study_uuid,
-        investigation=investigation,
-        description="This is a great DataSet"
+    latest_study = _create_dataset_objects(
+        dataset,
```
hackdna: Use the whole allowed width of the line

refinery/factory_boy/utils.py
```diff
@@ -70,43 +105,40 @@ def create_dataset_with_necessary_models(
...
     if create_nodes:
         for i in xrange(2):
+            file_store_item_uuid = str(uuid_builtin.uuid4())
+            FileStoreItemFactory(
+                uuid=file_store_item_uuid,
+                source="http://www.example.com/test{}.txt".format(i)
+            )
             node = NodeFactory(
-                study=study
+                study=latest_study,
```
hackdna: Use the whole allowed width of the line

refinery/factory_boy/utils.py
```diff
@@ -70,43 +105,40 @@ def create_dataset_with_necessary_models(
...
+        dataset,
+        is_isatab_based,
+        latest_version
     )
 
     assay_uuid = str(uuid_builtin.uuid4())
     assay = AssayFactory(
         uuid=assay_uuid,
-        study=study
+        study=latest_study
```
hackdna: This should fit on one line

refinery/factory_boy/utils.py
```diff
@@ -115,3 +147,40 @@ def create_dataset_with_necessary_models(
...
+                None if is_isatab_based else file_store_item.uuid
+            ),
+            identifier="{}: Investigation identifier".format(dataset),
+            description="{}: Investigation description".format(dataset),
+            title="{}: Investigation title".format(dataset)
+        )
+
+        study_uuid = str(uuid_builtin.uuid4())
+        study = StudyFactory(
+            uuid=study_uuid,
```
hackdna: Use the whole allowed width of the line

refinery/factory_boy/utils.py
```diff
@@ -115,3 +147,40 @@ def create_dataset_with_necessary_models(
...
+            investigation=investigation,
+            identifier="{}: Study identifier".format(dataset),
+            description="{}: Study description".format(dataset),
+            title="{}: Study title".format(dataset)
+        )
+        InvestigationLinkFactory(
```
hackdna: Use the whole allowed width of the line

refinery/factory_boy/utils.py
scottx611x: `isort` formats things this way. We'd have to change its configuration if want wanted something different

hackdna: OK then we should change its config.

refinery/factory_boy/utils.py
scottx611x: @hackdna I don't think there is anything specific in our style guide about this, and this pattern has been in use by myself and others for awhile now. I'm not sure why all of a sudden its blocking for this PR. 

I'd be open to a style change if you wanted to write something up, but the olny thing close in the wiki is:
```
"If an opening parenthesis or square bracket is at the end of a line then the closing one should be on a line by itself"
```
Which this pattern doesn't violate.

hackdna: This formatting has been used throughout the existing codebase for years. It is also implied in PEP8 and it was never added to the style guide because it was never in question. I'm not sure why this deviation has happened.

scottx611x: I think deviation has probably happened, because different developers have different styles, and we haven't explicitly enforced one way to do this. 

We could probably configure flake8 to do what we want, but I'd feel better about making a decision on this as a team.

jkmarx: Yeah, let's decide and config flake8. I didn't realize there was a preferred style.

scottx611x:
Going to merge this PR. I've opened: https://github.com/refinery-platform/refinery-platform/issues/2675 to decide on a coding style for [using the whole allowed width of a line](https://github.com/refinery-platform/refinery-platform/pull/2646#discussion_r174911812)

PR #2648, opened 2018-03-05
Hackdna/update fabfile

hackdna:
Remove a redundant checkout

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -14,7 +14,7 @@ Requirements:
 
 import os
 
-from fabric.api import env, execute, local, run, sudo
+from fabric.api import env, execute, run, sudo
 from fabric.context_managers import cd, prefix, shell_env
 from fabric.contrib.files import sed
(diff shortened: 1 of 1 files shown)


fabfile.py
```diff
@@ -120,7 +118,9 @@ def update_refinery():
     """Perform full update of a Refinery Platform instance"""
     puts("Updating Refinery")
     with cd(env.refinery_project_dir):
-        run("git checkout {branch}".format(**env))
+        # update current branch in Vagrant VM, otherwise checkout custom branch
+        if env.project_user != 'vagrant':
+            run("git checkout {branch}".format(**env))
```
mccalluc: The comment confused me a little bit, because it reads like an if-then, but with the sense reversed from the code... Maybe a comment after instead, something like
```python
# ... but if using Vagrant, current branch is fine.
```
Not a big deal either way. Thanks!

PR #2655, opened 2018-03-07
Scottx611x/delete related analysis objects

### LINKED ISSUE #1428: Delete related objects when analysis is deleted

scottx611x:
Resolves #1428 
Towards #1187

### DIFF
--- a/refinery/core/tests.py
+++ b/refinery/core/tests.py
@@ -625,18 +625,6 @@ class DataSetDeletionTest(TestCase):
                 self.user
             )
 
-    def _assert_related_objects_removed(self):
-        self.assertEqual(Analysis.objects.count(), 0)
-        self.assertEqual(AnnotatedNode.objects.count(), 0)
-        self.assertEqual(Assay.objects.count(), 0)
-        self.assertEqual(DataSet.objects.count(), 0)
-        self.assertEqual(FileStoreItem.objects.count(), 0)
-        self.assertEqual(Investigation.objects.count(), 0)
-        self.assertEqual(InvestigationLink.objects.count(), 0)
-        self.assertEqual(Node.objects.count(), 0)
-        self.assertEqual(NodeCollection.objects.count(), 0)
-        self.assertEqual(Study.objects.count(), 0)
-
     def test_transaction_rollback_on_dataset_delete_failure(self):
         with mock.patch.object(BaseResource, "delete", side_effect=Exception):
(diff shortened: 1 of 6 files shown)


scottx611x:
@hackdna I've made the changes you've requested, and I still feel wrong about it. I already made errors copying and pasting all of theses assertions with my local edits and I would think someone else extending these test cases in the future could easily do the same. 

What if we just avoided your point of unnecessarily retesting of code by calling these `_assert_related_objects*` in a few less places, but kept the repetitive assertions in their own methods still?

refinery/core/tests.py
```diff
@@ -679,6 +699,71 @@ def test_pre_isa_archive_deletion(self):
...
+                1,
+                self.user
+            )
+        self.analysis = self.analyses[0]
+
+        self.assertGreater(Analysis.objects.count(), 0)
+        self.assertGreater(AnalysisNodeConnection.objects.count(), 0)
+        self.assertGreater(AnalysisResult.objects.count(), 0)
+        self.assertGreater(AnalysisStatus.objects.count(), 0)
+        self.assertGreater(Node.objects.count(), 0)
```
hackdna: No testing in setUp(), move these five lines into a test.

refinery/core/tests.py
```diff
@@ -679,6 +699,71 @@ def test_pre_isa_archive_deletion(self):
...
 
+class AnalysisDeletionTest(TestCase):
+    """Testing for the deletion of Analyses"""
+    def setUp(self):
+        self.username = self.password = 'user'
+        self.user = User.objects.create_user(
+            self.username, '', self.password
+        )
+        self.analyses, self.dataset = \
+            make_analyses_with_single_dataset(
```
hackdna: This call should fit on a single line

hackdna:
I think there is minimal code duplication and the test cases are easier to read now.

PR #2656, opened 2018-03-08
Scottx611x/Add VPC to Terraform

### LINKED ISSUE #2600: Add VPC to Terraform infrastructure
Avoid deploying into the default VPC

https://aws.amazon.com/answers/networking/aws-single-vpc-design/
http://harish11g.blogspot.com/2014/01/Amazon-Virtual-Private-Cloud-VPC-best-practices-tips-for-architecture-migration.html
https://charity.wtf/2016/03/23/aws-networking-environments-and-you/

Examples:
https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Scenario2.html
https://040code.github.io/2017/06/18/terraform-aws-vpc/
https://hackernoon.com/manage-aws-vpc-as-infrastructure-as-code-with-terraform-55f2bdb3de2a
https://nickcharlton.net/posts/terraform-aws-vpc.html

scottx611x:
Resolves #2600 

EDIT (3/15/18):
We've decided that holding off on the docker host would be good for now since there is still some work to do on the django_docker_engine side of things. See: https://github.com/refinery-platform/refinery-platform/pull/2656#issuecomment-373488258
<details>
<summary>Prior Notes</summary>
<br>
Still a WIP, but looking for help/feedback.

This allowed me to run `terraform apply`, get the new VPC/Docker related outputs, and bring up a CloudFormation Refinery stack successfully.

I had to comment out things regarding the pseudo-Refinery EC2 currently in the TF code due to errors like: `module.refinery_host.output.refinery_hostname: Resource 'aws_instance.refinery_host' not found for variable 'aws_instance.refinery_host.0.public_ip'` when running `terraform plan`. Maybe we can get rid of the `refinery-host` module/ helper script for the time being, and revisit when we're ready to "terraform-ify" our CloudFormation code.

~~~I haven't figured out how/where to expose `DOCKER_HOST` as an envvar in our codebase, but doing an `export DOCKER_HOST=tcp://<docker_host>:2376` after ssh-ing in to the Refinery EC2 allowed me to launch `VisualizationTools` through the Refinery UI and have them start on the 2nd EC2 (`DOCKER_HOST`) w/ requests proxied properly.~~~ This was False! It just seemed to me that it was working at first. 

One thing that tripped me up was getting our `RDSInstance` within the newly created `VPC`. 
It seemed to me that I had to add an [`AWS::RDS::DBSubnetGroup`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbsubnet-group.html) from the lines: [`"If there's no DB subnet group, then the instance isn't a VPC DB instance."` ](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-dbsubnetgroupname) & [`"You must have either two private subnets or two public subnets available to create an Amazon RDS DB subnet group for an RDS DB instance to use in a VPC"`](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Tutorials.WebServerDB.CreateVPC.html). But to create the `DBSubnetGroup` I had to create another subnet within a different `AvailabilityZone` than the `public_subnet` that Chuck originally had. I don't understand the need for the 2nd subnet in a different AZ other than because I couldn't create the `AWS::RDS::DBSubnetGroup`  without it.

TODO: 
- [x] Figure out how to expose `DOCKER_HOST` as an env var in Refinery automatically
- [x] Figure out if theres a real necessity for the extra subnet or if I'm doing some weird things
- [x] Figure out what other resources (if any) should utilize `PUBLIC_SUBNET_ID_B`
- [x] Make two private subnets and one will just be a spare for now
- [x] Create `AWS::RDS::DBSubnetGroup` in TF; Send over one subnet ID to CF
- [x] DOCKER_HOST variable in wsgi_aws from override yaml so apache user has access


deployment/terraform/live/main.tf
```diff
@@ -34,24 +34,24 @@ module "docker_host" {
...
+//  vpc_id              = "${module.vpc.vpc_id}"
+//  security_group_name = "${terraform.workspace}-refinery"
+//  refinery_host_count = "${var.refinery_host_count}"
+//  key_name            = "${var.key_name}"
+//  docker_hostname     = "${module.docker_host.docker_hostname}"
+//}
 
 module "vpc" {
   source             = "../modules/vpc"
   name               = "${var.name}"
```
hackdna: Replace ${var.name} with ${terraform.workspace}

deployment/terraform/live/vars.tf
```diff
@@ -30,21 +30,26 @@ variable "vpc_cidr_block" {
...
 
-variable "public_cidr_block" {
+variable "public_cidr_block_a" {
   type    = "string"
-  default = "10.0.0.0/25"
+  default = "10.0.0.0/26"
 }
 
-variable "private_cidr_block" {
+variable "public_cidr_block_b" {
```
mccalluc: I thought you needed two private subnets in separate AZs for the RDS, not two public ones?

scottx611x: Technically its the former or the latter, but in our case it makes more sense to have the second private one

scottx611x:
@hackdna @mccalluc I've updated this PR to just take care of the VPC creation & passing the proper info down the line to get our assets still created by CloudFormation into the VPC. 

 I've just made: https://github.com/refinery-platform/refinery-platform/tree/scottx611x%2Fec2_for_docker for the docker EC2 specifics

deployment/terraform/live/vars.tf
```diff
@@ -12,7 +12,7 @@ variable "availability_zone_b" {
 }
 
 variable "identity_pool_name" {
-  description = "Cognito federated identity pool name"
+  description = "Cognito federated identity pool name. Constraints: must be unique within AWS account, [\\w]+"
```
hackdna: Identity pool names do not have to be unique within the AWS account (though it might be useful for easier pool management). Identity pool ID is used for connecting to the correct pool.

deployment/terraform/live/main.tf
```diff
@@ -15,7 +15,7 @@ provider "aws" {
 module "object_storage" {
   source           = "../modules/s3"
   name             = "${terraform.workspace} object_storage"
-  bucket_name_base = "${terraform.workspace}"
+  bucket_name_base = "${replace(terraform.workspace, "/[^A-Za-z0-9]/", "-")}"
```
hackdna: Illegal names should produce an error and not be silently converted.

deployment/terraform/modules/rds/main.tf
```diff
@@ -1,5 +1,5 @@
 resource "aws_db_subnet_group" "db_subnet_group" {
-  name       = "rds subnet group"
+  name       = "${var.name}"
```
hackdna: Why is name necessary?

deployment/terraform/live/main.tf
mccalluc: I've hunted for a little bit, but I'm not finding a way to constrain user-supplied variables with terraform. If we remove this and just let the error pop up downstream, it's a little hard to debug: it doesn't echo the value which fails, or tell you where it came from. Also, different parts of AWS have different constraints around naming: Forcing the least common denominator across the board is not necessary.

deployment/terraform/modules/rds/main.tf
mccalluc: Without this, I was getting an error about the name already being in use. I guess it should have recognized that the resource it needed already existed, but it didn't work that way for me... perhaps there was some other aspect of my environment which was weird, but if you're pretty confident this is unnecessary, we can revert.

scottx611x:
Planning on merging this today, to be able to startup a manual testing stack. @hackdna Do you have any qualms? If theres questions or concerns related to things not directly affecting infrastructure can they be revisited at a later time?

deployment/aws-config/config.yaml.erb
```diff
@@ -0,0 +1,95 @@
...
+
+# == Not changeable ==
+
+# These things that might become configurable later.
+
+# The AMI image for the EC2 instance
+# FIXED to "ami-d05e75b8"
+
+# The instance type for the EC2 instance
+# FIXED to "m3.medium"
```
hackdna: This file appears to duplicate `config.yaml.template`. Why is that necessary?

deployment/terraform/modules/s3/outputs.tf
```diff
@@ -1,3 +1,7 @@
 output "upload_bucket_name" {
   value = "${aws_s3_bucket.uploaded_files.id}"
 }
+
+output "bucket_name_base" {
```
hackdna: `upload_bucket_name` is derived from `bucket_name_base`, so having both is redundant

deployment/terraform/modules/s3/vars.tf
```diff
@@ -2,3 +2,4 @@
 variable "bucket_name_base" {
   description = "Prefix for the Refinery Platform S3 buckets"
 }
+variable "name" {}
```
hackdna: Name parameter is only needed when creating multiple instances of a module which does not appear to be the case for any of the modules https://www.terraform.io/docs/modules/usage.html#multiple-instances-of-a-module

deployment/terraform/modules/vpc/main.tf
```diff
@@ -0,0 +1,99 @@
...
+  vpc_id            = "${aws_vpc.vpc.id}"
+  cidr_block        = "${var.private_cidr_block_a}"
+  availability_zone = "${var.availability_zone_a}"
+
+  tags {
+    Name = "${terraform.workspace} (terraform private subnet a)"
+  }
+}
+
+// This subnet currently isn't utilized other than to satisfy the
```
hackdna: Single line comments start with #
Multi-line comments are wrapped with /* and */
https://www.terraform.io/docs/configuration/syntax.html#terraform-syntax

deployment/terraform/live/main.tf
hackdna: This expression does not cover all the cases anyway. It is much easier and more informative to describe these constraints for the workspace name in the docs, or if `bucket_name_base` is an input variable then in its `description` parameter.

deployment/terraform/modules/rds/main.tf
hackdna: I have never had this error. I suspect there was something wrong in your environment.

deployment/aws-config/config.yaml.erb
mccalluc: I'll get of the template then. The erb has the advantage of letting the output from terraform be filled in automatically.

deployment/terraform/modules/s3/outputs.tf
mccalluc: There's nothing unusual about deriving a value and storing it under a new name. I'm not sure what your concern here is.

in s3/main.tf:
```
bucket = "${var.bucket_name_base}-upload"
```

s3/outputs.tf:
```
output "upload_bucket_name" {
  value = "${aws_s3_bucket.uploaded_files.id}"
}
```

cognito/main.tf:
```
      "Resource": "arn:aws:s3:::${var.upload_bucket_name}/$${cognito-identity.amazonaws.com:sub}/*"
```

Remembering that  the way we form the name is to append "-upload", and doing that in two widely separated places does not feel like a better alternative. Also: If we stop creating that bucket for some reason, the downstream references will also break when terraform tries to construct the plan, rather than just relying on naming conventions being consistent.

deployment/terraform/modules/vpc/main.tf
mccalluc: will fix

deployment/terraform/modules/s3/vars.tf
mccalluc: got rid of it and reran apply and didn't have any problems. Pushed to branch.

scottx611x:
@hackdna How are things looking now?

deployment/terraform/modules/s3/outputs.tf
hackdna: `S3_BUCKET_NAME_BASE` is set in `config.yaml` then it is used as the name of the Terraform workspace and finally as an input parameter to the S3 module (`bucket_name_base`). There is no need to output this value to include back in `config.yaml` since it already exists there.

scottx611x: I think that when we looked at this we thought it was odd that we had to set `S3_BUCKET_NAME_BASE` in config.yaml and then be sure to provide the same value as the terraform workspace name. 

Wouldn't it make sense to set this value once upon workspace creation and have `S3_BUCKET_NAME_BASE` be updated automatically?

hackdna: There is no need to automate creation of `config.yaml` because it is going away after we switch the rest of the infrastructure from CloudFormation to Terraform.

deployment/terraform/live/main.tf
```diff
@@ -9,18 +9,35 @@ terraform {
...
 module "object_storage" {
   source           = "../modules/s3"
-  bucket_name_base = "${terraform.workspace}"
+  bucket_name_base = "${replace(terraform.workspace, "/[^A-Za-z0-9]/", "-")}"
 }
 
 module "identity_pool" {
   source                   = "../modules/cognito"
-  identity_pool_name       = "${var.identity_pool_name}"
+  name                     = "${terraform.workspace} identity pool"
```
hackdna: Name parameter is only needed when creating multiple instances of a module which does not appear to be the case for this module https://www.terraform.io/docs/modules/usage.html#multiple-instances-of-a-module

deployment/terraform/live/main.tf
```diff
@@ -9,18 +9,35 @@ terraform {
...
+  identity_pool_name       = "${replace(terraform.workspace, "/[^A-Za-z0-9]/", "")}"
   upload_bucket_name       = "${module.object_storage.upload_bucket_name}"
   iam_resource_name_prefix = "${terraform.workspace}"
 }
+
+module "vpc" {
+  source               = "../modules/vpc"
+  name                 = "${terraform.workspace} vpc"
```
hackdna: Name parameter is only needed when creating multiple instances of a module which does not appear to be the case for this module https://www.terraform.io/docs/modules/usage.html#multiple-instances-of-a-module

deployment/terraform/live/main.tf
```diff
@@ -9,18 +9,35 @@ terraform {
...
+  vpc_cidr_block       = "${var.vpc_cidr_block}"
+  public_cidr_block    = "${var.public_cidr_block}"
+  private_cidr_block_a = "${var.private_cidr_block_a}"
+  private_cidr_block_b = "${var.private_cidr_block_b}"
+  availability_zone    = "${var.availability_zone_a}"
+}
+
+module "rds" {
+  source           = "../modules/rds"
+  name             = "${terraform.workspace} rds"
```
hackdna: Name parameter is only needed when creating multiple instances of a module which does not appear to be the case for this module https://www.terraform.io/docs/modules/usage.html#multiple-instances-of-a-module

deployment/terraform/live/vars.tf
```diff
@@ -1,7 +1,43 @@
...
+}
+
+variable "private_cidr_block_a" {
+  type    = "string"
+  default = "10.0.0.64/26"
+}
+
+variable "private_cidr_block_b" {
+  type    = "string"
+  default = "10.0.0.128/25"
```
hackdna: Why is this value different from `private_cidr_block_a`?

deployment/terraform/modules/vpc/main.tf
```diff
@@ -0,0 +1,101 @@
...
+}
+
+resource "aws_route_table_association" "public_subnet" {
+  subnet_id      = "${aws_subnet.public_subnet.id}"
+  route_table_id = "${aws_route_table.route_table.id}"
+}
+
+############
+
+resource "aws_eip" "nat" {
```
hackdna: Why is an elastic IP necessary?

deployment/terraform/modules/vpc/main.tf
```diff
@@ -0,0 +1,101 @@
...
+  vpc = true
+}
+
+resource "aws_nat_gateway" "nat" {
```
hackdna: Why is a NAT gateway necessary if there is an Internet gateway defined?

deployment/terraform/live/vars.tf
scottx611x: These two CIDR blocks split the remaining ip address space in half, lending each half to their respective subnet.
I guess they could both have the same cidr block but it made sense to have some distinction.

scottx611x: we'll fix

deployment/terraform/modules/s3/outputs.tf
scottx611x: is this blocking?

deployment/terraform/live/vars.tf
hackdna: Sorry, I was referring to `/25` vs `/26' specifically: why are the private subnets sized differently?

deployment/terraform/modules/vpc/main.tf
scottx611x: There might be other ways, is this a blocking issue?

deployment/terraform/modules/vpc/main.tf
scottx611x: There might be other ways, is this a blocking issue?

deployment/terraform/modules/vpc/main.tf
hackdna: This is an unnecessary resource.

deployment/terraform/modules/vpc/main.tf
hackdna: This is an unnecessary resource.

deployment/terraform/modules/s3/outputs.tf
mccalluc: It's easier for me at least to think about this if information can flow in one direction, rather that starting in config.yaml, then terraform, then back to config.yaml.

I've run this a couple times, and I feel much more confident in procedures which minimize manual intervention. If we can throw out the ERB in a week or two, great.

The goal we're aiming for is to have it start (and end) with terraform, so this feels like the right direction.

If this is a blocking issue for you, please tell precisely what you want reverted so we can just do it.

deployment/terraform/live/main.tf
mccalluc: pushed fix

deployment/terraform/live/main.tf
mccalluc: pushed fix

deployment/terraform/live/main.tf
mccalluc: pushed fix

deployment/terraform/live/vars.tf
mccalluc: It was a typo: Pushed fix. (Changed the order: now the two smaller private cidrs are the lower ranges: I recognize 64 and 128 immediately, but I don't recognize 64+128)

This is helpful: http://www.davidc.net/sites/default/subnets/subnets.html?network=10.0.0.0&mask=24&division=5.30

deployment/terraform/modules/vpc/main.tf
mccalluc: My sense is that for the private subnets to access the outside world, they do need to go through an NAT: I don't think the regular gateway is sufficient.

I'm less clear if the EIP is necessary.

If we change these, it won't be enough just to make sure apply works: We'll need to bring up a full stack, import visualizations, and run then on this branch, to make sure it all works the same.

hackdna: Yes, NAT gateway is needed for Internet access in private subnets. However, if there is no separate Docker host in this PR then there is no need for the gateway and most certainly no need for the EIP.

scottx611x: @hackdna We've realized the docker host portion has been pulled out [here](https://github.com/refinery-platform/refinery-platform/tree/scottx611x%2Fec2_for_docker) and have since removed these resources.

hackdna:
Need to either update or get rid of `live/terraform.tfvars.template`

deployment/terraform/live/main.tf
```diff
@@ -9,18 +9,32 @@ terraform {
...
 module "object_storage" {
   source           = "../modules/s3"
-  bucket_name_base = "${terraform.workspace}"
+  bucket_name_base = "${replace(terraform.workspace, "/[^A-Za-z0-9]/", "-")}"
 }
 
 module "identity_pool" {
   source                   = "../modules/cognito"
-  identity_pool_name       = "${var.identity_pool_name}"
+  identity_pool_name       = "${replace(terraform.workspace, "/[^A-Za-z0-9]/", "")}"
```
hackdna: Not necessary to sanitize: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-identitypool.html#cfn-cognito-identitypool-identitypoolname

deployment/terraform/modules/s3/outputs.tf
hackdna: OK. I've created PR #2683 to show the proposed changes.

deployment/terraform/live/main.tf
```diff
@@ -9,18 +9,32 @@ terraform {
...
 module "object_storage" {
   source           = "../modules/s3"
-  bucket_name_base = "${terraform.workspace}"
+  bucket_name_base = "${replace(terraform.workspace, "/[^A-Za-z0-9]/", "-")}"
 }
 
 module "identity_pool" {
   source                   = "../modules/cognito"
-  identity_pool_name       = "${var.identity_pool_name}"
+  identity_pool_name       = "${terraform.workspace}"
```
hackdna: Actually, it turns out that "Pool names can only contain numbers, letters, underscores and spaces", so this should be: `"${replace(terraform.workspace, "/[^A-Za-z0-9_ ]/", " ")}"`

scottx611x:
@hackdna Are you ok with these changes after #2683 is merged?

deployment/terraform/modules/vpc/main.tf
```diff
@@ -0,0 +1,73 @@
...
+/*
+This subnet currently isn't utilized other than to satisfy the
+requirements of module.rds.aws_db_subnet_group. It will be utilized once
+we have a multi-AZ RDS instance
+*/
+resource "aws_subnet" "private_subnet_b" {
+  vpc_id                  = "${aws_vpc.vpc.id}"
+  cidr_block              = "${var.private_cidr_block_b}"
+  availability_zone       = "${var.availability_zone_b}"
+  map_public_ip_on_launch = true
```
hackdna: Does this private subnet need this option?

scottx611x: No it doesn't

PR #2657, opened 2018-03-09
Hackdna/delete files

hackdna:
Ensure file is deleted on FileStoreItem instance updates when FileField is cleared

### DIFF
--- a/refinery/file_store/models.py
+++ b/refinery/file_store/models.py
@@ -348,7 +348,7 @@ class FileStoreItem(models.Model):
         except ValueError:
             if core.utils.is_absolute_url(self.source):
                 if self.source.startswith('s3://'):
-                    return None
+                    return None  # file is in the UPLOAD_BUCKET
                 return self.source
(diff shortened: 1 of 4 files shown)


refinery/file_store/models.py
```diff
@@ -169,17 +169,34 @@ def __unicode__(self):
...
+                old_instance = FileStoreItem.objects.get(pk=self.pk)
+            except FileStoreItem.DoesNotExist:
+                pass  # this is a newly created instance
+            except FileStoreItem.MultipleObjectsReturned as exc:
+                logger.critical(
+                    "Error retrieving FileStoreItem with primary key '%s': %s",
+                    self.pk, exc
+                )
+            else:
+                old_instance.delete_datafile(save_instance=False)
```
scottx611x: This is tripping me up. Why are we worried about deleting a FileStoreItem's datafile here if we enter the else on 187 only if there isn't a datafile associated?

PR #2663, opened 2018-03-12
Upgrade django-docker

mccalluc:
I think it might be better to leave this out of this release: The real problem was in the visualization itself, and that should be fixed now, and it's always possible that this introduces some other new problem.


scottx611x:
Closing. There's newer `django_docker_engine` versions that we'll use moving forward

PR #2664, opened 2018-03-12
Hackdna/remove create task

hackdna:
This task was never used as a Celery task and serves no real purpose.

### DIFF
--- a/refinery/file_store/models.py
+++ b/refinery/file_store/models.py
@@ -17,7 +17,7 @@ import urlparse
 
 from django.conf import settings
 from django.db import models
-from django.db.models.signals import post_delete, post_save
+from django.db.models.signals import post_delete
 from django.dispatch import receiver
 
 from celery.result import AsyncResult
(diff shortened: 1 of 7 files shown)


refinery/file_store/models.py
```diff
@@ -399,6 +390,19 @@ def get_file_object(file_name):
...
+def _set_filetype(sender, instance, created, **kwargs):
+    """Set file type using file extension if no file type provided"""
+    if created and not instance.filetype:
+        try:
+            extension = instance.get_file_extension()
+        except RuntimeError as exc:
+            logger.warn("Could not assign type to file '%s': %s",
+                        instance, exc)
+        else:
+            instance.filetype = extension.filetype
```
scottx611x: Wouldn't this instance need to `save()` again after updating this field? Or rather `update()` as to not trigger the signal again.

Maybe this is a better fit for the `pre_save` signal?

PR #2665, opened 2018-03-13
Jkmarx/solr perms

jkmarx:
- Remove unused solr api
- Update config.yaml with current user files column
- Remove unused visualize/genome (igv legacy)
- Resolves #2665 with reverting the JsonResponse


refinery/core/views.py
```diff
@@ -549,51 +519,9 @@ def solr_core_search(request):
             if annotations:
                 response['response']['annotations'] = annotation_data
 
-    return JsonResponse(response)
-
+            return JsonResponse(response)
```
mccalluc: Is this intentional? Unless `allIds or annotations` and `response_ids.status_code == 200` it should return `None`

mccalluc: oh, sorry, just scrolled down and saw the `HTTPResponse`

PR #2672, opened 2018-03-16
Hackdna/update apache config

hackdna:
Lower HTTP timeout value and remove unnecessary values from Apache log

### DIFF
--- a/deployment/refinery-modules/refinery/manifests/apache2.pp
+++ b/deployment/refinery-modules/refinery/manifests/apache2.pp
@@ -1,7 +1,9 @@
 class refinery::apache2 {
 
   # must be set to the ELB IdleTimeout value in the CloudFormation template
-  $timeout = 60  # seconds
+  # this is higher than the default of 60 to accommodate long synchronous
+  # processing of metadata during data set import
+  $timeout = 180  # seconds
 
   class { '::apache':
     default_mods           => false,  # to allow declaration of ::apache::mod::reqtimeout below
--- a/deployment/refinery-modules/refinery/manifests/apache2.pp
+++ b/deployment/refinery-modules/refinery/manifests/apache2.pp
@@ -40,6 +40,11 @@ class refinery::apache2 {
     content => "AcceptFilter http none\nAcceptFilter https none",
   }
 
+  # log Django messages only
+  apache::custom_config { 'error-log-format':
+    content => 'ErrorLogFormat "%M"',
+  }
+
   if $::tls_rewrite == 'true' {
     $rewrites = [
       {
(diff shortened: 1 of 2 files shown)


deployment/refinery-modules/refinery/manifests/apache2.pp
```diff
@@ -1,9 +1,9 @@
 class refinery::apache2 {
 
-  # must be set to the ELB Idle Timeout value provided in the CloudFormation template
-  # for chunked uploads of large files (MD5 calculation on the server can take a long time)
-  # to be adjusted to a more reasonable value after enabling file uploads to S3
-  $timeout = 1800
+  # must be set to the ELB IdleTimeout value in the CloudFormation template
+  # this is higher than the default of 60 to accommodate long synchronous
+  # processing of metadata during data set import
+  $timeout = 180  # seconds
```
mccalluc: How did you pick this number? If the time increases with larger files, how big a file will typically work in 180s?

(If you're tweaking the comment, could you capitalize the first letter and add a period at the end?)

deployment/refinery-modules/refinery/manifests/apache2.pp
```diff
@@ -40,6 +40,11 @@
     content => "AcceptFilter http none\nAcceptFilter https none",
   }
 
+  # log Django messages only
+  apache::custom_config { 'error-log-format':
+    content => 'ErrorLogFormat "%M"',
+  }
+
```
mccalluc: *Only* the error messages? Time and perhaps some request info could be useful...

deployment/refinery-modules/refinery/manifests/apache2.pp
hackdna: 180 seconds was more than enough to import each of the several randomly selected ISA-archives. In any case, HTTP requests should not take this long, so if there is more time required to process metadata then this is a larger problem that should be addressed in the metadata processing code.

[Puppet style guide](https://puppet.com/docs/puppet/5.4/style_guide.html#comments) does not address capitalization or punctuation, so I think this is acceptable.

deployment/refinery-modules/refinery/manifests/apache2.pp
hackdna: Django error messages contain timestamps. Request info goes into the access log.

deployment/refinery-modules/refinery/manifests/apache2.pp
scottx611x: I thought we bumped this value initially to handle large file uploads?

hackdna: Large file uploads are sent directly to S3.

PR #2677, opened 2018-03-18
Scottx611x/provide vis tools with url to their input data

scottx611x:
Introduce detail route to `/api/v2/tools/` that returns a launched Tools input data as JSON. This approach (sending a url pointing to a Tool's input data) will be used moving forward for its ease of use w/ a 2nd host running docker. Before, we noticed that there were underlying ssh commands in `django_docker_engine` to create directories on said host to place a launched `VisualizationTool's` input data, and ssh key management between the Refinery EC2 and Docker EC2 was troublesome to automate. This will also prove reusable if we ever need to move to something like ECS or Fargate.

TODO:
- [x] Update existing VisualizationTools to read input data from a url
  - [refinery-developer-vis-tool](https://github.com/refinery-platform/refinery-developer-vis-tool) has been [updated to do this](https://github.com/refinery-platform/refinery-developer-vis-tool/blob/master/context/on_startup.py#L6-L12) so far

**Note:** This shouldn't be merged until we have the 2nd host for docker up and running w/ terraform. See #2597 EDIT: We have the 2nd host working more or less and are just making some final adjustments

### DIFF
--- a/refinery/tool_manager/models.py
+++ b/refinery/tool_manager/models.py
@@ -330,7 +330,8 @@ class Tool(OwnableResource):
 
     @property
     def container_input_json_url(self):
-        """Return the url that will expose a Tool's input data (as JSON) on
+        """
+        Return the url that will expose a Tool's input data (as JSON) on
         GET requests
         """
         return urljoin(
--- a/refinery/tool_manager/models.py
+++ b/refinery/tool_manager/models.py
@@ -17,8 +17,9 @@ from bioblend.galaxy.dataset_collections import (CollectionDescription,
 from constants import UUID_RE
 from django_docker_engine.container_managers.docker_engine import \
     ExpectedPortMissing
-from django_docker_engine.docker_utils import (DockerClientWrapper,
-                                               DockerContainerSpec)
+from django_docker_engine.docker_utils import (DockerClientRunWrapper,
+                                               DockerContainerSpec,
+                                               DockerClientSpec)
 from django_extensions.db.fields import UUIDField
 from docker.errors import APIError, NotFound
 
@@ -29,6 +30,7 @@ from analysis_manager.utils import create_analysis, validate_analysis_config
 from core.models import (INPUT_CONNECTION, OUTPUT_CONNECTION, Analysis,
                          AnalysisNodeConnection, DataSet, OwnableResource,
                          Workflow)
+from core.utils import get_absolute_url
 from data_set_manager.models import Node
 from data_set_manager.utils import (get_file_url_from_node_uuid,
                                     get_solr_response_json)
@@ -306,8 +308,13 @@ class Tool(OwnableResource):
         return "Tool: {}".format(self.get_tool_name())
 
     @property
-    def _django_docker_client(self):
-        return DockerClientWrapper(settings.DJANGO_DOCKER_ENGINE_DATA_DIR)
+    def django_docker_client(self):
+        return DockerClientRunWrapper(
+            DockerClientSpec(
+                settings.DJANGO_DOCKER_ENGINE_DATA_DIR,
+                input_json_url=get_absolute_url(self.container_input_json_url)
+            )
+        )
 
     @property
     def relaunch_url(self):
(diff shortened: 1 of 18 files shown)


refinery/tool_manager/models.py
```diff
@@ -488,9 +514,17 @@ def _create_container_input_dict(self):
                 self.tool_definition.get_extra_directories()
         }
 
+    def create_container_spec(self):
+        return DockerContainerSpec(
+            image_name=self.tool_definition.image_name,
+            container_name=self.container_name,
+            labels={self.uuid: ToolDefinition.VISUALIZATION},
+            extra_directories=self.tool_definition.get_extra_directories()
```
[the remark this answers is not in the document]
mccalluc: @scottx611x : Yes: This is how HiGlass gets a directory for logs that is in the host FS, instead of the limited-capacity union FS.

refinery/tool_manager/models.py
```diff
@@ -11,29 +12,37 @@
 from django.http import JsonResponse
 
 import bioblend
-from bioblend.galaxy.dataset_collections import (CollectionDescription,
-                                                 CollectionElement,
-                                                 HistoryDatasetElement)
+from bioblend.galaxy.dataset_collections import (
+    CollectionDescription, CollectionElement, HistoryDatasetElement
+)
```
mccalluc: Did you find the setting to tweak in isort?

refinery/tool_manager/models.py
```diff
@@ -556,17 +590,9 @@ def launch(self):
...
-            input=self._create_container_input_dict(),
-            extra_directories=self.tool_definition.get_extra_directories()
-        )
-
-        self._django_docker_client.run(container)
-
+        # Pulls docker image if it doesn't exist yet, and launches container
+        # asynchronously
+        start_container.delay(self)
         return JsonResponse({Tool.TOOL_URL: self.get_relative_container_url()})
```
mccalluc: Was it synchronous before? I worry a little about race conditions... are we sure it will be able to return a container_url no matter where in the launch sequence it is?

refinery/tool_manager/tests.py
```diff
@@ -309,6 +308,7 @@ def create_solr_mock_response(self, nodes):
             }
         )
 
+    @override_settings(CELERY_ALWAYS_EAGER=True)
```
mccalluc: It makes me worry a little if the tests depend on Celery...? Particularly since the actual run seems to be mocked.

refinery/tool_manager/tests.py
```diff
@@ -2865,9 +2866,7 @@ def test_relaunch_failure_tool_doesnt_exist(self):
         get_request = self.factory.get(relaunch_url)
         force_authenticate(get_request, self.user)
         get_response = self.tool_relaunch_view(get_request, uuid=tool_uuid)
-        self.assertEqual(get_response.status_code, 400)
-        self.assertIn("Couldn't retrieve VisualizationTool",
-                      get_response.content)
+        self.assertEqual(get_response.status_code, 404)
```
mccalluc: I think the underlying error message should still be coming through... but if a 404 doesn't seem appropriate, let me know?

refinery/tool_manager/views.py
```diff
@@ -180,7 +182,7 @@ class AutoRelaunchProxy(Proxy, object):
     """
     def __init__(self):
         super(AutoRelaunchProxy, self).__init__(
-            settings.DJANGO_DOCKER_ENGINE_DATA_DIR,
+            None,  # TODO: django_docker_engine code should remove need for arg
```
mccalluc: I'll look into this. Thanks!

refinery/tool_manager/models.py
scottx611x: Yeah! In an earlier PR: [see multi_line_output](https://github.com/refinery-platform/refinery-platform/blob/develop/.isort.cfg#L27)

refinery/tool_manager/models.py
scottx611x: It was before but now that we have the nice please wait page I think this is okay. 
A VisualizationTool will always be able to return its `get_relative_container_url` regardless of the container state.

Without this, one could hit a state where a large docker image hasn't be downloaded yet, and upon clicking `launch` the page would seemingly be frozen, while the pending request waits for the image download to complete.

refinery/tool_manager/tests.py
scottx611x: This setting runs Celery on a single thread so that we don't have to worry about sleeping/waiting after launching a Tool. I could probably mock the underlying `delay()` call but this seemed easier.

refinery/tool_manager/tests.py
scottx611x: Now that we use `django.shortcuts.get_object_or_404` a resonable error message is returned. Something like `No VisualizationTool matches the given query.` I was hesitant to test Django internals

PR #2683, opened 2018-03-21
Remove upload_bucket_name output from root and bucket_name_base outpu…

hackdna:
…t from S3 module


deployment/terraform/live/outputs.tf
```diff
@@ -1,7 +1,3 @@
-output "upload_bucket_name" {
-  value = "${module.object_storage.upload_bucket_name}"
-}
```
mccalluc: Looking at the config.json now, we don't reference the upload bucket: I had missed that. Thanks for cleaning this up.

scottx611x:
@hackdna Is this ok to merge?

PR #2684, opened 2018-03-21
Scottx611x/ec2 for docker

scottx611x:
This includes @mccalluc's earlier terraform specific docker host changes as well as what it would take to expose `DOCKER_HOST` env var so that Refinery can talk to the remote docker machine. 

**Note**: I can't remember if I got the second part (`DOCKER_HOST` env var) fully working and this should be tested thoroughly.

**Another Note:** Having a 2nd Docker host depends on the changes from #2677 so that the Refinery & Docker hosts don't have to talk over SSH. That PR then depends on us updating all exiting Vis Tools to read their input data from the provided `INPUT_JSON_URL`


scottx611x:
Closing this in favor of exiting PR #2597

PR #2704, opened 2018-03-28
Scottx611x/collect stats

### LINKED ISSUE #2694: Add Collection of Requested Metrics
We want to be able to gather the metrics from [this doc](https://docs.google.com/document/d/18SQl7uwail3U5rY6CVbDgdtdyf52QS_jSfPwGnp48oU/edit?usp=sharing)

**Refinery Platform Usage**
- #### Weekly
  - [x] \# Data sets Uploaded
    - [ ] Uploaded by Core Consultants
    - [ ] Uploaded by Other Users
  - [x] \# Data sets shared
  - [x] \# Users Created
  - [x] \# Groups Created
  - [x] \# Unique User Logins
  - [x] \# Total User Logins
  - [x] \# Total Workflow Launches
  - [x] \# Total Visualization Launches

scottx611x:
Resolves #2694 

Things to note:
- `ExtendedGroup` creation currently has a `post_save` signal handler that runs `core.models.create_manager_group` so the `groups_created` metric will always be 2x (I don't think this is what we want)
  - [x] Exclude `manager` groups
- We currently have no notion of a `User` being a `Core Consultant`
  - This doesn't matter for now
- There are probably some edge cases when/if `Users` get deleted because this could skew things like: `total_user_logins`
- [x] Run this script daily
- The collected information is currently available through the admin ui, but I'm not sure if anything more is expected. Maybe an admin_only view?
  - Maybe in the future, numbers are fine for now

### DIFF
--- a/refinery/config/settings/base.py
+++ b/refinery/config/settings/base.py
@@ -323,7 +323,6 @@ EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 CELERYD_MAX_TASKS_PER_CHILD = get_setting("CELERYD_MAX_TASKS_PER_CHILD")
 CELERY_ROUTES = {"file_store.tasks.import_file": {"queue": "file_import"}}
 CELERY_ACCEPT_CONTENT = ['pickle']
-# TODO: Does this belong here or in config.json.erb?
 CELERYBEAT_SCHEDULE = {
     'collect_site_statistics': {
         'task': 'core.tasks.collect_site_statistics',
--- a/refinery/config/settings/base.py
+++ b/refinery/config/settings/base.py
@@ -325,6 +325,13 @@ CELERY_ROUTES = {"file_store.tasks.import_file": {"queue": "file_import"}}
 CELERY_ACCEPT_CONTENT = ['pickle']
 # TODO: Does this belong here or in config.json.erb?
 CELERYBEAT_SCHEDULE = {
+    'collect_site_statistics': {
+        'task': 'core.tasks.collect_site_statistics',
+        'schedule': timedelta(weeks=1),
+        'options': {
+            'expires': 30,  # seconds
+        }
+    },
     'django_docker_cleanup': {
         'task': 'tool_manager.tasks.django_docker_cleanup',
         'schedule': timedelta(seconds=30),
--- a/refinery/config/settings/base.py
+++ b/refinery/config/settings/base.py
@@ -327,7 +327,7 @@ CELERY_ACCEPT_CONTENT = ['pickle']
 CELERYBEAT_SCHEDULE = {
     'collect_site_statistics': {
         'task': 'core.tasks.collect_site_statistics',
-        'schedule': timedelta(weeks=1),
+        'schedule': timedelta(days=1),
         'options': {
             'expires': 30,  # seconds
         }
--- a/refinery/core/admin.py
+++ b/refinery/core/admin.py
@@ -9,10 +9,14 @@ from django.contrib import admin
 from django_extensions.admin import ForeignKeyAutocompleteAdmin
 from guardian.admin import GuardedModelAdmin
 
-from .models import (Analysis, AnalysisNodeConnection, AnalysisResult, DataSet,
-                     Download, ExtendedGroup, InvestigationLink, Invitation,
-                     Ontology, Project, SiteProfile, Tutorials, UserProfile,
-                     Workflow, WorkflowEngine)
+from tool_manager.utils import AdminFieldPopulator
+
+from .models import (
+    Analysis, AnalysisNodeConnection, AnalysisResult, DataSet, Download,
+    ExtendedGroup, InvestigationLink, Invitation, Ontology, Project,
+    SiteProfile, SiteStatistics, Tutorials, UserProfile, Workflow,
+    WorkflowEngine
+)
 from .utils import admin_ui_deletion
 
 
@@ -147,6 +151,10 @@ class SiteProfileAdmin(GuardedModelAdmin):
     list_display = ['__unicode__', 'site']
 
 
+class SiteStatisticsAdmin(AdminFieldPopulator):
+    pass
+
+
 admin.site.register(ExtendedGroup, ExtendedGroupAdmin)
 admin.site.register(Project, ProjectAdmin)
 admin.site.register(DataSet, DataSetAdmin)
@@ -162,3 +170,4 @@ admin.site.register(UserProfile, UserProfileAdmin)
 admin.site.register(Tutorials, TutorialsAdmin)
 admin.site.register(Ontology, OntologyAdmin)
 admin.site.register(SiteProfile, SiteProfileAdmin)
+admin.site.register(SiteStatistics, SiteStatisticsAdmin)
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -784,6 +784,10 @@ class DataSet(SharableResource):
     def is_isatab_based(self):
         return bool(self.get_investigation().isa_archive)
 
+    @property
+    def shared(self):  # is_shared breaks a tastypie interal
+        return len(self.get_groups()) > 0
+
 
 @receiver(pre_delete, sender=DataSet)
 def _dataset_delete(sender, instance, *args, **kwargs):
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -2243,10 +2243,10 @@ class SiteStatistics(models.Model):
     def collect(self):
         self.datasets_uploaded = self._get_datasets_uploaded()
         self.datasets_shared = self._get_datasets_shared()
-        self.groups_created = self._get_number_of_groups_created()
+        self.groups_created = self._get_groups_created()
         self.users_created = self._get_users_created()
         self.unique_user_logins = self._get_unique_user_logins()
-        self.total_user_logins = self._get_number_of_total_user_logins()
+        self.total_user_logins = self._get_total_user_logins()
         self.total_workflow_launches = self._get_total_workflow_launches()
         self.total_visualization_launches = \
             self._get_total_visualization_launches()
@@ -2293,11 +2293,14 @@ class SiteStatistics(models.Model):
             date_joined__gte=self._get_previous_instance().run_date
         ).count()
 
-    def _get_number_of_groups_created(self):
+    def _get_groups_created(self):
         return ExtendedGroup.objects.exclude(manager_group=None).count() - sum(
-            [s.groups_created for s in SiteStatistics.objects.all()]
+            s.groups_created for s in SiteStatistics.objects.exclude(
+                id=self.id
+            )
         )
 
-    def _get_number_of_total_user_logins(self):
-        return sum([u.login_count for u in UserProfile.objects.all()]) - \
-               sum([s.total_user_logins for s in SiteStatistics.objects.all()])
+    def _get_total_user_logins(self):
+        return sum(u.login_count for u in UserProfile.objects.all()) - \
+               sum(s.total_user_logins for s in
+                   SiteStatistics.objects.exclude(id=self.id))
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -38,30 +38,33 @@ from bioblend import galaxy
 from django_auth_ldap.backend import LDAPBackend
 from django_extensions.db.fields import UUIDField
 from guardian.models import UserObjectPermission
-from guardian.shortcuts import (assign_perm, get_groups_with_perms,
-                                get_objects_for_group, get_users_with_perms,
-                                remove_perm)
+from guardian.shortcuts import (
+    assign_perm, get_groups_with_perms, get_objects_for_group,
+    get_users_with_perms, remove_perm
+)
 import pysolr
 from registration.models import RegistrationManager, RegistrationProfile
 from registration.signals import user_activated, user_registered
 
-from data_set_manager.models import (Assay, Investigation, Node,
-                                     NodeCollection, Study)
+from data_set_manager.models import (
+    Assay, Investigation, Node, NodeCollection, Study
+)
 from data_set_manager.search_indexes import NodeIndex
-from data_set_manager.utils import (add_annotated_nodes_selection,
-                                    index_annotated_nodes_selection)
+from data_set_manager.utils import (
+    add_annotated_nodes_selection, index_annotated_nodes_selection
+)
 from file_store.models import FileStoreItem, FileType
 from file_store.tasks import rename
 from galaxy_connector.models import Instance
 import tool_manager
 
-from .utils import (add_or_update_user_to_neo4j, add_read_access_in_neo4j,
-                    async_update_annotation_sets_neo4j, delete_data_set_index,
-                    delete_data_set_neo4j, delete_ontology_from_neo4j,
-                    delete_user_in_neo4j, email_admin,
-                    invalidate_cached_object, remove_read_access_in_neo4j,
-                    skip_if_test_run, sync_update_annotation_sets_neo4j,
-                    update_data_set_index)
+from .utils import (
+    add_or_update_user_to_neo4j, add_read_access_in_neo4j,
+    async_update_annotation_sets_neo4j, delete_data_set_index,
+    delete_data_set_neo4j, delete_ontology_from_neo4j, delete_user_in_neo4j,
+    email_admin, invalidate_cached_object, remove_read_access_in_neo4j,
+    skip_if_test_run, sync_update_annotation_sets_neo4j, update_data_set_index
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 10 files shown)


refinery/core/admin.py
```diff
@@ -9,10 +9,14 @@
...
-                     Ontology, Project, SiteProfile, Tutorials, UserProfile,
-                     Workflow, WorkflowEngine)
+from tool_manager.utils import AdminFieldPopulator
+
+from .models import (
+    Analysis, AnalysisNodeConnection, AnalysisResult, DataSet, Download,
+    ExtendedGroup, InvestigationLink, Invitation, Ontology, Project,
+    SiteProfile, SiteStatistics, Tutorials, UserProfile, Workflow,
+    WorkflowEngine
+)
```
hackdna: Unnecessary line breaks

refinery/core/tests.py
```diff
@@ -18,35 +20,41 @@
 from guardian.shortcuts import assign_perm, get_objects_for_group, remove_perm
 import mock
 import mockcache as memcache
-from rest_framework.test import (APIClient, APIRequestFactory, APITestCase,
-                                 force_authenticate)
+from rest_framework.test import (
+    APIClient, APIRequestFactory, APITestCase, force_authenticate
+)
```
hackdna: Unnecessary line break

refinery/core/tests.py
```diff
@@ -18,35 +20,41 @@
...
-                                     Study)
-from factory_boy.django_model_factories import (GalaxyInstanceFactory,
-                                                WorkflowEngineFactory,
-                                                WorkflowFactory)
-from factory_boy.utils import (create_dataset_with_necessary_models,
-                               make_analyses_with_single_dataset)
+from core.tasks import collect_site_statistics
+from data_set_manager.models import (
+    AnnotatedNode, Assay, Contact, Investigation, Node, NodeCollection, Study
+)
```
hackdna: Unnecessary reformatting

refinery/core/tests.py
```diff
@@ -18,35 +20,41 @@
...
+from factory_boy.django_model_factories import (
+    GalaxyInstanceFactory, WorkflowEngineFactory, WorkflowFactory
+)
+from factory_boy.utils import (
+    create_dataset_with_necessary_models, create_tool_with_necessary_models,
+    make_analyses_with_single_dataset
+)
```
hackdna: Unnecessary line break

refinery/core/tests.py
```diff
@@ -18,35 +20,41 @@
...
-                     AnalysisNodeConnection, AnalysisResult, BaseResource,
-                     DataSet, ExtendedGroup, InvestigationLink, Project,
-                     Tutorials, UserProfile, Workflow, WorkflowEngine,
-                     invalidate_cached_object)
+from .models import (
+    INPUT_CONNECTION, OUTPUT_CONNECTION, Analysis, AnalysisNodeConnection,
+    AnalysisResult, BaseResource, DataSet, ExtendedGroup, InvestigationLink,
+    Project, SiteStatistics, Tutorials, UserProfile, Workflow, WorkflowEngine,
+    invalidate_cached_object
+)
```
hackdna: Unnecessary line break

refinery/core/tests.py
```diff
@@ -18,35 +20,41 @@
...
 from .search_indexes import DataSetIndex
-from .utils import (filter_nodes_uuids_in_solr, get_aware_local_time,
-                    get_resources_for_user, move_obj_to_front,
-                    which_default_read_perm)
+from .utils import (
+    filter_nodes_uuids_in_solr, get_aware_local_time, get_resources_for_user,
+    move_obj_to_front, which_default_read_perm
+)
```
hackdna: Unnecessary line break

refinery/core/tests.py
```diff
@@ -2297,3 +2305,151 @@ def test_create_workflow_engine_bad_group_name(self):
...
+
+    def test_total_visualization_launches(self):
+        self.assertEqual(
+            self.site_statistics_a.total_visualization_launches, 0)
+        self.assertEqual(
+            self.site_statistics_b.total_visualization_launches, 0)
+        self.assertEqual(
+            self.site_statistics_c.total_visualization_launches,
+            1
+        )
```
hackdna: Unnecessary line breaks

refinery/core/models.py
```diff
@@ -38,30 +38,33 @@
 from django_auth_ldap.backend import LDAPBackend
 from django_extensions.db.fields import UUIDField
 from guardian.models import UserObjectPermission
-from guardian.shortcuts import (assign_perm, get_groups_with_perms,
-                                get_objects_for_group, get_users_with_perms,
-                                remove_perm)
+from guardian.shortcuts import (
+    assign_perm, get_groups_with_perms, get_objects_for_group,
+    get_users_with_perms, remove_perm
+)
```
hackdna: Unnecessary line breaks

refinery/core/models.py
```diff
@@ -38,30 +38,33 @@
...
 import pysolr
 from registration.models import RegistrationManager, RegistrationProfile
 from registration.signals import user_activated, user_registered
 
-from data_set_manager.models import (Assay, Investigation, Node,
-                                     NodeCollection, Study)
+from data_set_manager.models import (
+    Assay, Investigation, Node, NodeCollection, Study
+)
```
hackdna: Unnecessary line breaks

refinery/core/models.py
```diff
@@ -38,30 +38,33 @@
...
 from data_set_manager.search_indexes import NodeIndex
-from data_set_manager.utils import (add_annotated_nodes_selection,
-                                    index_annotated_nodes_selection)
+from data_set_manager.utils import (
+    add_annotated_nodes_selection, index_annotated_nodes_selection
+)
```
hackdna: Unnecessary line breaks

refinery/core/models.py
```diff
@@ -781,6 +784,10 @@ def get_node_uuids(self):
     def is_isatab_based(self):
         return bool(self.get_investigation().isa_archive)
 
+    @property
+    def shared(self):  # is_shared breaks a tastypie interal
```
hackdna: Why a property?

refinery/core/models.py
```diff
@@ -2213,3 +2220,87 @@ class SiteProfile(models.Model):
...
+    total_workflow_launches = models.IntegerField()
+    users_created = models.IntegerField()
+    unique_user_logins = models.IntegerField()
+
+    class Meta:
+        verbose_name_plural = "site statistics"
+
+    def save(self, force_insert=False, force_update=False, using=None,
+             update_fields=None):
+        self.run_date = timezone.now()
```
hackdna: Set `auto_now=True` on run_date instead?
https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.DateField.auto_now

refinery/core/models.py
```diff
@@ -2213,3 +2220,87 @@ class SiteProfile(models.Model):
...
+        # Return the previous instance by run_date, or the current instance
+        # if a prior one doesn't exist
+        return previous_instance or self
+
+    def _get_datasets_shared(self):
+        return [
+            dataset for dataset in DataSet.objects.filter(
+                modification_date__gte=self._get_previous_instance().run_date
+            ) if dataset.shared
+        ]
```
hackdna: Return just the number of data sets instead of the whole array?

refinery/core/models.py
```diff
@@ -2213,3 +2220,87 @@ class SiteProfile(models.Model):
...
+
+    def _get_datasets_uploaded(self):
+        return DataSet.objects.filter(
+            creation_date__gte=self._get_previous_instance().run_date
+        )
```
hackdna: Return just the number of data sets instead of the queryset?

refinery/core/models.py
```diff
@@ -2213,3 +2220,87 @@ class SiteProfile(models.Model):
...
+
+    def _get_total_visualization_launches(self):
+        return tool_manager.models.VisualizationTool.objects.filter(
+            creation_date__gte=self._get_previous_instance().run_date
+        )
```
hackdna: Return just the number of tools instead of the queryset?

refinery/core/models.py
```diff
@@ -2213,3 +2220,87 @@ class SiteProfile(models.Model):
...
+
+    def _get_total_workflow_launches(self):
+        return tool_manager.models.WorkflowTool.objects.filter(
+            creation_date__gte=self._get_previous_instance().run_date
+        )
```
hackdna: Return just the number of tools instead of the queryset?

refinery/core/models.py
```diff
@@ -2213,3 +2220,87 @@ class SiteProfile(models.Model):
...
+
+    def _get_unique_user_logins(self):
+        return User.objects.filter(
+            last_login__gte=self._get_previous_instance().run_date
+        )
```
hackdna: Return just the number of logins instead of the queryset?

refinery/core/models.py
```diff
@@ -2213,3 +2220,87 @@ class SiteProfile(models.Model):
...
+
+    def _get_total_workflow_launches(self):
+        return tool_manager.models.WorkflowTool.objects.filter(
+            creation_date__gte=self._get_previous_instance().run_date
+        )
+
+    def _get_unique_user_logins(self):
+        return User.objects.filter(
+            last_login__gte=self._get_previous_instance().run_date
```
hackdna: This is probably not very useful since `last_login` is only updated when a user actually logs in with a username and password.

refinery/core/models.py
```diff
@@ -2213,3 +2220,87 @@ class SiteProfile(models.Model):
...
+        )
+
+    def _get_users_created(self):
+        return User.objects.filter(
+            date_joined__gte=self._get_previous_instance().run_date
+        )
```
hackdna: Return just the number of users instead of the queryset?

refinery/core/tests.py
```diff
@@ -2297,3 +2305,151 @@ def test_create_workflow_engine_bad_group_name(self):
...
+        self.assertEqual(initial_site_statistics.unique_user_logins, 1)
+        self.assertEqual(initial_site_statistics.total_user_logins, 2)
+        self.assertEqual(initial_site_statistics.total_workflow_launches, 0)
+        self.assertEqual(
+            initial_site_statistics.total_visualization_launches, 0)
+
+
+class SiteStatisticsTests(TestCase):
+    def setUp(self):
+        public_group = ExtendedGroup.objects.public_group()
```
hackdna: Avoid relying on database migrations (#2531)

refinery/core/tests.py
```diff
@@ -2297,3 +2305,151 @@ def test_create_workflow_engine_bad_group_name(self):
...
+                'schedule': timedelta(days=1),
+                'options': {
+                    'expires': 30,  # seconds
+                }
+            }
+        )
+
+    def test_collect_site_statistics_creates_new_instance(self):
+        initial_site_statistics_count = SiteStatistics.objects.count()
+        with self.settings(CELERY_ALWAYS_EAGER=True):
```
hackdna: `CELERY_ALWAYS_EAGER` should not be necessary since `collect_site_statistics()` is called directly.

refinery/core/admin.py
scottx611x: Regarding all of your comments about unnecessary line breaks/reformatting of import statements: the new isort config does this automatically. [This configuration](https://github.com/refinery-platform/refinery-platform/blob/develop/.isort.cfg#L27) was the closest I could find that satisfied your concerns from [here](https://github.com/refinery-platform/refinery-platform/pull/2646#discussion_r174910995)

If this is a big deal then it should be brought up with the team and we should make a decision on it.

refinery/core/models.py
scottx611x: I thought `dataset.shared` read easier than `dataset.shared()`

refinery/core/models.py
scottx611x: If I set `auto_now=True` on `run_date` I wouldn't have access to the value until after `save()` was complete. I need to know the `run_date` in the methods called within `save()`

refinery/core/models.py
scottx611x: I left this method not returning a count so that it would be easier to extend in the future when we want to know about the objects created rather than just their counts

refinery/core/models.py
scottx611x: I left this method not returning a count so that it would be easier to extend in the future when we want to know about the objects created rather than just their counts

refinery/core/models.py
scottx611x: I left this method not returning a count so that it would be easier to extend in the future when we want to know about the objects created rather than just their counts

refinery/core/models.py
scottx611x: I left this method not returning a count so that it would be easier to extend in the future when we want to know about the objects created rather than just their counts

refinery/core/models.py
scottx611x: I left this method not returning a count so that it would be easier to extend in the future when we want to know about the objects created rather than just their counts

refinery/core/models.py
scottx611x: It was a metric that I was asked to collect here: https://github.com/refinery-platform/refinery-platform/issues/2694#issue-308064403

refinery/core/models.py
scottx611x: I left this method not returning a count so that it would be easier to extend in the future when we want to know about the objects created rather than just their counts

refinery/core/admin.py
hackdna: The decision to apply new formatting rules was made without consulting the team, so that's the conversation that needs to happen first.

refinery/core/models.py
hackdna: "Always implement things when you actually need them, never when you just foresee that you need them." https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

refinery/core/admin.py
jkmarx: At the last attended dev meeting, I thought we concluded we didn't care which as long as we picked one and stuck with it.

scottx611x: This is different. Ilya commented on a PR of mine [here](https://github.com/refinery-platform/refinery-platform/pull/2646#discussion_r174910995) regarding the styling of import statements. I took this as blocking for that PR, and implemented the closest [`isort` configuration](https://github.com/refinery-platform/refinery-platform/blob/develop/.isort.cfg#L27) that I could find that addressed his concerns.

hackdna: Right, there was nothing decided. I don't think applying new formatting rules is a good idea just because a tool insists on doing so. My vote is to follow existing de facto style and avoid extra line breaks.

refinery/core/tests.py
```diff
@@ -2297,3 +2305,151 @@ def test_create_workflow_engine_bad_group_name(self):
...
+        self.assertEqual(initial_site_statistics.users_created, 3)
+        self.assertEqual(initial_site_statistics.groups_created, 1)
+        self.assertEqual(initial_site_statistics.unique_user_logins, 1)
+        self.assertEqual(initial_site_statistics.total_user_logins, 2)
+        self.assertEqual(initial_site_statistics.total_workflow_launches, 0)
+        self.assertEqual(
+            initial_site_statistics.total_visualization_launches, 0)
+
+
+class SiteStatisticsTests(TestCase):
```
hackdna: This looks suspiciously like a suite of integration tests. I would simplify this by unit testing individual collection methods then write just one of two integration tests.

refinery/config/settings/base.py
```diff
@@ -325,6 +325,13 @@ def get_setting(name, settings=local_settings, default=None):
 CELERY_ACCEPT_CONTENT = ['pickle']
 # TODO: Does this belong here or in config.json.erb?
```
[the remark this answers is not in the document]
hackdna: Not user-configurable, so does not belong in config.json.

refinery/core/models.py
mccalluc: Here and below, if you're pretty sure you'll be using it that way, sure, but otherwise, it's generally good to start with just what's needed.

refinery/core/models.py
```diff
@@ -2213,3 +2220,87 @@ class SiteProfile(models.Model):
...
+
+    def _get_number_of_groups_created(self):
+        return ExtendedGroup.objects.exclude(manager_group=None).count() - sum(
+            [s.groups_created for s in SiteStatistics.objects.all()]
+        )
+
+    def _get_number_of_total_user_logins(self):
+        return sum([u.login_count for u in UserProfile.objects.all()]) - \
+               sum([s.total_user_logins for s in SiteStatistics.objects.all()])
```
mccalluc: I think you can just use a generator here and above: you don't need to make a comprehension from it:
```
sum(i*2 for i in range(3))
6
```

refinery/core/tests.py
```diff
@@ -2297,3 +2305,151 @@ def test_create_workflow_engine_bad_group_name(self):
...
+        self.assertEqual(self.site_statistics_b.users_created, 0)
+        self.assertEqual(
+            self.site_statistics_c.users_created,
+            self.site_statistics_c._get_previous_instance().users_created + 2
+        )
+
+    def test_groups_created(self):
+        self.assertEqual(self.site_statistics_a.groups_created, 0)
+        self.assertEqual(self.site_statistics_b.groups_created, 0)
+        self.assertEqual(self.site_statistics_c.groups_created, 1)
```
mccalluc: For all of these, I might compare triples: If the first assertion failed, you'd probably be interested in what the 2nd and 3rd were doing, too.

refinery/core/tests.py
```diff
@@ -2297,3 +2305,151 @@ def test_create_workflow_engine_bad_group_name(self):
...
+        self.assertEqual(
+            self.site_statistics_c.total_visualization_launches,
+            1
+        )
+
+    def test__get_previous_instance_returns_itself_if_only_instance(self):
+        SiteStatistics.objects.all().delete()
+
+        site_statistics = SiteStatistics.objects.create()
+        self.assertIsNotNone(site_statistics._get_previous_instance())
```
mccalluc: drop this?

Issue #2675, opened 2018-03-16
How should we handle Python coding style for open/close tokens spanning multiple lines?


scottx611x:
There are many ways to go about doing this, but we should configure our linter to adhere to a single style for consistency.

jkmarx:
https://community.lsst.org/t/pep8-style-hanging-indentation-and-closing-brackets/618

jkmarx:
Ref three and one (fixed } ) 
- [ ] Add to style wiki

PR #2711, opened 2018-04-05
Update Solr queries to not limit # of docs returned

scottx611x:
Currently a user can select as many nodes as they want for input to a `VisualizationTool`, but the response from Solr (which gets passed into the Tool as input) will only include, at most, 10 results.


refinery/data_set_manager/utils.py
```diff
@@ -1172,7 +1172,8 @@ def _create_solr_params_from_node_uuids(node_uuids):
     return {
         "q": "django_ct:data_set_manager.node",
         "wt": "json",
-        "fq": "uuid:({})".format(" OR ".join(node_uuids))
+        "fq": "uuid:({})".format(" OR ".join(node_uuids)),
+        "rows": 100
```
jkmarx: why limit it to 100?

jkmarx: ```The default value is "10", which is used if the parameter is not specified. If you want to tell Solr to return all possible results from the query without an upper bound, specify rows to be 10000000 or some other ridiculously large value that is higher than the possible number of rows that are expected.```

refinery/tool_manager/tests.py
```diff
@@ -3630,6 +3630,22 @@ def assertions(tool):
             assertions
         )
 
+    @override_settings(REFINERY_SOLR_DOC_LIMIT=10)
+    def test_input_node_limit(self):
+        tool = self.create_tool(ToolDefinition.VISUALIZATION)
+        tool_launch_config = self.tool.get_tool_launch_config()
+        tool_launch_config[Tool.FILE_RELATIONSHIPS] = str(
+                [uuid.uuid4()] * (10 + 1)
```
hackdna: `10 + 1 = 11`? :)

refinery/tool_manager/tests.py
```diff
@@ -3630,6 +3630,22 @@ def assertions(tool):
...
+            )
+        tool.set_tool_launch_config(tool_launch_config)
+        with self.assertRaises(VisualizationToolError) as context:
+            tool.launch()
+
+        self.assertIn(
+            "Input Node limit of: 10 reached",
+            context.exception.message
+        )
```
hackdna: This can fit on two lines instead of four

refinery/config/settings/base.py
```diff
@@ -637,3 +637,6 @@ def get_setting(name, settings=local_settings, default=None):
 }
 REFINERY_VISUALIZATION_REGISTRY = \
     "https://github.com/refinery-platform/visualization-tools/"
+
+# The upper bound of documents returned from a Solr query
+REFINERY_SOLR_DOC_LIMIT = 10000000
```
hackdna: If this is not user-configurable then shouldn't this be just a data_set_manager constant?

refinery/tool_manager/tests.py
scottx611x: Sounds like this issue again: https://github.com/refinery-platform/refinery-platform/issues/2675

refinery/tool_manager/models.py
```diff
@@ -488,6 +488,14 @@ def _create_container_input_dict(self):
                 self.tool_definition.get_extra_directories()
         }
 
+    def _check_input_node_limit(self):
+        if len(self.get_input_node_uuids()) > settings.REFINERY_SOLR_DOC_LIMIT:
```
jkmarx: Maybe a warning instead of an error?

refinery/user_files_manager/tests.py
```diff
@@ -109,6 +109,7 @@ def test_user_files_csv(self):
 
 class UserFilesUtilsTests(TestCase):
 
+    @override_settings(REFINERY_SOLR_DOC_LIMIT=10)
```
jkmarx: Can we set the default to 10000 for tests? In general, ten is the solr default and would be unnecessary to include.

refinery/core/utils.py
```diff
@@ -15,6 +15,7 @@
 from django.utils import timezone
 
 from celery.task import task
+from constants import REFINERY_SOLR_DOC_LIMIT
```
hackdna: `import constants` and move it into the "local" import section below since this is not a 3rd party module

refinery/data_set_manager/tests.py
```diff
@@ -11,49 +11,53 @@
 
 from django.conf import settings
 from django.contrib.auth.models import User
-from django.core.files.uploadedfile import (InMemoryUploadedFile,
-                                            SimpleUploadedFile)
+from django.core.files.uploadedfile import (
+    InMemoryUploadedFile, SimpleUploadedFile
+)
```
hackdna: Can we stop rewriting existing code until we agree on the formatting rules?

refinery/data_set_manager/tests.py
```diff
@@ -11,49 +11,53 @@
...
 from django.db.models import Q
 from django.http import QueryDict
 from django.test import LiveServerTestCase, TestCase
 
 from celery.states import PENDING, STARTED, SUCCESS
-from constants import NOT_AVAILABLE
+from constants import NOT_AVAILABLE, REFINERY_SOLR_DOC_LIMIT
```
hackdna: import constants and move it into the "local" import section below since this is not a 3rd party module

refinery/data_set_manager/utils.py
```diff
@@ -17,12 +17,15 @@
 from django.db.models import Q
 from django.utils.http import urlquote, urlunquote
 
+from constants import REFINERY_SOLR_DOC_LIMIT
```
hackdna: import constants and move it into the "local" import section below since this is not a 3rd party module

refinery/tool_manager/models.py
```diff
@@ -11,27 +11,32 @@
...
 import bioblend
-from bioblend.galaxy.dataset_collections import (CollectionDescription,
-                                                 CollectionElement,
-                                                 HistoryDatasetElement)
-from constants import UUID_RE
-from django_docker_engine.container_managers.docker_engine import \
+from bioblend.galaxy.dataset_collections import (
+    CollectionDescription, CollectionElement, HistoryDatasetElement
+)
+from constants import REFINERY_SOLR_DOC_LIMIT, UUID_RE
```
hackdna: import constants and move it into the "local" import section below since this is not a 3rd party module

refinery/user_files_manager/tests.py
```diff
@@ -8,11 +8,13 @@
 from django.http import QueryDict
 from django.test import RequestFactory, TestCase, override_settings
 
+from constants import REFINERY_SOLR_DOC_LIMIT
```
hackdna: `import constants` and move it into the "local" import section below since this is not a 3rd party module

refinery/data_set_manager/tests.py
```diff
@@ -25,12 +25,12 @@
 from mock import ANY
 from rest_framework.test import APIClient, APIRequestFactory, APITestCase
 
-import constants
 from core.models import (INPUT_CONNECTION, OUTPUT_CONNECTION, Analysis,
                          AnalysisNodeConnection, DataSet, ExtendedGroup,
                          InvestigationLink)
 from core.tests import TestMigrations
 from core.views import NodeViewSet
+import constants
```
mccalluc: Trying not to care, but should constants come before core alphabetically?

refinery/data_set_manager/tests.py
```diff
@@ -943,12 +943,13 @@ def test_generate_solr_params_no_params(self):
...
                          '&fq=is_annotation%3Afalse'
                          '&start=0'
-                         '&rows=10000000'
+                         '&rows={}'
                          '&q=django_ct%3Adata_set_manager.node'
                          '&wt=json'
                          '&facet=true'
                          '&facet.limit=-1'.format(
-                                 self.valid_uuid))
+                             self.valid_uuid, constants.REFINERY_SOLR_DOC_LIMIT
```
mccalluc: Reverse?

refinery/tool_manager/tests.py
```diff
@@ -3648,6 +3651,22 @@ def assertions(tool):
...
         )
 
+    def test_input_node_limit(self):
+        with mock.patch("constants.REFINERY_SOLR_DOC_LIMIT", 10):
+            tool = self.create_tool(ToolDefinition.VISUALIZATION)
+            tool_launch_config = self.tool.get_tool_launch_config()
+
+            # Crete one more entry than what REFINERY_SOLR_DOC_LIMIT permits
+            tool_launch_config[Tool.FILE_RELATIONSHIPS] = str(
+                [uuid.uuid4()] * 11
```
mccalluc: `(constants.REFINERY_SOLR_DOC_LIMIT + 1)` instead? And then the comment probably isn't needed.

refinery/data_set_manager/tests.py
scottx611x: This is what we want here (rows corresponding to the REFINERY_SOLR_DOC_LIMIT)

refinery/tool_manager/tests.py
scottx611x: I ran into trouble running this test quickly (if at all) when making an array that large. `constants.REFINERY_SOLR_DOC_LIMIT + 1` -> 10000001

I think it was probably trying to write something that big to the db

How about something like:
```
with mock.patch("constants.REFINERY_SOLR_DOC_LIMIT", 10) as solr_limit:
    tool = self.create_tool(ToolDefinition.VISUALIZATION)
    tool_launch_config = self.tool.get_tool_launch_config()

    # Crete one more entry than what REFINERY_SOLR_DOC_LIMIT permits
    tool_launch_config[Tool.FILE_RELATIONSHIPS] = str(
        [uuid.uuid4()] * (solr_limit + 1)
    )
```

hackdna: If that's the case then `REFINERY_SOLR_DOC_LIMIT` should also be set to a lower number in `constants.py` to prevent potential outages.

Issue #2359, opened 2017-11-09
Rewrite import_file task


hackdna:
Problems in the existing implementation:
* returns a value (needs to be stored)
* returned values are inconsistent in case of errors (None or exception)
* violates single responsibility principle
* is large and contains duplicated code

All this leads to issues like #1351
Proper testing depends on #2993

To do:
- [x] update import from absolute file system path
- [x] update import from the upload bucket
- [x] update import from URL
- [x] clean up temp files on revokes and failures
- [x] update to move files from upload bucket to media bucket
- [x] avoid returning a value
- [x] make sure it is not called directly or check `self.request.called_directly` to avoid `IntegrityError` when task is run synchronously (http://docs.celeryproject.org/en/3.1/userguide/tasks.html#custom-states)
- [x] report progress for file copy and download operations
- [x] report progress using built-in STARTED state instead of custom PROGRESS state (to avoid dealing with an additional state)?
- [x] test locally with tabular and ISA-archive data sets (uploaded and downloaded files) using file system storage backend/file system uploads and S3 uploads
- [x] test locally with tabular and ISA-archive data sets (uploaded and downloaded files) using S3 storage backend/S3 uploads
- [x] test on AWS with tabular and ISA-archive data sets (uploaded and downloaded files) using file system storage backend/S3 uploads (from scratch and from snapshots)
- [x] test on AWS with tabular and ISA-archive data sets (uploaded and downloaded files) using S3 storage backend/S3 uploads
- [x] make sure log messages are output from functions in file_store.utils

PR #2723, opened 2018-04-18
File import bug fixes

hackdna:
* Terminate file import task when corresponding FileStoreItem instance is deleted
* Avoid starting a file import task when one is already running for a given file
* Update file import state in Solr is case of import errors
* Improve logging in file store Celery tasks
* Improve directory making logic


refinery/data_set_manager/search_indexes.py
```diff
@@ -118,46 +117,17 @@ def prepare(self, node):
...
-                        except TaskMeta.DoesNotExist:
-                            logger.debug(
-                                "No file_import task for FileStoreItem with "
-                                "UUID: %s",
-                                file_store_item.uuid
-                            )
-                            download_url = NOT_AVAILABLE
-                        else:
-                            download_url = PENDING
+            download_url_or_state = file_store_item.get_datafile_url()
```
scottx611x: Can we just call this `download_url` and below use `import_state`/`import_status`? It seems that `get_datafile_url()` won't return a FileStoreItem's import state, and `get_import_status()` won't return download url information.

refinery/file_store/test_models.py
```diff
@@ -304,3 +291,60 @@ def test_translate_from_relative_path_with_no_args(self):
...
+
+    def test_no_terminate_on_save_with_no_new_datafile(self):
+        with mock.patch.object(
+            FileStoreItem, 'terminate_file_import_task'
+        ) as mock_terminate_task:
+            self.item.save()
+            mock_terminate_task.assert_not_called()
+
+
+@override_settings(REFINERY_DATA_IMPORT_DIR='/import/path')
```
scottx611x: Does the entire test class need this decorator, or just: `test_symlink_new_item_with_import_dir_path_source`?

refinery/file_store/test_models.py
```diff
@@ -220,26 +216,17 @@ def test_set_local_file_type_update(self):
...
 
-    @override_storage()
     def test_delete_local_file_on_instance_delete(self):
         self.item.datafile.save(self.file_name, ContentFile(''))
-        with mock.patch.object(FieldFile, 'delete') as mock_delete:
-            self.item.delete()
-            mock_delete.assert_called_with(save=False)
-
-    @override_storage()
-    def test_delete_local_file_on_instance_update(self):
```
scottx611x: Is this a behavior that we don't want to test anymore?

refinery/file_store/tasks.py
```diff
@@ -42,39 +39,101 @@ def import_file(uuid, refresh=False, file_size=0):
...
+
+    # exit if an import task is already running for this file
+    if item.import_task_id:
+        result = celery.result.AsyncResult(item.import_task_id)
+        if result.state in list(celery.states.UNREADY_STATES) + ['PROGRESS']:
+            logger.error(
+                "File import is already in progress for '%s' - task ID: '%s'",
+                item, item.import_task_id
+            )
+            import_file.update_state(state=celery.states.FAILURE,
```
scottx611x: It seems like this pattern:
```
import_file.update_state(state=celery.states.FAILURE,
                         meta='Failed to import file')
# http://stackoverflow.com/a/33143545
raise celery.exceptions.Ignore()
```
is repeated in a few places within `import_file`. Would it be worth factoring out?

refinery/file_store/models.py
```diff
@@ -177,24 +171,11 @@ def save(self, *args, **kwargs):
...
-                pass  # this is a newly created instance
-            except FileStoreItem.MultipleObjectsReturned as exc:
-                logger.critical(
-                    "Error retrieving FileStoreItem with primary key '%s': %s",
-                    self.pk, exc
-                )
-            else:
-                old_instance.delete_datafile(save_instance=False)
+        # symlink datafile if possible
+        if (not self.datafile and os.path.isabs(self.source) and
```
scottx611x: If this logic is required for a datafile to be symlinked, would it be better off residing within `_symlink_datafile`?

refinery/file_store/test_models.py
hackdna: I haven't checked but when settings are explicit tests are less brittle.

refinery/file_store/test_models.py
hackdna: Correct.

refinery/file_store/tasks.py
hackdna: The whole `import_file` needs to be split up but this is a [separate issue](https://github.com/refinery-platform/refinery-platform/issues/2359).

refinery/file_store/models.py
hackdna: No, this is a separate piece of functionality; `_symlink_datafile` only handles symlinking files on disk.

scottx611x: Okay, but couldn't `_symlink_datafile` get used elsewhere and improperly symlink data files if the large `if` above it isn't copied over properly?

hackdna: That is why it is declared as a "private" function.

scottx611x: Yes, but that doesn't prevent reuse within `FileStoreItem` (or anywhere else). I think that it makes sense that if there is a specific set of criteria for being able to symlink a data file, `_symlink_datafile` would handle it. Otherwise, this interface can potentially provide undesired functionality.

`_symlink_datafile` is already handling the case where the specified `source` isn't a file. Why wouldn't we move this logic within it as well?

refinery/file_store/tasks.py
scottx611x: I think having these ~10 similar chunks of code is just creating more overhead for when https://github.com/refinery-platform/refinery-platform/issues/2359 is tackled, but its ultimately up to you.

refinery/file_store/models.py
hackdna: When a function is declared "private" it is not a part of the API, i.e. not to be used outside the file_store app.
What this does provide is an abstraction of low-level functionality of manipulating the file system from high-level functionality of saving a FileStoreItem instance. This makes the code easy to understand and test.

scottx611x: Yes, I understand that...

Let's say someone calls `self._symlink_datafile()` within a new `FileStoreItem` method they introduce with the intent to symlink a datafile. They would need to remember to check:
```
if (not self.datafile and os.path.isabs(self.source) and settings.REFINERY_DATA_IMPORT_DIR 
    not in self.source and get_temp_dir() not in self.source)
```
beforehand, otherwise they could do something that wasn't originally intended, correct?

PR #2724, opened 2018-04-18
Remove unused django-resumable Python module

hackdna:


hackdna:
This module provides functionality similar to django-chunked-upload but it was never actually used AFAIK. Resumable uploads could be implemented with django-chunked-upload but there is no immediate demand.

PR #2731, opened 2018-04-25
Add creation_date timestamp to Visualizations

scottx611x:
@jkmarx I can change the base of this PR if it makes more sense to include it in the `fix-vis-ui` branch

screen shot 2018-04-25 at 11 35 36 am

### DIFF
--- a/refinery/ui/source/js/data-set-visualization/partials/visualization.html
+++ b/refinery/ui/source/js/data-set-visualization/partials/visualization.html
@@ -62,7 +62,7 @@
             </span>
           </div>
           <div class="col-md-5 analysis-name">
-            {{ vis.name }}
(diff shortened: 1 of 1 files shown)


jkmarx:
Isn't this misleading? Why add it only to the front end, especially if creation_time is already provided?

hackdna:
What if you have clients in different time zones?

ngehlenborg:
It doesn't matter. This is only to distinguish vis tool launches from each other.

hackdna:
It does matter because it would be difficult for users to distinguish vis tool launches from one another if the timestamps are in the wrong time zone.

ngehlenborg:
After speaking to Scott, I decided to postpone this. We need at least an additional field to support what I had in mind.

PR #2733, opened 2018-04-25
Escape @ symbol to fix filter bug.

### LINKED ISSUE #2383: Analysis Filters don't seem to work on some Analyses
* Specific code commit: b301acb2e32eb5f04cecd7264188154fbfaf5f2c
* Version of the web browser and OS: any
`beta.stemcellcommons.org`
### Steps to reproduce
1.  go here: https://beta.stemcellcommons.org/data_sets/ca5ac159-4468-4317-b03d-e9d00550f3e1/#/files/
2. Filter by :`RNA-seq paired end with TopHat - hg19 2017-8-1@12:10:41`

### Observed behavior
* The url is updated -> https://beta.stemcellcommons.org/data_sets/ca5ac159-4468-4317-b03d-e9d00550f3e1/#/files/?%7B%22REFINERY_ANALYSIS_UUID_604_306_s%22:%22RNA-seq%2520paired%2520end%2520with%2520TopHat%2520-%2520hg19%25202017-8-1%254012%253A10%253A41%22%7D but the listing is empty
screen shot 2017-11-16 at 4 36 55 pm
* This behavior does not seem to exist when filtering with newer Analyses
* No console errors

### Expected behavior
* To be able to filter on any Analysis

### General Notes
* Newer Analyses names don't have an `@` symbol in them like the legacy ones do. Not sure if this would affect filtering, but it could be a lead to fix the issue?

jkmarx:
Resolves #2383 
@scottx611x mentioned an issue with the @ symbol.

attribute-filter-fix

### DIFF
--- a/refinery/data_set_manager/utils.py
+++ b/refinery/data_set_manager/utils.py
@@ -21,8 +21,10 @@ import requests
 
 import core
 
-from .models import (AnnotatedNode, AnnotatedNodeRegistry, Assay, Attribute,
-                     AttributeOrder, Node, Study)
+from .models import (
+    AnnotatedNode, AnnotatedNodeRegistry, Assay, Attribute, AttributeOrder,
+    Node, Study
+)
 from .search_indexes import NodeIndex
 from .serializers import AttributeOrderSerializer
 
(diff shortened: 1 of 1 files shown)


refinery/data_set_manager/utils.py
```diff
@@ -21,8 +21,10 @@
 
 import core
 
-from .models import (AnnotatedNode, AnnotatedNodeRegistry, Assay, Attribute,
-                     AttributeOrder, Node, Study)
+from .models import (
```
jkmarx: auto isort fix, should I revert it?

PR #2736, opened 2018-04-27
Scottx611x/prettier tool auth page

scottx611x:
Before:
pasted image at 2018_04_25 11_51 pm

After:
screen shot 2018-04-27 at 10 22 44 am


refinery/templates/tool_manager/vis-tool-user-not-allowed.html
```diff
@@ -0,0 +1,28 @@
+{% extends "base.html" %}
+
+{% block title %}
+Visualization Error
```
jkmarx: the formatting seems off. HTML -> 1 tab/ 2 space. Some blocks' interiors need indenting.

refinery/tool_manager/views.py
```diff
@@ -214,8 +220,11 @@ def _proxy_view(self, request, container_name, url):
...
         )
         if not user_has_access_to_tool(request.user, visualization_tool):
-            return HttpResponseForbidden(
-                "User does not have permission to view Tool: {}".format(
+            return render_vis_tool_user_not_allowed_template(
+                request,
+                visualization_tool.name,
+                "User: {} does not have permission to view Tool: {}".format(
+                    request.user.username,
                     visualization_tool.uuid
```
jkmarx: Would the tool name be more useful than the tool_uuid?

scottx611x: We still have the Tool name listed in the sub header. I figured it was useful to include something unique about the Tool that erred to help someone debug if they got a screen shot from an end user

jkmarx: That's a fair point. I'd recommend including the tool name in the main body also. We are getting rid of the subheader soon (1-2 milestones).

PR #2737, opened 2018-05-02
Add basic functionality for creating S3 user data object keys

hackdna:

### DIFF
--- a/refinery/file_store/test_utils.py
+++ b/refinery/file_store/test_utils.py
@@ -4,22 +4,66 @@ from urlparse import urljoin
 from django.conf import settings
 from django.test import SimpleTestCase
 
-from .utils import SymlinkedFileSystemStorage
+import mock
+
+from .utils import S3MediaStorage, SymlinkedFileSystemStorage
+
+
+class S3MediaStorageTest(SimpleTestCase):
+
+    def setUp(self):
+        self.storage = S3MediaStorage(bucket_name='test-bucket')
+
+    @mock.patch('file_store.utils.S3MediaStorage.exists')
+    def test_file_name_format(self, mock_exists):
+        mock_exists.return_value = False
+        name = self.storage.get_available_name('test.fastq')
+        self.assertRegexpMatches(name, '[\w]{7}/test\.fastq')
+
+    @mock.patch('file_store.utils.S3MediaStorage.exists')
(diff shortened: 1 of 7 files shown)


refinery/file_store/test_utils.py
```diff
@@ -4,22 +4,66 @@
...
+class S3MediaStorageTest(SimpleTestCase):
+
+    def setUp(self):
+        self.storage = S3MediaStorage(bucket_name='test-bucket')
+
+    @mock.patch('file_store.utils.S3MediaStorage.exists')
+    def test_file_name_format(self, mock_exists):
+        mock_exists.return_value = False
+        name = self.storage.get_available_name('test.fastq')
+        self.assertRegexpMatches(name, '[\w]{7}/test\.fastq')
```
scottx611x: What does the 7 correlate to here?

I'm guessing: `/xx/xx/`? Maybe a more explicit regex would be clearer?

refinery/file_store/utils.py
```diff
@@ -2,11 +2,28 @@
...
 from django.conf import settings
 from django.core.files.storage import FileSystemStorage
+from django.utils.crypto import get_random_string
 from django.utils.deconstruct import deconstructible
+from django.utils.text import get_valid_filename
 
 from storages.backends.s3boto3 import S3Boto3Storage
 
 
+class S3MediaStorage(S3Boto3Storage):
```
scottx611x: Would it make sense to have an abstract base for these two with a placeholder get_available_name that raises NotImplementedError or something of the like?

I also don't envision there being any other storage backends for now so that might be overkill

refinery/file_store/test_utils.py
hackdna: That 7 corresponds to the one in [S3MediaStorage.get_available_name()](https://github.com/refinery-platform/refinery-platform/pull/2737/files#diff-e861c3dd97417068aa544e06b9ba89fdR22) which is what's being tested here.

Forward slash is not a special character: https://docs.python.org/2/library/re.html#regular-expression-syntax

refinery/file_store/utils.py
hackdna: Their common parent is the [Storage class provided by Django](https://docs.djangoproject.com/en/1.8/_modules/django/core/files/storage/#Storage)

PR #2740, opened 2018-05-03
Scottx611x/fix process metadata table command

scottx611x:
- Fixes the `process_meta_data_table` command
- Add test coverage

This fix is necessary to be able to import the metadata and data files for the `encode-x` instance of Refinery


refinery/data_set_manager/management/commands/process_metadata_table.py
```diff
@@ -98,6 +98,12 @@ def add_arguments(self, parser):
             default=False,
             help='flag for whether this data set will be visible to the public'
         )
+        parser.add_argument(
+            '--delimiter',
+            action='store_true',
+            default="tab",
+            help='Delimiter to use for metadata file parsing'
```
mccalluc: Do you want plain `action='store'` here?

Also, from your default value, it also looks like you'll look up the actual character from some enumeration?

refinery/data_set_manager/tests.py
```diff
@@ -2291,6 +2292,11 @@ def test_post_good_isa_tab_url(self, delay_mock):
         self.successful_import_assertions()
 
 
+# This override_settings allows for test-data/ fixtures to be copied into
+# Refinery's file_store on import rather than being moved in Travis CI test
+# runs
```
mccalluc: Grammar: add "."

mccalluc: And I might be out of the loop, but the explanation isn't really clear to me... this isn't python `import`, right? And it's a copy in either case, right?

PR #2746, opened 2018-05-08
Jkmarx/fix history bug

jkmarx:
- Fix logic to test whether a tool is an analysis or visualization.


jkmarx:
@scottx611x That works for me. I can close this pull request then.

PR #2748, opened 2018-05-08
Switch network adapter type to paravirtualized for increased performance

hackdna:
Increase guest OS network performance by ~100x for faster S3 transfers and provisioning

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -18,6 +18,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provider "virtualbox" do |v|
     v.memory = 2048
     v.cpus = 1
+    # To increase guest network performance (https://superuser.com/a/850389)
+    v.customize ["modifyvm", :id, "--nictype1", "virtio"]
   end
 
   config.ssh.forward_agent = true  # to enable cloning from Github over SSH


hackdna:
We do now.

scottx611x:
@hackdna Oh, okay I wasn't aware. 

So someone couldn't properly develop locally if their network connection went down? 
Does one have to run terraform commands after provisioning a vagrant box?

hackdna:
Well, it's not really possible to develop "locally" anymore because a lot of tests require an Internet connection.

Regardless, the file storage backend can be specified using [DEFAULT_FILE_STORAGE Django setting](https://docs.djangoproject.com/en/1.8/ref/settings/#default-file-storage). Default is `SymlinkedFileSystemStorage` (derived from `django.core.files.storage.FileSystemStorage`) in Vagrant and it does not require an Internet connection.

The only requirement to enable the new functionality in this PR is to run `vagrant reload`.

scottx611x:
Okay sure, I think I was more confused about your comment here: https://github.com/refinery-platform/refinery-platform/pull/2748#issuecomment-387503096... But you're just switching the storage backend to S3Storage to do some file_store specific testing, correct? It's not something we normally do, but something we can do if necessary. 

Do you think it would be useful to have some notes on how one could configure their Vagrant instance to be able to talk to s3? Its a little more involved than switching the storage backend, no?

Why not just bring up an AWS stack to test such functionality?

hackdna:
The change in this PR is completely backward-compatible and does not require any configuration or updates. Once S3 storage backend support is fully functional it will also have all the necessary documentation.

hackdna:
Happy to discuss this in depth at the next dev meeting.

PR #2752, opened 2018-05-09
Jkmarx/fix group api bug

### LINKED ISSUE #1724: Dashboard: Filter with Groups Bug
* Specific code commit: 34678985d9904406311d556150f1a01bd01acf2c
* Version of the web browser and OS: Version 58.0.3029.110 (64-bit), v10.11.6
Test site

### Steps to reproduce

1. On the dashboard, select own or public
2. Then select a group set

### Observed behavior
dashboard-filter-bug

### Expected behavior
Should show the intersection

jkmarx:
Resolves #1724 

dashboard-filter-fix
launch-pad-filter

### DIFF
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -552,6 +552,20 @@ class DataSetResource(SharableResourceAPIInterface, ModelResource):
 
         investigation_link = bundle.obj.get_latest_investigation_link()
 
+        groups = GroupObjectPermission.objects.filter(
+            object_pk=bundle.obj.id
+        )
+
+        if bundle.obj.public is None:
+            for group in groups:
+                if group.group == ExtendedGroup.objects.public_group():
+                    bundle.data["public"] = True
+
+        if bundle.obj.is_owner is None:
+            bundle.data["is_owner"] = self.request.user.has_perm(
+                'core.share_dataset', bundle.obj
+            )
+
         bundle.data["analyses"] = analyses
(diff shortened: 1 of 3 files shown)


refinery/core/api.py
```diff
@@ -677,9 +677,15 @@ def filter_by_group(self, request, obj_list):
                 group = None
 
             if group:
-                obj_list = list(get_objects_for_group(
+                filtered_obj_list = []
+                group_obj_list = list(get_objects_for_group(
```
scottx611x: It seems odd that `get_objects_for_group` is returning more than what we want, but if thats the case you could save a few lines with an approach like:
```
objects_for_group = get_objects_for_group(
    group, 'core.read_meta_dataset'
)
# get_objects_for_group returns too much, filter against initial obj_list
return [obj for obj in objects_for_group if obj in obj_list]
```

Also maybe making the comment more clear could be helpful. I'm not sure what the `ref in bundle` portion means in this case

jkmarx: @scottx611x Take a look at get_object_list
filter_by_group is a helper method, 'referenced in data bundle'. I can update the comment.

We're grabbing the (owned/public) filtered obj list and then grabbing the group filtered list. I just added the intersection code. Perviously, the code just dropped the first filtered obj list if a group filter was added.

I think it's worth a few lines of extra code if it's easier to read.

PR #2754, opened 2018-05-10
Jkmarx/min data set api

jkmarx:
- Alleviate some performance issues with the dashboard data set retrieval

### DIFF
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -510,53 +510,54 @@ class DataSetResource(SharableResourceAPIInterface, ModelResource):
         }
 
     def dehydrate(self, bundle):
-        if not bundle.data['owner']:
-            owner = bundle.obj.get_owner()
-            try:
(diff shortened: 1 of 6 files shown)


refinery/core/api.py
```diff
@@ -510,53 +510,54 @@ class Meta:
...
-        analyses = []
-        for analysis in bundle.obj.get_analyses():
-            analysis_dict = {}
-            analysis_dict["uuid"] = analysis.uuid
-            analysis_dict["name"] = analysis.name
-            analysis_dict["status"] = analysis.status
-            analysis_dict['is_owner'] = False
-            owner = analysis.get_owner()
-            if owner:
+        if 'min_response' not in bundle.request.GET:
```
scottx611x: If we flipped this logic we could save all the indenting below
```
if 'min_response' in bundle.request.GET:
    return bundle
```

PR #2755, opened 2018-05-11
Jkmarx/refresh group list

jkmarx:
- Fix issue with groups not refreshing after last invite revokes
- Fix padding 
- Refactor a bit so only a group invite is refreshed


refinery/ui/source/styles/refinery-style.less
```diff
@@ -1951,15 +1951,6 @@ a.modal-close {
   color: @gray-light;
 }
 
-
-.reinvite-icon {
```
jkmarx: unused icon

PR #2770, opened 2018-05-15
Scottx611x/terminate file import tasks on analysis cancel

scottx611x:
I observed today that an Analysis' `import_file` tasks will not be terminated if an analysis is cancelled before they have completed.

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -1209,6 +1209,7 @@ class Analysis(OwnableResource):
     def cancel(self):
         """Mark analysis as cancelled"""
         self.canceled = True
+        self.terminate_file_import_tasks()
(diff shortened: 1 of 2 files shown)


refinery/core/models.py
```diff
@@ -1209,6 +1209,7 @@ def galaxy_cleanup(self):
     def cancel(self):
         """Mark analysis as cancelled"""
         self.canceled = True
+        self.terminate_file_import_tasks()
```
hackdna: Doesn't this already happen here: https://github.com/refinery-platform/refinery-platform/blob/45ab3a3d9ed1705f4675f58f6db6984c759d6ad6/refinery/analysis_manager/tasks.py#L298

scottx611x: Yeah I believe you're right. I thought I observed different behavior yesterday

Issue #2545, opened 2018-01-29
Support for custom Tool launch names


scottx611x:
Back in the day, we were able to Launch Workflows with a custom name. As of now (bea7f6d3995d227a4c553ef113a8ac8ff22c81ac) this isn't the case.

I think with minimal ui and `/api/v2/tools` api tweaks this would be possible again.

### Potential Blockers:
- ~~~Legacy Analyses (Ones containing an `@` symbol in their name) [aren't currently able to be filtered in the File Browser](https://github.com/refinery-platform/refinery-platform/issues/2383). If we are to introduce this change we should fix that as well and be sure to properly validate any incoming custom names.~~~ Fixed by: https://github.com/refinery-platform/refinery-platform/pull/2733
- Allowing for custom names for `Tools` could create trouble when routing traffic to containers if we don't properly validate user input (currently we route traffic to docker containers with a url like: `/api/v2/tools/<tool_name_without_spaces>-<tool_uuid>/` which preserves uniqueness and remains a [compliant docker container name](https://github.com/moby/moby/issues/3138)). It would be better to just add a `display_name` field to the `Tool` model and store a user-defined name there.

scottx611x:
Currently when Tools are launched its confusing in the current Tool Launch listing as to which entry corresponds to which Tool that was launched.

See: 
<img width="421" alt="screen shot 2018-05-14 at 2 36 47 pm" src="https://user-images.githubusercontent.com/5629547/40016489-7a2bbffc-5784-11e8-82ae-3a668f36278f.png">

This is mainly due to lack of custom name support

jkmarx:
![screen shot 2018-05-18 at 8 27 16 am](https://user-images.githubusercontent.com/9956764/40234561-51225f78-5a75-11e8-859b-ba8a281a2dc8.png)

PR #2772, opened 2018-05-17
Scottx611x/custom tool launch names api changes

scottx611x:
@jkmarx This is the beginnings towards a fix for: #2545 

This adds a field: `"display_name"` to the `ToolLaunchConfiguration`/`Tool` model

Things to think about: 
 - If a user doesn't specify a custom `display_name` A reasonable default (Tool name + creation_date + Tool owner's username) is set. I think this is reasonable, but I could be wrong (especially with the timestamp part...)
- How would the Tool launch UI display the form to allow a user to set this value before a Tool is launched?


jkmarx:
@scottx611x I'll try to do a mock up tomorrow morning before the team meeting

refinery/tool_manager/models.py
```diff
@@ -298,6 +298,11 @@ class Tool(OwnableResource):
     )
     tool_launch_configuration = models.TextField()
     tool_definition = models.ForeignKey(ToolDefinition)
+    display_name = models.CharField(
+        max_length=250,
+        unique=True,
```
jkmarx: Unique names per tool? Is this is necessary or wanted? @ngehlenborg @mccalluc  Thoughts? How are we error handling this (what would be the api response)?

scottx611x: User-supplied `display_names` are to be unique per @ngehlenborg's request.
The api response will be a 400 error with the message: `A Tool already exists with a display_name of: '<tool_display_name>'`

PR #2804, opened 2018-06-04
Jkmarx/ui tool launch input

### LINKED ISSUE #2545: Support for custom Tool launch names

jkmarx:
Resolves #2545 

screen shot 2018-06-04 at 2 31 56 pm
screen shot 2018-06-04 at 2 32 03 pm

### DIFF
--- a/refinery/ui/source/js/dashboard/partials/history-card.html
+++ b/refinery/ui/source/js/dashboard/partials/history-card.html
@@ -41,7 +41,9 @@
              <span ng-if="!tool.owner.full_name">
                {{ tool.owner.username }}
              </span>
-              launched a visualization tool: {{ tool.name }}.
+              launched a visualization tool: {{ tool.display_name }} <span
+             ng-if="!tool
+             .display_name">{{ tool.name }}</span>
            </a>
          </span>
        </li>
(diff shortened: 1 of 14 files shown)


refinery/ui/source/js/dashboard/partials/history-card.html
```diff
@@ -28,7 +28,8 @@ <h4 class="emphasized text-align-center m-t-1-5">
              <span ng-if="!tool.owner.full_name">
                {{ tool.owner.username }}
              </span>
-             launched an analysis: {{ tool.name }}
+             launched an analysis: {{ tool.display_name }} <span ng-if="!tool
```
jkmarx: this should over old vis/analysis tools w/o any display names

PR #2806, opened 2018-06-05
Hackdna/import file s3

hackdna:
Add functionality for importing files from POSIX paths to S3

### DIFF
--- a/refinery/file_store/models.py
+++ b/refinery/file_store/models.py
@@ -15,7 +15,6 @@ import re
 import urlparse
 
 from django.conf import settings
-from django.core.files import File
 from django.core.files.storage import default_storage
 from django.db import models
 from django.db.models.signals import post_delete
@@ -204,23 +203,6 @@ class FileStoreItem(models.Model):
             return _get_extension_from_string(self.datafile.name)
         else:
             return _get_extension_from_string(self.source)
-
-    def import_file(self):
-        """Add file specified by source field to file store"""
-        logger.debug("Saving file '%s' to file store", self.source)
-        with open(self.source, 'rb') as external_file:
-            self.datafile.save(os.path.basename(self.source),
-                               File(external_file))
-        logger.info("Saved file to '%s'", self.datafile.name)
-
-        if self.source.startswith((settings.REFINERY_DATA_IMPORT_DIR,
-                                   get_temp_dir())):
-            try:
-                os.unlink(self.source)
-            except IOError as exc:
-                logger.debug("Failed to delete '%s': %s", self.source, exc)
-            else:
-                logger.debug("Deleted source file '%s'", self.source)
 
     def is_symlinked(self):
         '''Check if the data file is a symlink.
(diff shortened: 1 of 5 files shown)


refinery/file_store/models.py
```diff
@@ -34,9 +34,9 @@ def _mkdir(path):
     # https://stackoverflow.com/a/14364249
     try:
         os.makedirs(path)
-    except OSError:
+    except OSError as exc:
         if not os.path.isdir(path):
-            raise
+            raise RuntimeError(str(exc))
```
scottx611x: Does this need to be wrapped as a string?

refinery/file_store/tasks.py
```diff
@@ -20,11 +21,129 @@
...
+    if delete_source:
+        _delete_file(source_path)
+
+
+def _symlink_file(source_path, link_path):
+    _mkdir(os.path.dirname(link_path))
+    try:
+        os.symlink(source_path, link_path)
+    except OSError as exc:
+        raise RuntimeError("Error creating symlink '%s': %s", link_path, exc)
```
scottx611x: Should the dir created from: `_mkdir(os.path.dirname(link_path))` be removed on error?

refinery/file_store/tasks.py
```diff
@@ -20,11 +21,129 @@
...
+    file_store_name = storage.get_name(os.path.basename(source_path))
+
+    try:
+        s3.upload_file(source_path, settings.MEDIA_BUCKET, file_store_name,
+                       ExtraArgs={'ACL': 'public-read'},
+                       Callback=ProgressPercentage(source_path,
+                                                   import_file.request.id))
+    except botocore.exceptions.ClientError as exc:
+        s3_url = 's3://' + settings.MEDIA_BUCKET + '/' + file_store_name
+        raise RuntimeError("Error transferring file '%s' to %s: %s",
```
scottx611x: What if a large `source_path` is referenced here and fails right at the end? Do we leave the incomplete data in S3 or does it get removed somehow? Isn't there something tricky about multipart uploads where one can still be charged for those smaller chunks if a erring upload happens?

refinery/file_store/models.py
hackdna: Yes, we need to pass a string to RuntimeError. Explicit is better than implicit. https://www.python.org/dev/peps/pep-0020/

refinery/file_store/tasks.py
hackdna: I don't think it is necessary in this context.

refinery/file_store/tasks.py
hackdna: This will be handled by setting an appropriate bucket policy.

scottx611x: Is this bucket policy on our timeline anywhere? This will be handled in the Terraform portion of our deployment correct?

hackdna: Yes and yes.

PR #2807, opened 2018-06-06
Jkmarx/update to registration form

### LINKED ISSUE #2797: Spam Registrations
Hinder spam bots from registering on instance
- [ ] UI side, users can not register with the same first & last name
- [ ] UI side, hide a form which when filled will deactivate submit button

jkmarx:
Resolves #2797

### DIFF
--- a/refinery/core/forms.py
+++ b/refinery/core/forms.py
@@ -45,6 +45,12 @@ class RegistrationFormCustomFields(RegistrationForm):
         label=_("Affiliation")
     )
 
+    def clean_last_name(self):
+        if self.cleaned_data['last_name'] == self.cleaned_data['first_name']:
+            raise ValidationError('First and last name cannot be the same.',
+                                  code='invalid_names')
+        return self.cleaned_data['last_name']
+
 
 class RegistrationFormWithCustomFields(
     RegistrationFormUniqueEmail,
--- a/refinery/core/forms.py
+++ b/refinery/core/forms.py
@@ -44,12 +44,22 @@ class RegistrationFormCustomFields(RegistrationForm):
         error_messages={'required': _("You must provide an Affiliation")},
         label=_("Affiliation")
     )
+    spam_filter = forms.CharField(
+        widget=forms.HiddenInput(),
+        required=False
+    )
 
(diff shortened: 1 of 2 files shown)


refinery/templates/registration/registration_form.html
```diff
@@ -93,6 +93,15 @@ <h4>Terms of Service</h4>
 			</label>
 			{% endcomment %}
 
+      {% if form.spam_filter.errors %}
```
jkmarx: the spacing looks normal in pycharm. Unsure what's going on here.

refinery/core/forms.py
```diff
@@ -44,6 +44,22 @@ class RegistrationFormCustomFields(RegistrationForm):
         error_messages={'required': _("You must provide an Affiliation")},
         label=_("Affiliation")
     )
+    spam_filter = forms.CharField(
+        widget=forms.HiddenInput(),
+        required=False
+    )
+
+    def clean_last_name(self):
```
scottx611x: Just thinking about this again, and its pretty uncommon for someone to have the same first and last name, but it does happen. Theres Li Li in Paul's lab for example. Maybe the spam filter portion is sufficient?

refinery/templates/registration/registration_form.html
scottx611x: Wondering if showing errors for hidden fields is necessary if this is a scenario that only spam bots will run into

jkmarx: @scottx611x There are form filling extensions which may fill this field. It will annoy the user, but most of our users are manually filling these forms.

refinery/core/forms.py
jkmarx: @scottx611x I really think this is fine. Li Li is the first time I've ever encountered the same first and last name ever. But we can mention it in the team meeting for feedback if you're concerned.

scottx611x: @jkmarx Alright, its really up to you. Though,  it would probably be easier to allow same first and last names now rather than having to go back and make a change when a user that fits this scenario comes along (as unlikely as it may be)

PR #2809, opened 2018-06-06
Mccalluc/more event logging

mccalluc:
Follow up on #2808. Right now I'm assuming tool-save == tool-creation, but that won't be true if tools can be renamed...


refinery/core/models.py
```diff
@@ -2357,15 +2357,21 @@ def render_dataset_visualization_creation(self):
...
+        blob = json.dumps({
+            'display_name': display_name
+        })
+        event = Event(dataset=dataset, user=user, json=blob,
+                      type=Event.UPDATE, sub_type=Event.ANALYSIS_CREATION)
+        event.save()
+
+    def render_dataset_analysis_creation(self):
+        data = json.loads(self.json)
+        return '{:%x %X}: {} launched analysis {} on data set {}'.format(
```
scottx611x: `{:%x %X}` is cool, I'll probably reuse when I trim the timestamp for `Tool's` default `display_names`

refinery/core/tests.py
```diff
@@ -2003,13 +2004,35 @@ def test_dataset_metadata_edit(self):
...
+
+        events = Event.objects.all()
+        self.assertEqual(len(events), 2)
+        self.assertRegexpMatches(
+            str(events[0]),
+            self.pre_re + r'created' + self.post_re
+        )
+        self.assertRegexpMatches(
+            str(events[1]),
+            self.pre_re + r'launched visualization None on' + self.post_re
```
scottx611x: The `None` here and below confuse me. It could be that you may not have the latest develop code, but I thought that `Tools` got a reasonable default for their `display_names`.

mccalluc: Let me look into that... It might be that I'm not filling in the object correctly, and that `None` is the actual value, and not the `str` for some object.

mccalluc:
I'll try making the dataset -> data_set change in the earlier PR, and then rebase this branch, and then fix any remaining ones.

mccalluc:
Actually, I'll merge this back to my other branch and fix the naming there.

Issue #2700, opened 2018-03-26
Metadata/Data File Editing for Existing DataSets


scottx611x:
Its a common use case for users to be able to edit errors in their metadata, and upload and associate new data files or replace existing data files after a successful metadata upload.

HSCI Scenarios:

- 1: An analyst uploads an ISA archive and corresponding data files but needs to correct an error in metadata
  - Conditions:
    - No files can be added or removed. Only typographical errors or errors in the content of the metadata are allowed to be edited.
    - Only ISATabs will be supported in this iteration. Tabular metadata upload requires some extra user intervention i.e. selection of `Data File Column`.
    - No `Analyses` or `Visualizations` can be performed on the DataSet
    - Data files uploaded before the edits have been made will be reassociated with the new `DataSet`

- 2: An analyst successfully uploads an ISA archive but needs to add some data files later
  - Functionality will be added to the File Browser to be able to upload and associate files with existing `Nodes` as well as provide a valid url to a data file.
    - This will potentially mean overwriting/purging existing data files so we should be wary of that
  - This functionality will exist for all `DataSets` regardless if they are ISATab or Tabular

### ToDos:
- Frontend
  - [x] Button to `Update DataSet` in the FileBrowser dependent on aforementioned conditions
  - ~~~[ ] Modal to allow selection of new ISATab metadatafile~~~ Redirects to import pages now
  - [x] Extend File Browser code to be able to upload and associate new data files with existing `Nodes`
- Backend
  - [x] Extend API endpoints for `DataSet` update
    - [x] DataSets should be able to report on their "cleanliness"  Ref: #2784
    - ~~~[ ] Check that newly referenced data files are a 1::1 match with what existed prior~~~ We ended up vetoing this and allowing removal and addition of datafiles during a metadata revision
    - [x] Reassociate data files with updated `DataSet` if referenced in new metadata
    - [x] `DataSet` version should be incremented

Add/remove/replace data files for any existing data set that `is_clean`

scottx611x:
Speeding up ISA-Tab import performance could alleviate the need to run some background tasks for upload.

scottx611x:
Are Nodes mutable?
If so, how does this affect DataSet versioning? i.e. if I edit a Node's meta data/(replace/remove/add) a datafile/add a new Node what does the resulting provenance graph structure look like, and what does this do to a DataSet's version?

scottx611x:
[...]

PR #2811, opened 2018-06-08
Scottx611x/metadata revisions

scottx611x:
Towards #2700 

Allows one to edit metadata files (ISATab or Tabular) outside of Refinery, and proceed to update an existing DataSet with revised content (grammatical errors, column and row introductions, etc.) while persisting previously uploaded data files (if they are referenced in the new metadata file, otherwise they will be deleted) and allowing a `User` to upload new ones along the way.

TODOS: 
- [x] Make it obvious that a metadata revision is happening on the import pages
- [x] Changes for tabular metadata revisions
- [x] Tests for tabular metadata revisions
- [x] Update "check datafiles" functionality
  - [ ] Add this for ISATab import as well???
- [x] Add tests for `CheckDataFilesView`
- [x] Add tests for rows being removed in a metadata revision
- [ ] Update button in details tab (Update Dataset metadata) with tooltip explaining what can happen. There is possibility for this button to exist in the file browser somewhere as well
- [x] Only DataSet owners should be able to edit metadata
- [x] Be able to delete rows during a meta data revision.
- [x] Have the check datafiles functionality be part of a Dataset upload/metadata revision, alerting the user as to what is actually happening with the datafiles, removed/added etc. before the actual dataset submission goes through


refinery/data_set_manager/tasks.py
```diff
@@ -260,10 +261,7 @@ def create_dataset(investigation_uuid, username, identifier=None, title=None,
     dataset.file_size = dataset.get_file_size()
     dataset.file_count = dataset.get_file_count()
     dataset.save()
-    # Finally index data set
-    update_data_set_index(dataset)
     add_data_set_to_neo4j(dataset, user.id)
-    async_update_annotation_sets_neo4j()
```
scottx611x: These are already happening in the `DataSet` model's `post_save` signal handler

refinery/data_set_manager/models.py
```diff
@@ -515,13 +567,13 @@ def get_file_store_item(self):
...
         there isn't one
         """
-        try:
-            return FileStoreItem.objects.get(
-                uuid=self.file_uuid)
-        except (FileStoreItem.DoesNotExist,
-                FileStoreItem.MultipleObjectsReturned) as e:
-            logger.error(e)
-            return None
+        if self.file_uuid:
```
scottx611x: Many Nodes aren't file Nodes. and this logger.error was widespread throughout logs making things confusing

refinery/data_set_manager/utils.py
```diff
@@ -1192,3 +1194,94 @@ def get_solr_response_json(node_uuids):
...
+
+
+def _ensure_revised_investigation_references_existing_datafiles(
+        existing_dataset, revised_investigation
+):
+    existing_investigation = existing_dataset.get_investigation()
+    existing_datafile_names, new_datafile_names = (
+        investigation.get_datafile_names(exclude_metadata_file=True) for
+        investigation in [existing_investigation, revised_investigation]
+    )
```
mccalluc: plain old assignment might be fine here, and/or, make both sets now, since that's what you need?

refinery/data_set_manager/utils.py
```diff
@@ -1192,3 +1194,94 @@ def get_solr_response_json(node_uuids):
...
+    # Ensure that the names of existing datafiles are a subset of the new ones
+    if not set(existing_datafile_names).issubset(set(new_datafile_names)):
```
mccalluc: If we assume we have sets at this point, why not do an assert with an operator? Also, is the default str output ok, and the default list str? 
```python
assert existing_datafile_names <= new_datafile_names,
    'Not all data files from {} are referenced "
    'in the revised metadata file. The following are "
    'missing: {}".format(
        existing_dataset, existing_datafile_names - new_datafile_names
    )

Comment (mccalluc) [reply refinery/data_set_manager/utils.py]: oh, I see below where this will be sent to the user.

File: refinery/ui/source/js/data-set-about/partials/details.html [hunk 5]
```diff
@@ -397,6 +397,12 @@ <h3>Meta Data</h3>
   </a>
     - {{ADCtrl.fileStoreItem.file_size | fileSizeFormat}}
 </p>
+<a
+  id="import-revised-metadata-button"
+  class="refinery-base btn btn-primary"
+  href="/data_set_manager/import/?data_set_name={{ ADCtrl.dataSet.name }}/#/isa-tab-import?data_set_uuid={{ ADCtrl.dataSet.uuid }}">
```
mccalluc: I think you acknowledged that this should change. URL encode, if nothing else?

refinery/ui/source/js/data-set-about/partials/details.html
```diff
@@ -408,6 +414,12 @@ <h3>Meta Data</h3>
 </div>
 <div ng-if="!(ADCtrl.dataSet.isa_archive || ADCtrl.dataSet.pre_isa_archive && ADCtrl.fileStoreItem.datafile)">
 <p class="emphasized">Not available.</p>
+<a
+  id="import-revised-metadata-button"
+  class="refinery-base btn btn-primary"
+  href="/data_set_manager/import/?data_set_name={{ ADCtrl.dataSet.name }}/#/?data_set_uuid={{ ADCtrl.dataSet.uuid }}">
```
mccalluc: ditto

refinery/ui/source/js/data-set-about/partials/details.html
```diff
@@ -408,6 +414,12 @@ <h3>Meta Data</h3>
...
+  Revise this DataSet's content by uploading a new metadata file
```
mccalluc: Intercapping is just for code: "data set" to the user?

refinery/ui/source/js/data-set-import/controllers/isa-tab-import.js
```diff
@@ -82,8 +84,12 @@
       formData.append('identity_id', AWS.config.credentials.identityId);
     }
 
+    var queryParams = {};
+    if (this.isMetaDataRevision) {
+      queryParams = { data_set_uuid: this.dataSetUUID };
+    }
```
mccalluc: Maybe just use `this.dataSetUUID` directly, instead of introducing the `this.isMetaDataRevision`? Eh.

refinery/ui/source/js/data-set-import/controllers/metadata-table.js
```diff
@@ -320,8 +325,12 @@ MetadataTableImportCtrl.prototype.startImport = function () {
     formData.append('identity_id', AWS.config.credentials.identityId);
   }
 
+  var queryParams = {};
+  if (this.isMetaDataRevision) {
+    queryParams = { data_set_uuid: this.dataSetUUID };
+  }
   return this.tabularFileImportApi
-    .create({}, formData)
+    .create(queryParams, formData)
```
mccalluc: This and the file above are probably on the radar to refactor?

refinery/ui/source/js/data-set-import/controllers/isa-tab-import.js
mccalluc: Never mind, I think you're right that it makes it more clear.

refinery/data_set_manager/tasks.py
```diff
@@ -233,7 +232,8 @@ def create_dataset(investigation_uuid, username, identifier=None, title=None,
 
     datasets = DataSet.objects.filter(name=dataset_name)
     # check if the investigation already exists
-    if len(datasets):  # if not 0, update dataset with new investigation
+    # if not 0, update dataset with new investigation
```
jkmarx: do we need this comment

refinery/data_set_manager/tests.py
```diff
@@ -1605,6 +1610,22 @@ def test_update_annotated_nodes(self):
         self.assertEqual(len(nodes_after), 0)
         # TODO: Is this the behavior we expect?
 
+    def test_update_existing_dataset_with_revised_investigation(self):
```
jkmarx: why not break it into three unit tests, one for get_version, get_investigation, and get_latest_investigation_link. It was make debugging easier down the road.

refinery/ui/source/js/data-set-import/partials/metadata-table.html
```diff
@@ -90,10 +91,10 @@ <h2>
         ng-if="metadataTable.isErrored">
         <div class="text-centered">
           <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
-          <p>
-            Import of this data set failed. Please ensure that the structure
-            of your metadata file matches that of the example metadata file.
-          </p>
+          <p ng-show="metadataTable.isMetaDataRevision">Metadata Revision of this data set failed.</p>
```
jkmarx: revision

refinery/ui/source/js/data-set-import/controllers/isa-tab-import.js
```diff
@@ -6,17 +6,20 @@
     .controller('IsaTabImportCtrl', IsaTabImportCtrl);
 
   IsaTabImportCtrl.$inject = [
-    '$log', '$rootScope', '$timeout', '$window', 'isaTabImportApi', 'settings'
+    '$log', '$rootScope', '$timeout', '$window', '$location', 'isaTabImportApi', 'settings'
   ];
```
jkmarx: ['$location', '$log', '$rootScope',.....]

refinery/data_set_manager/views.py
```diff
@@ -311,7 +320,8 @@ def post(self, request, *args, **kwargs):
                     request.user.username,
                     False,
                     response['data']['temp_file_path'],
-                    identity_id=identity_id
+                    identity_id=identity_id,
+                    existing_dataset_uuid=existing_dataset_uuid
```
jkmarx: data_set

refinery/data_set_manager/utils.py
```diff
@@ -1192,3 +1194,94 @@ def get_solr_response_json(node_uuids):
...
+                prior_file_store_item.datafile = None
+                prior_file_store_item.save()
+
+
+def _check_data_set_cleanliness(data_set):
+    if not data_set.is_clean():
+        raise RuntimeError("DataSet with UUID: {} is not clean (There have "
+                           "been Analyses or Visualizations performed on it) "
+                           "Remove these objects and try again"
+                           .format(data_set.uuid))
```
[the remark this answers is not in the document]
hackdna: I don't think this function needs to exist at all.

refinery/data_set_manager/views.py
```diff
@@ -65,7 +65,15 @@ class DataSetImportView(View):
 
     def get(self, request, *args, **kwargs):
         form = ImportISATabFileForm()
-        context = RequestContext(request, {'form': form})
+        data_set_name = request.GET.get('data_set_name')
+        context = RequestContext(
+            request,
+            {
+                'form': form,
+                'data_set_name': data_set_name.strip("/") if data_set_name
```
hackdna: Move data_set_name manipulations to line 68 above where it is initialized.

refinery/data_set_manager/views.py
hackdna: No need for an extra variable: `existing_dataset_uuid=request.GET.get('data_set_uuid')`

refinery/data_set_manager/views.py
```diff
@@ -525,14 +536,15 @@ def post(self, request, *args, **kwargs):
                 custom_delimiter_string=request.POST.get(
                     'custom_delimiter_string', False
                 ),
-                identity_id=identity_id
+                identity_id=identity_id,
+                existing_dataset_uuid=existing_dataset_uuid
```
hackdna: No need for an extra variable: `existing_dataset_uuid=request.GET.get('data_set_uuid')`

refinery/data_set_manager/utils.py
```diff
@@ -1192,3 +1194,94 @@ def get_solr_response_json(node_uuids):
         'data_set_manager'
     )
     return format_solr_response(solr_response)
+
+
+def update_existing_dataset_with_revised_investigation(existing_dataset_uuid,
```
hackdna: The long name is a code smell. This should simply be a method of the DataSet model.

refinery/data_set_manager/utils.py
```diff
@@ -1192,3 +1194,94 @@ def get_solr_response_json(node_uuids):
...
+    existing_dataset.update_investigation(
+        revised_investigation,
+        "Metadata Revision: for {}".format(updated_dataset_title)
+    )
+    data_set_manager.tasks.annotate_nodes(revised_investigation.uuid)
+    existing_dataset.set_title(updated_dataset_title)
+
+
+@transaction.atomic()
+def associate_datafiles_from_existing_dataset(existing_dataset,
```
hackdna: The long name is a code smell. This should simply be a method of the DataSet model.

refinery/core/models.py
```diff
@@ -800,6 +777,66 @@ def is_clean(self):
...
 
+    def set_title(self, title):
+        self.title = title
+        self.save()
+
+    def update_with_revised_investigation(self, investigation):
+        """
+        Update an existing DataSet's Investigation with a new Investigation as
+        long as some specific constraints are met.
+        - Said DataSet must be "clean" (No Analyses or Visualizations
```
scottx611x: Update this docstring

refinery/data_set_manager/tasks.py
```diff
@@ -233,7 +231,8 @@ def create_dataset(investigation_uuid, username, identifier=None, title=None,
...
+    if len(datasets):
```
scottx611x: Why do I need to check for `and not existing_dataset_uuid` below in an identical `if`?

refinery/ui/source/js/data-set-import/controllers/isa-tab-import.js
```diff
@@ -72,6 +74,15 @@
     this.$rootScope.$broadcast('clearFileInput', 'isaTabFile');
   };
 
+  IsaTabImportCtrl.prototype.confirmImport = function () {
```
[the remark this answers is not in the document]
jkmarx: https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#controlleras-with-vm

refinery/ui/source/js/data-set-import/controllers/isa-tab-import.spec.js
```diff
@@ -0,0 +1,89 @@
...
+  describe('IsaTabImportCtrl', function () {
+    var ctrl;
+    var importConfirmationService;
+    var isaTabImportApiService;
+
+    beforeEach(function () {
+      module('refineryApp');
+      module('refineryDataSetImport');
+      inject(function ($controller, $rootScope, _isaTabImportApi_, _importConfirmationService_) {
+        isaTabImportApiService = _isaTabImportApi_;
```
jkmarx: Previously we discussed avoiding the _syntax_. Dependencies should be alphabetical with outside libraries first and then alphabetical with our libraries. importConfir... before isaTab...

refinery/ui/source/js/data-set-import/controllers/isa-tab-import.spec.js
```diff
@@ -0,0 +1,89 @@
...
+      });
+    });
+
+    describe('vm.isMetaDataRevision', function () {
+      it('should be initially false', function () {
+        expect(ctrl.isMetaDataRevision).toEqual(false);
+      });
+    });
+
+    describe('vm.isaTabImportApi', function () {
```
jkmarx: We typically don't check for dependencies in unit tests. Is this a common thing elsewhere?

refinery/ui/source/js/data-set-import/controllers/metadata-table.js
```diff
@@ -11,7 +11,8 @@ function MetadataTableImportCtrl (
   fileSources,
   settings,
   tabularFileImportApi,
-  metadataStatusService
+  metadataStatusService,
+  importConfirmationService
```
jkmarx: fix order

refinery/ui/source/js/data-set-import/controllers/metadata-table.spec.js
```diff
@@ -0,0 +1,178 @@
...
+  describe('MetadataTableImportCtrl', function () {
+    var ctrl;
+    var importConfirmationService;
+    var fileSourcesService;
+    var metadataStatusService;
+    var tabularFileImportApiService;
+
+    beforeEach(function () {
+      module('refineryApp');
+      module('refineryDataSetImport');
```
jkmarx: same comments as above

refinery/ui/source/js/data-set-import/services/import-confirmation-service.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
```
jkmarx: should be wrapped in an iffy and https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#style-y040 

Angular Tool module should be up to date with code style also.

refinery/config/settings/base.py
```diff
@@ -435,7 +435,8 @@ def get_setting(name, settings=local_settings, default=None):
                 "styles/refinery-style-bootstrap-responsive.css",
                 "styles/refinery-style.css",
                 "vendor/fontawesome/css/font-awesome.min.css",
-                "vendor/intro-js/minified/introjs.min.css"]
+                "vendor/intro-js/minified/introjs.min.css",
+                "vendor/angular-confirm/dist/angular-confirm.min.css"]
```
jkmarx: If all we need is a modal, please use https://angular-ui.github.io/bootstrap/

refinery/ui/source/js/data-set-import/controllers/isa-tab-import.spec.js
scottx611x: Not sure, I was following a pattern I saw in another test file.

refinery/ui/source/js/data-set-import/services/import-confirmation-service.js
```diff
@@ -0,0 +1,34 @@
+/**
```
jkmarx: This should really be a component and not a service. Services should not be interacting with the view.  https://stackoverflow.com/questions/27398741/angularjs-controller-vs-service

refinery/ui/source/js/data-set-import/services/import-confirmation-service.spec.js
```diff
@@ -0,0 +1,20 @@
+// UNIT TESTING
```
jkmarx: iffy please

refinery/ui/source/js/data-set-import/services/file-sources.js
```diff
@@ -7,7 +7,7 @@ function fileSources ($resource, settings, dataSetImportSettings) {
     {},
     {
       check: {
-        isArray: true,
+        isArray: false,
```
jkmarx: I believe default is false

refinery/ui/source/js/data-set-import/controllers/metadata-table.spec.js
```diff
@@ -0,0 +1,178 @@
...
+    var importConfirmationService;
+    var metadataStatusService;
+    var tabularFileImportApiService;
+
+    beforeEach(function () {
+      module('refineryApp');
+      module('refineryDataSetImport');
+      inject(function ($controller, $rootScope, _fileSources_, _importConfirmationService_,
+                       _metadataStatusService_, _tabularFileImportApi_) {
+        fileSourcesService = _fileSources_;
```
jkmarx: Avoid _

refinery/ui/source/js/data-set-import/controllers/isa-tab-import.spec.js
```diff
@@ -0,0 +1,89 @@
...
+
+  describe('IsaTabImportCtrl', function () {
+    var ctrl;
+    var importConfirmationService;
+    var isaTabImportApiService;
+
+    beforeEach(function () {
+      module('refineryApp');
+      module('refineryDataSetImport');
+      inject(function ($controller, $rootScope, _importConfirmationService_, _isaTabImportApi_) {
```
jkmarx: avoid _

refinery/ui/source/js/data-set-import/services/import-confirmation-service.js
scottx611x: I think I'm getting a bit confused here. The link is comparing controllers vs. services. Is a component a broader term for a controller?

jkmarx: No, a component is a simplified directive. For any dom related things (buttons, modals, etc), they should be components. 
 
"Components are the basic UI building block of an Angular app"

"Services are typically used as the business logic of your application. Services are similar to singletons in the sense that they are created once, and the instance is maintained throughout the entire life cycle of your application. It is a good place to put your logical functions which many views or components will require, and also hold global cache which needs to be accessed throughout multiple areas in your application."

jkmarx: https://docs.angularjs.org/guide/concepts

scottx611x: Okay cool thank you for the clarification. Am I correct in saying that components are to be bundled with directives underneath the `xxxx/directives` directories? I haven't found anything in our codebase suggesting otherwise

PR #2812, opened 2018-06-08
Events API

mccalluc:
Think I've got access controls in place, but comments in the code where I felt like I didn't really understand things. As a skeleton, the biggest gap is probably the lack of group events: When they get filled in, it will probably uncover places where I was just thinking about data sets.

@jkmarx : I'd suggest just using the `message` field for now, to get something out quickly, and then later that can be replaced with html with links, but it's your call.

(And if I didn't resolve things from the last PR to your satisfaction, let me know that, too.)

If it looks ok, I'll probably leave this to you, now, and pick up scatterplot and the proxy again.


mccalluc:
Closing in favor of #2812: Will address access control later.

PR #2813, opened 2018-06-11
Mccalluc/events revert access control

mccalluc:
This is a subset of #2812... I'm having more trouble adding the access controls than I expected, but this much should be good.

### DIFF
--- a/refinery/core/serializers.py
+++ b/refinery/core/serializers.py
@@ -168,6 +168,11 @@ class NodeSerializer(serializers.HyperlinkedModelSerializer):
 
 
 class EventSerializer(serializers.ModelSerializer):
+    data_set = serializers.SlugRelatedField(
+        read_only=True,
+        slug_field='uuid'
+    )
+
     class Meta:
         model = Event
         fields = ['data_set', 'group', 'user', 'type', 'sub_type', 'json']
--- a/refinery/core/serializers.py
+++ b/refinery/core/serializers.py
@@ -7,7 +7,7 @@ from rest_framework.validators import UniqueValidator
 from data_set_manager.models import Node
 from file_store.models import FileStoreItem
 
-from .models import DataSet, Workflow
+from .models import DataSet, Event, Workflow
 
 logger = logging.getLogger(__name__)
 
@@ -165,3 +165,9 @@ class NodeSerializer(serializers.HyperlinkedModelSerializer):
             'ready_for_igv_detail_view',
             'file_uuid'
         ]
+
+
+class EventSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Event
+        fields = ['dataset', 'group', 'user', 'type', 'sub_type', 'json']
--- a/refinery/core/serializers.py
+++ b/refinery/core/serializers.py
@@ -198,8 +198,11 @@ class EventSerializer(serializers.ModelSerializer):
 
     json = _DeserializeJSONField()
 
+    date_time = serializers.DateTimeField()
+
     class Meta:
         model = Event
         fields = [
-            'data_set', 'group', 'user', 'type', 'sub_type', 'json', 'message'
+            'date_time', 'data_set', 'group', 'user',
+            'type', 'sub_type', 'json', 'message'
         ]
--- a/refinery/core/serializers.py
+++ b/refinery/core/serializers.py
@@ -173,6 +173,11 @@ class EventSerializer(serializers.ModelSerializer):
         slug_field='uuid'
     )
 
+    user = serializers.SlugRelatedField(
+        read_only=True,
+        slug_field='username'
+    )
+
     class Meta:
         model = Event
         fields = ['data_set', 'group', 'user', 'type', 'sub_type', 'json']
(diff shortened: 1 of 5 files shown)


mccalluc:
Test failure is from cypress on Satori:
```
  1) Satori Visualization Sorts nodes in list graph by name:
     CypressError: Timed out retrying: Expected to find element: '.node.lock-directly', but never found it.
```

refinery/core/serializers.py
```diff
@@ -165,3 +166,43 @@ class Meta:
...
+    )
+
+    user = serializers.SlugRelatedField(
+        read_only=True,
+        slug_field='username'
+    )
+
+    message = _StrObjectField()
+
+    json = _DeserializeJSONField()
```
scottx611x: Is the standard [JSONField](http://www.django-rest-framework.org/api-guide/fields/#jsonfield) not sufficient here?

refinery/core/serializers.py
```diff
@@ -165,3 +166,43 @@ class Meta:
...
+        read_only=True,
+        slug_field='uuid'
+    )
+
+    user = serializers.SlugRelatedField(
+        read_only=True,
+        slug_field='username'
+    )
+
+    message = _StrObjectField()
```
scottx611x: I think the same effect could be achieved without having to create a custom field with the introduction of an `Event.get_message()` and a [SerializerMethodField](http://www.django-rest-framework.org/api-guide/fields/#serializermethodfield) here

refinery/core/test_views.py
```diff
@@ -934,3 +939,86 @@ def test_get_workflow_graph(self):
             uuid=self.workflow.uuid
         )
         self.assertEqual(get_response.content, self.mock_workflow_graph)
+
+
+class EventApiV2Tests(APIV2TestCase):
+    def setUp(self):
+        create_public_group()
```
scottx611x: Not sure if quick test has wide enough adoption to warrant this change

refinery/core/test_views.py
```diff
@@ -934,3 +939,86 @@ def test_get_workflow_graph(self):
...
+        user = User.objects.create_user('testuser')
+        CuserMiddleware.set_user(user)
+
+        create_tool_with_necessary_models("VISUALIZATION")
+        create_tool_with_necessary_models("WORKFLOW")
+
+        events = Event.objects.all()
+        self.assertEqual(len(events), 4)
+
+        messages = [str(_) for _ in events]
```
scottx611x: `_` -> `event`?

refinery/core/views.py
```diff
@@ -698,6 +699,14 @@ class NodeViewSet(viewsets.ModelViewSet):
     # permission_classes = (IsAuthenticated,)
 
 
+class EventViewSet(viewsets.ModelViewSet):
+    """API endpoint that allows Events to be viewed"""
+    queryset = Event.objects.all()
+    serializer_class = EventSerializer
+    # Don't think I need a lookup_field?
```
scottx611x: Yeah I think the default is the `Object.pk`, and at the moment I don't think we're trying to have a detail view for `Events`

refinery/core/serializers.py
scottx611x: You can use `JSONField` while specifying a `source` that points to a method on the `Event model that will spit out `Event.json` as a dict

See an example here: https://github.com/refinery-platform/refinery-platform/blob/develop/refinery/tool_manager/serializers.py#L79

refinery/core/test_views.py
```diff
@@ -934,3 +939,86 @@ def test_get_workflow_graph(self):
...
+        # Necessary if using "--settings=config.settings.quick_test".
+        # TODO: Move to superclass?
+
+        super(EventApiV2Tests, self).setUp(
+            api_base_name="events/",
+            view=EventViewSet.as_view({"get": "list"})
+        )
+
+    def test_get_event_list(self):
+        user = User.objects.create_user('testuser')
```
jkmarx: snake_case

refinery/core/test_views.py
```diff
@@ -934,3 +939,86 @@ def test_get_workflow_graph(self):
...
+        display_names = [
+            json.loads(_.json).get('display_name') for _ in events
+        ]
+        date_times = [
+            _.date_time.isoformat().replace('+00:00', 'Z') for _ in events
+        ]
+
+        get_request = self.factory.get(urljoin(self.url_root, '/'))
+        get_response = self.view(get_request).render()
+        # TODO: Why do I need render()?
```
jkmarx: In general, can we remove todo questions from the code and add them to the dev meeting agenda?

refinery/core/test_views.py
```diff
@@ -934,3 +939,86 @@ def test_get_workflow_graph(self):
...
+
+        self.maxDiff = None
+        self.assertEqual(
```
jkmarx: Can we separate the len check and the content check? The way the test is written it's repetitive to check the length and check if the arrays are equal. For content, it'd be great to break it down a bit more and check for particular fields or focus on particular difference (ex one vis subtype vs one empty subtype). I'm just thinking how would I debug this test if xx changed?

refinery/core/views.py
jkmarx: Comment on which events are returned? Per user?

scottx611x:
Access control provided in PR: https://github.com/refinery-platform/refinery-platform/pull/2816

PR #2819, opened 2018-06-14
Hackdna/add files to dataset

hackdna:
Back end support for adding files to an existing data set.

### DIFF
--- a/refinery/data_set_manager/urls.py
+++ b/refinery/data_set_manager/urls.py
@@ -11,12 +11,11 @@ from django.views.decorators.csrf import csrf_exempt
 from constants import UUID_RE
 from rest_framework.routers import DefaultRouter
 
-from .views import (
-    Assays, AssaysAttributes, AssaysFiles, CheckDataFilesView,
-    ChunkedFileUploadCompleteView, ChunkedFileUploadView, DataSetImportView,
-    ImportISATabView, ProcessISATabView, ProcessMetadataTableView,
-    TakeOwnershipOfPublicDatasetView
-)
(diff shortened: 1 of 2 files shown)


refinery/data_set_manager/urls.py
```diff
@@ -45,7 +44,8 @@
     url(r'^import/take_ownership/$',
         login_required(TakeOwnershipOfPublicDatasetView.as_view()),
         name='take_ownership_of_public_dataset'),
-
+    url(r'^add-files/$', login_required(AddFilesToDataSetView.as_view()),
```
jkmarx: add_files align with our other urls

hackdna: Align in what sense?

jkmarx: our urls are snake_case

hackdna: OK. I didn't see anything about that mentioned in the coding style guide.
Also, Google says:
"We recommend that you use hyphens (-) instead of underscores (_) in your URLs."
https://support.google.com/webmasters/answer/76329

jkmarx: I just double checked and we mostly use snake_case except for endpoints in data_set_manager related to imports. In two scoops, it recommends underscores to dashes. I believe we agree on using it as a guide for our code.

hackdna: Actually, Two scoops only talks about underscores in URL pattern names (section 1.6.2). A quick [Google search](https://www.google.com/search?q=url+hyphen+vs+underscore) also appears to suggest that hyphens are preferable to underscores in URLs. Perhaps it is time to set the standard?

Issue #2331, opened 2017-11-06
Is GuardianTastyPieAuthz.py needed?


scottx611x:
We have [refinery-platform/refinery/GuardianTastyPieAuthz.py:GuardianAuthorization](https://github.com/refinery-platform/refinery-platform/blob/master/refinery/GuardianTastypieAuthz.py#L12) lying around with its only usages in our codebase being [commented out](https://github.com/refinery-platform/refinery-platform/search?utf8=%E2%9C%93&q=GuardianAuthorization&type=).

If we truly don't need this then it should be removed, but this brings up a bigger issue if our TastyPie endpoints are properly checking for Auth or if they even need to be.

We've also had talk in the past about migrating all of our TastyPie endpoints to Django Rest Framework ones.

scottx611x:
Closed by #2822

PR #2822, opened 2018-06-15
Core API cleanup

hackdna:
Add unit tests for NodeResource, solve #2331, make make_api_uri() more robust, move core API tests and related code to test_api.py, code cleanup


refinery/data_set_manager/views.py
```diff
@@ -1018,3 +1018,41 @@ def post(self, request):
...
+                         node.file_uuid)
+            return HttpResponseNotFound()
+        except FileStoreItem.MultipleObjectsReturned:
+            logger.critical("Multiple FileStoreItems found with UUID '%s'",
+                            node.file_uuid)
+            return HttpResponseServerError()
+
+        file_store_item.delete()
+        node.file_uuid = None
+        node.save()
```
jkmarx: How are we handling updating solr, since the file browser grabs it info from assay/files -> solr. Relevant to adding files also.

refinery/data_set_manager/views.py
```diff
@@ -1018,3 +1018,41 @@ def post(self, request):
                 import_file.delay(file_store_item.uuid)
 
         return HttpResponse()  # 200
+
+
+class RemoveFileFromNodeView(View):
+    """Remove file from a Node instance"""
+    def post(self, request):
+        try:
```
jkmarx: maybe worth setting a variable node_uuid = request.POST['node_uuid'].  I think it would improve readability

refinery/data_set_manager/views.py
```diff
@@ -1018,3 +1018,41 @@ def post(self, request):
                 import_file.delay(file_store_item.uuid)
 
         return HttpResponse()  # 200
+
+
+class RemoveFileFromNodeView(View):
```
jkmarx: Add-files and remove-files are used to update a data-set (edit for file_uuid), should they be a patch to the data_set v2 api or the assay/files api, since we're modifying a data set or the node v2 api. To the user they are just updating a field. @scottx611x Thoughts?

refinery/data_set_manager/views.py
hackdna: Not sure what needs to happen with Solr. I think it happens automatically in case of adding the files.

refinery/data_set_manager/views.py
hackdna: I don't think introducing an extra variable that's used only once would increase readability.

refinery/data_set_manager/views.py
scottx611x: To me it would be intuitive to see this change as a patch in the v2 Node api if all we are changing is the `file_uuid` field

refinery/data_set_manager/views.py
scottx611x: A reindex would still need to happen here for the Node that is being altered. Otherwise there would still be a link provided in the FileBrowser (populated by the Solr response) pointing to the data file that has been removed here

hackdna:
Just to summarize:
* move this functionality to `core.api.NodeResource` and convert to handle `PATCH`
* update Solr index for Node instance
* add tests

Correct?

scottx611x:
@hackdna Correct 👍

Issue #2781, opened 2018-05-22
Perms Enhancement: Add Primary Group


jkmarx:
All users may designate a primary group
- [ ]  Update UI to designate the primary group
- [ ] Extend user model to include a primary group
- [ ] Extend user api to add primary group
     - validate user is a member of group
     - ? how do we handle the primary group when user leaves/deletes/removed from group
- [ ] Update user API to send primary group info
- [ ] Create DRF api to patch user profile

jkmarx:
![screen shot 2018-05-23 at 9 17 09 am](https://user-images.githubusercontent.com/9956764/40427146-dee13e3a-5e6a-11e8-8fb2-a34a9220d7d4.png)

PR #2823, opened 2018-06-18
Jkmarx/primary group api unit tests

jkmarx:
Ref #2781 
- Unit tests for the UserProfile V2 API

### DIFF
--- a/refinery/core/test_views.py
+++ b/refinery/core/test_views.py
@@ -21,7 +21,8 @@ from factory_boy.utils import create_dataset_with_necessary_models
 
 from .models import (Analysis, DataSet, ExtendedGroup, Project,
                      Workflow, WorkflowEngine)
-from .views import AnalysesViewSet, DataSetsViewSet, WorkflowViewSet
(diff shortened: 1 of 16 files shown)


refinery/core/test_views.py
```diff
@@ -934,3 +935,78 @@ def test_get_workflow_graph(self):
...
+        super(UserProfileApiV2Tests, self).setUp(
+            api_base_name="user_profiles/",
+            view=UserProfileViewSet.as_view()
+        )
+        self.user_lm = User.objects.create_user('lab_member',
+                                                'member@fake.com',
+                                                self.password)
+        self.lab_group = ExtendedGroup.objects.create(name="Lab Group")
+        self.non_lab_group = ExtendedGroup.objects.create(name="Test Group")
+        self.lab_group.group_ptr.user_set.add(self.user_lm)
```
scottx611x: I think you can access `user_set` directly without going through `group_ptr`:
`self.lab_group.user_set...`

PR #2824, opened 2018-06-18
Scottx611x/update dashboard to use events api

scottx611x:
This PR gets us this far:

screen shot 2018-06-18 at 3 27 57 pm

@jkmarx There was some debate as to where/how the `Event` messages should be created with the proper html links which is still up for discussion. Lets talk later


jkmarx:
@scottx611x This looks almost done. Let me fix the HTML in this pull request and add a unit test for the new service.

PR #2838, opened 2018-06-29
Jkmarx/connect add file service

jkmarx:
Ref #2700 
- Convert API to DRF
- Add service to client
- Update file upload code to handle adding data file
- To Do: DRF unit tests in separate pull request
screen shot 2018-06-29 at 10 24 47 am
screen shot 2018-06-29 at 10 24 58 am
screen shot 2018-06-29 at 10 25 23 am


refinery/data_set_manager/views.py
```diff
@@ -992,21 +992,24 @@ def put(self, request, uuid, format=None):
...
+class AddFilesToDataSetView(APIView):
     """Add file(s) to an existing data set from upload directory or bucket"""
+    http_method_names = ['post']
+
     def post(self, request):
         try:
-            data_set = DataSet.objects.get(uuid=request.POST['data_set_uuid'])
+            data_set = DataSet.objects.get(uuid=request
+                                           .data.get('data_set_uuid'))
         except KeyError:
```
scottx611x: I don't think this KeyError has the potential to be raised anymore

jkmarx: @scottx611x Yes, that got updated when I wrote some unit tests.

Issue #2817, opened 2018-06-12
Implement support for removing files from a data set


hackdna:
It should be possible to delete files in a data set using the data set browser.

* implement as a PATCH to Node DRF API in `core.views.py`
* update Solr index for Node instance (move this to Node.save() or post_save signal handler?)
* add tests

Blocked by #2834

jkmarx:
Resolved by pull request #2844.

PR #2844, opened 2018-07-05
Jkmarx/remove unused igv stuff

jkmarx:
- Resolves #2841 
- Reference #2817 

Feel free to test using swagger.

@hackdna Unsure the status of the remove link, but this api should be able to support it.

@scottx611x Unit test will come in the next pull request, if this works for Ilya. If it doesn't, then we'll just remove the Node V2 API.


refinery/core/views.py
```diff
@@ -693,13 +696,79 @@ def graph(self, request, *args, **kwargs):
...
+        data_set_owner = node.study.get_dataset().get_owner()
+
+        if data_set_owner == request.user:
+            serializer = NodeSerializer(node, data=request.data, partial=True)
+            if serializer.is_valid():
+                serializer.save()
+
+                try:
+                    FileStoreItem.objects\
+                        .get(uuid=old_file_uuid).delete_datafile()
```
scottx611x: I don't think we would want to delete the datafile on every patch, correct? I would think that would only be necessary if the old file_uuid differed from the new one

jkmarx: Yes, correct. I'll fix that.

jkmarx: FYI, this api is only supporting what we need which right now is removing file_uuid. It does not support adding or changing.

refinery/core/views.py
```diff
@@ -693,13 +696,79 @@ def graph(self, request, *args, **kwargs):
         )
 
 
-class NodeViewSet(viewsets.ModelViewSet):
```
scottx611x: What was the need for the `ModelViewSet` to `APIView` change?

refinery/core/serializers.py
```diff
@@ -109,100 +107,30 @@ class Meta:
...
-            'is_auxiliary_node',
-            'file_extension',
-            'auxiliary_file_generation_task_state',
-            'ready_for_igv_detail_view',
-            'file_uuid'
-        ]
+        fields = ['uuid', 'file_uuid']
+
+    def validate_file_uuid(self, file_uuid):
+        if file_uuid != 'None':
```
scottx611x: Are we setting `Node.file_uuid` to the string `'None'` by default? That seems off to me. 
Maybe you're looking for: `if file_uuid is not None:`?

refinery/core/views.py
scottx611x: I was more so curious about the underlying need to switch from one to the other

refinery/core/views.py
```diff
@@ -693,13 +696,80 @@ def graph(self, request, *args, **kwargs):
...
+              description: uuid for file store item
+              type: string
+              paramType: form
+              required: false
+    ...
+    """
+    http_method_names = ['get', 'patch']
+
+    def update_file_store(self, old_uuid, new_uuid):
+        if new_uuid is None:
```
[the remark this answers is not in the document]
jkmarx: Even though the api is only supporting file_uuid deletion, this is needed to avoid future issues when the file_uuid is not changed and other fields are.

refinery/core/views.py
```diff
@@ -693,13 +696,80 @@ def graph(self, request, *args, **kwargs):
...
+              paramType: path
+              required: true
+            - name: file_uuid
+              description: uuid for file store item
+              type: string
+              paramType: form
+              required: false
+    ...
+    """
+    http_method_names = ['get', 'patch']
```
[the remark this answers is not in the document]
jkmarx: Not a viewset.

PR #2850, opened 2018-07-10
Jkmarx/update serializers fields syntax

jkmarx:
- Update syntax for fields and exclude fields based on docs. The brackets did not cause any issues but updated based on docs.
http://www.django-rest-framework.org/api-guide/serializers/


jkmarx:
all fixed @scottx611x

PR #2854, opened 2018-07-11
Jkamrx/fix node api remove datafile

### LINKED ISSUE #2700: Metadata/Data File Editing for Existing DataSets

jkmarx:
Resolves #2700 
During local manual tests:
- Issues with removing & adding files
  Fixed: We should not remove the node's file_uuid. It needs the file_store_item uuid to future uploads. 1) Keep delete datafile 2) Remove import_task_id -> update solr index to 'N/A'

@hackdna @scottx611x Since we're not updating the node object, should we move this to a file_store_item v2 api? I'd remove the node v2 api all together, since we're not using it.
@jkmarx The short answer is probably yes, the long answer is here: https://github.com/refinery-platform/refinery-platform/pull/2854#issuecomment-404239106

- Added datafile index to prevent remove for remote download datafiles
screen shot 2018-07-12 at 10 52 37 pm

@hackdna Should we do anything with the celery task object? 
@jkmarx `FileStoreItem.delete_datafile()` takes care of that

### DIFF
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -730,7 +730,6 @@ class NodeViewSet(APIView):
                 logger.error(e)
             if file_store_item:
                 file_store_item.delete_datafile()
-                file_store_item.delete_file_import_task()
 
     def get_object(self, uuid):
         try:
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -748,6 +748,13 @@ class NodeViewSet(APIView):
         new_file_uuid = request.data.get('file_uuid')
         data_set = node.study.get_dataset()
 
+        if not data_set.is_clean():
+            return Response(
+                'Files cannot be removed once an analysis or visualization '
+                'have ran on a data set ',
+                status=status.HTTP_400_BAD_REQUEST
+            )
+
         if data_set.get_owner() == request.user:
             # to remove the data file, we need to delete it and update index,
             #  the file store item uuid should remain
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -724,10 +724,13 @@ class NodeViewSet(APIView):
     def update_file_store(self, old_uuid, new_uuid):
         if new_uuid is None:
             try:
-                FileStoreItem.objects.get(uuid=old_uuid).delete_datafile()
+                file_store_item = FileStoreItem.objects.get(uuid=old_uuid)
             except (FileStoreItem.DoesNotExist,
                     FileStoreItem.MultipleObjectsReturned) as e:
                 logger.error(e)
+            if file_store_item:
+                file_store_item.delete_datafile()
+                file_store_item.delete_file_import_task()
 
     def get_object(self, uuid):
         try:
@@ -753,20 +756,16 @@ class NodeViewSet(APIView):
 
     def patch(self, request, uuid):
         node = self.get_object(uuid)
-        old_file_uuid = node.file_uuid
+        new_file_uuid = request.data.get('file_uuid')
         data_set_owner = node.study.get_dataset().get_owner()
 
         if data_set_owner == request.user:
-            serializer = NodeSerializer(node, data=request.data, partial=True)
-            if serializer.is_valid():
-                serializer.save()
-                self.update_file_store(old_file_uuid, node.file_uuid)
-                NodeIndex().update_object(node, using="data_set_manager")
-                return Response(
-                    serializer.data, status=status.HTTP_202_ACCEPTED
-                )
+            # to remove the data file, we need to delete it and update index,
+            #  the file store item uuid should remain
+            self.update_file_store(node.file_uuid, new_file_uuid)
+            NodeIndex().update_object(node, using="data_set_manager")
             return Response(
-                serializer.errors, status=status.HTTP_400_BAD_REQUEST
+                NodeSerializer(node).data, status=status.HTTP_202_ACCEPTED
             )
 
         return Response(uuid, status=status.HTTP_401_UNAUTHORIZED)
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -751,7 +751,7 @@ class NodeViewSet(APIView):
         if not data_set.is_clean():
             return Response(
                 'Files cannot be removed once an analysis or visualization '
-                'have ran on a data set ',
+                'has ran on a data set ',
                 status=status.HTTP_400_BAD_REQUEST
             )
 
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -752,7 +752,7 @@ class NodeViewSet(APIView):
             serializer = NodeSerializer(node)
             return Response(serializer.data)
 
-        return Response(node, status=status.HTTP_401_UNAUTHORIZED)
+        return Response(uuid, status=status.HTTP_401_UNAUTHORIZED)
 
     def patch(self, request, uuid):
         node = self.get_object(uuid)
@@ -952,7 +952,7 @@ class DataSetsViewSet(APIView):
             )
         else:
             return Response(
-                self.data_set, status=status.HTTP_401_UNAUTHORIZED
+                uuid, status=status.HTTP_401_UNAUTHORIZED
             )
 
     def send_transfer_notification_email(self, old_owner, new_owner,
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -771,7 +771,7 @@ class NodeViewSet(APIView):
 
             NodeIndex().update_object(node, using="data_set_manager")
             return Response(
-                NodeSerializer(node).data, status=status.HTTP_202_ACCEPTED
+                NodeSerializer(node).data, status=status.HTTP_200_OK
             )
 
         return Response(uuid, status=status.HTTP_401_UNAUTHORIZED)
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -721,16 +721,6 @@ class NodeViewSet(APIView):
     """
     http_method_names = ['get', 'patch']
 
-    def update_file_store(self, old_uuid, new_uuid):
-        if new_uuid is None:
-            try:
-                file_store_item = FileStoreItem.objects.get(uuid=old_uuid)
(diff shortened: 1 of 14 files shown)


scottx611x:
I think it makes sense to keep this functionality in a `Node` API. The mentioned `FileStoreItem` v2 API would still have to know about a `Node` to update its corresponding Solr index entry. I think that, given the hierarchy, its more appropriate for `Nodes` to know about and act upon `FileStoreItems` rather than the other way around.

hackdna:
The issue is that the existing data model is not designed to support file removal. Here's an outline for a fix:

* replace `file_uuid` attribute of `Node` with a foreign key to `FileStoreItem`: `Node.data_file = ForeignKey(FileStoreItem, null=True, unique=True, on_delete=SET_NULL)`
* add Solr update logic to `Node.save()`
* move the `source` attribute from `FileStoreItem` to `Node`

This would make file operations simpler and more logical:

Removing data files: `DELETE` request to `FileStoreItem` API. This will automatically stop the file import task (if any) then set `Node.data_file` to null and update Solr when the corresponding `Node` instance is saved after update

Adding data files:` POST` request to `AddFilesToDataSetView`
This will be very similar to the existing implementation: iterate over file `Node` instances in the data set, create corresponding `FileStoreItem` instances and start file import tasks
or
for each new file: `PATCH` request to Node API (to create a new FileStoreItem instance, associate with the `Node` instance and start file import task)

This also makes it easier to move files from one Node to another in case of metadata updates (simply associate the `FileStoreItem` instance with a different `Node`) and  would allow for much simplified `file_store` app operations in general.

scottx611x:
This sounds like a good approach, but I think there are some things to consider.

When manipulating the fields on `FileStoreItem` and `Nodes` there will need to be some thought into the data migrations that need to be introduced to update existing `Nodes` and `FileStoreItems` in our system.

The majority of `Nodes` don't encapsulate files, so it wasn't clear to me why a `source` field would move there.

Overall, this seems like quite a bit of work to tack on to the end of this milestone, but I could be wrong.

refinery/core/views.py
```diff
@@ -754,20 +755,16 @@ def get(self, request, uuid):
...
-                return Response(
-                    serializer.data, status=status.HTTP_202_ACCEPTED
-                )
+            # to remove the data file, we need to delete it and update index,
+            #  the file store item uuid should remain
+            self.update_file_store(node.file_uuid, new_file_uuid)
+            NodeIndex().update_object(node, using="data_set_manager")
             return Response(
-                serializer.errors, status=status.HTTP_400_BAD_REQUEST
+                NodeSerializer(node).data, status=status.HTTP_202_ACCEPTED
```
hackdna: All the operations above are synchronous, so HTTP 200 status is appropriate here.

refinery/core/views.py
```diff
@@ -728,7 +728,8 @@ def update_file_store(self, old_uuid, new_uuid):
             except (FileStoreItem.DoesNotExist,
                     FileStoreItem.MultipleObjectsReturned) as e:
                 logger.error(e)
-            file_store_item.delete_datafile()
+            if file_store_item:
```
hackdna: `else:` is more appropriate here

hackdna:
Data migrations would simply need to have some extra code to copy and update information in the corresponding fields.

The source should be a part of Node to avoid losing that information when a corresponding FileStoreItem instance is deleted. Also, there is already some of file-related functionality in the Node model, so adding a source field is not really an issue. Alternatively, there can be a FileNode model derived from Node that would encapsulate file-related functionality.

This is a bit of work but it is likely we will keep running into this limitation down the road.

refinery/core/views.py
```diff
@@ -754,20 +755,16 @@ def get(self, request, uuid):
...
-                self.update_file_store(old_file_uuid, node.file_uuid)
-                NodeIndex().update_object(node, using="data_set_manager")
-                serializer.save()
-                return Response(
-                    serializer.data, status=status.HTTP_202_ACCEPTED
-                )
+            # to remove the data file, we need to delete it and update index,
+            #  the file store item uuid should remain
+            self.update_file_store(node.file_uuid, new_file_uuid)
+            NodeIndex().update_object(node, using="data_set_manager")
```
hackdna: This statement is now repeated several times throughout the codebase and thus should be moved to `Node.save()` or similar.

refinery/core/views.py
```diff
@@ -754,20 +755,16 @@ def get(self, request, uuid):
...
-            if serializer.is_valid():
-                self.update_file_store(old_file_uuid, node.file_uuid)
-                NodeIndex().update_object(node, using="data_set_manager")
-                serializer.save()
-                return Response(
-                    serializer.data, status=status.HTTP_202_ACCEPTED
-                )
+            # to remove the data file, we need to delete it and update index,
+            #  the file store item uuid should remain
+            self.update_file_store(node.file_uuid, new_file_uuid)
```
hackdna: If this is a method on Node model then it could also handle Solr index update.

refinery/core/views.py
hackdna: Also, is `file_uuid` actually updated in the database?

jkmarx: @hackdna No, see pull request comments above.

refinery/core/views.py
jkmarx: For this limited api, the Node model is not updated. (Comments above). Technical debt issue perhaps.

jkmarx:
@scottx611x I made some changes to this pull request including adding the datafile index and updating the UI for it. Would you please re-review it?

refinery/core/views.py
```diff
@@ -754,20 +745,33 @@ def get(self, request, uuid):
...
+                    )
+                except (FileStoreItem.DoesNotExist,
+                        FileStoreItem.MultipleObjectsReturned) as e:
+                    logger.error(e)
+                else:
+                    file_store_item.delete_datafile()
+
+            NodeIndex().update_object(node, using="data_set_manager")
+            return Response(
+                NodeSerializer(node).data, status=status.HTTP_202_ACCEPTED
```
scottx611x: I think an `HTTP 200` would be more appropriate here, `202` would be the case if we called a celery task or something else async

refinery/data_set_manager/search_indexes.py
```diff
@@ -153,6 +154,7 @@ def prepare(self, node):
 
         filetype = "" if file_store_item is None else file_store_item.filetype
         data.update({
+            NodeIndex.DATAFILE: node.get_file_store_item().datafile,
```
scottx611x: I think this is troublesome for many `Nodes`. The majority of `Nodes` don't have `FileStoreItems`, which would yield `AttributeErrors` when trying to access `.datafile` on a `None`.

I think adding a few tests to `data_set_manager.tests.NodeIndexTests` regarding what `NodeIndex.DATAFILE` should contain based on at least the two scenarios (`Node` has/doesn't have a `FileStoreItem`) would be a good idea

jkmarx:
@scottx611x Can you please take another look?

refinery/core/views.py
```diff
@@ -754,20 +745,33 @@ def get(self, request, uuid):
...
+                    file_store_item = FileStoreItem.objects.get(
+                        uuid=node.file_uuid
+                    )
+                except (FileStoreItem.DoesNotExist,
+                        FileStoreItem.MultipleObjectsReturned) as e:
+                    logger.error(e)
+                else:
+                    file_store_item.delete_datafile()
+
+            NodeIndex().update_object(node, using="data_set_manager")
```
scottx611x: With the changes I just pushed to `scottx611x/pending_import_state_for_ui_datafile_addition` you could do: `node.update_solr_index()` here instead.

refinery/data_set_manager/search_indexes.py
```diff
@@ -151,8 +152,11 @@ def prepare(self, node):
             if type(value) is set:
                 data[key] = " + ".join(str(i) for i in sorted(value))
 
+        datafile = "" if file_store_item is None else file_store_item.datafile
```
scottx611x: Solr is expecting strings for index preparation,  but `FileStoreItem.datafile` returns `FieldFile` instances. 

I believe you would want: `datafile = file_store_item.datafile.name` here instead

refinery/core/views.py
jkmarx: I'll update once we merge your pull request.

PR #2855, opened 2018-07-11
Upgrade django-docker-engine

### LINKED ISSUE #2847: UI exposure for Docker logs
As of 0.0.59, django_docker_engine makes the Docker logs available. This will probably be most useful when the visualization never leaves the please-wait page, though it could help diagnose problems after startup. (Though those seem to be much less common.)

I think there are two tasks here:
- Expose the information now available through the API in the UI... probably just a bare page containing the logs in the pre-block.
- Link to this page from the please-wait page, with appropriate messaging.

mccalluc:
Fix #2847. Not adding a link in the UI, since this is really just a debugging tool.

### DIFF
--- a/refinery/config/settings/base.py
+++ b/refinery/config/settings/base.py
@@ -220,7 +220,7 @@ INSTALLED_APPS = (
     'rest_framework',
     'rest_framework_swagger',
     'django_docker_engine',
-    'httpproxy',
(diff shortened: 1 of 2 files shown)


mccalluc:
We're closing this because exposing docker logs is a potential security concern.

Issue #2259, opened 2017-10-23
Data set link from cross-dataset browser is missing ID, sometimes


mccalluc:
When Sarah is looking at the cross-dataset browser, the dataset links do not work: The ID is missing from the url. There should be a check that only displays these links when they are valid. Perhaps looking at out-of-date code? b301acb2e32eb5f04cecd7264188154fbfaf5f2c

mccalluc:
When I hit their [api endpoint](https://beta.genometranslationcommons.org/api/v2/files/?fq=&limit=100&sort=), I get back:
```
...
    "nodes": [
        {
            "django_id": "456672",
            "name": "BM_10_609",
            "assay_uuid": "3bbffb2a-6d2e-4253-b99e-83ecfd9bc1ad",
            "file_uuid": "2fdfe413-a304-4a13-8360-57ae69a4c84a",
            "study_uuid": "2bc08eb3-3210-48c8-a01a-38abebcb4ae9",
            "type": "Raw Data File"
        },
...
```

At the very least, for the link to work, it needs a `data_set_uuid`. @drj11 : Would you be able to reindex genometranslationcommons? Or is that not the issue?

gmnelson:
assay_uuid
:
"e4789532-b7a0-4960-9d41-399cc2709854"
django_id
:
"144912"
file_uuid
:
"f2c923e6-1e81-4507-9179-5a5946e4a133"
filename_Characteristics_generic_s
:
"s5_p42_E2_45min.subsample.fastq"
name
:
"s5_p42_E2_45min.subsample.fastq"
organism_Characteristics_generic_s
:
"Human"
sample_id_Characteristics_generic_s
:
"s5_p42_E2_45min.subsample"
study_uuid
:
"1f0c5593-d046-4c8f-850a-d2ed929a6252"
type
:
"Raw Data File"

mccalluc:
Again, data set UUID is missing. Why is it not in solr? To debug, try a reindex: is it failing at that point, or before, or after?

gmnelson:
This seems to be a problem common to all data sets that have recently been uploaded, including: tabular metadata w/ and w/o files, ISA-Tab data set that has been shared with me. I haven't found a recent data set where its files are properly linked (i.e. have a data set UUID in the URL) in the cross-dataset browser.

PR #2857, opened 2018-07-12
Jkmarx/add missing UUID

jkmarx:
Resovles #2259  

redirect-fb

screen shot 2018-07-11 at 8 24 58 pm

### DIFF
--- a/refinery/data_set_manager/tasks.py
+++ b/refinery/data_set_manager/tasks.py
@@ -215,7 +215,7 @@ def create_dataset(investigation_uuid, username, identifier=None, title=None,
         user = User.objects.create_user(username, "", "test")
     if investigation_uuid is None:
         return None  # TODO: make sure this is never happens
-    annotate_nodes(investigation_uuid)
+
     dataset = None
     investigation = Investigation.objects.get(uuid=investigation_uuid)
     if identifier is None:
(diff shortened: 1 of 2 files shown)


refinery/data_set_manager/tasks.py
```diff
@@ -253,6 +253,7 @@ def create_dataset(investigation_uuid, username, identifier=None, title=None,
     if public:
         public_group = ExtendedGroup.objects.public_group()
         dataset.share(public_group)
+    annotate_nodes(investigation_uuid)
```
jkmarx: Relocate annotation after dataset is created

jkmarx:
@scottx611x Sorry I meant to add you since you've recently touched the search_indexes

refinery/data_set_manager/tasks.py
scottx611x: If we need to `annotate_nodes()` whenever a dataset is created, then I think we could move this call to the post_save signal handler within an if like:

```
if kwargs["created"]:
    investigation = instance.get_investigation()
    if investigation is not None:
        annotate_nodes(investigation.uuid)
```

jkmarx: Talked in person, confusion about where the investigation is set (post creation). All set.

hackdna: Overriding `save()` is generally recommended since neither `pre_save` nor `post_save` are called during bulk operations anyway and it allows keeping model update code in one place.
https://docs.djangoproject.com/en/1.8/topics/db/models/#overriding-predefined-model-methods

PR #2858, opened 2018-07-12
Jkmarx/remove public group anon user from api

jkmarx:
- Dashboard Collaboration Card -> remove AnonymousUser from member_list api response
- Add ordering in UI by last_name
- Fixes a UI bug related to empty columns/divs being ignored
Bad: 
screen shot 2018-07-12 at 9 36 43 am
Fixed:
screen shot 2018-07-12 at 12 12 01 pm

### DIFF
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -1852,7 +1852,10 @@ class ExtendedGroupResource(ModelResource):
                 'last_name': u.last_name,
                 'is_manager': self.user_authorized(u, ext_group)
             },
-            ext_group.user_set.all().filter(is_active=True))
+            ext_group.user_set.all().exclude(
+                is_active=False, username='AnonymousUser'
+            )
+        )
 
     # Override ORM methods for customization.
     def obj_get(self, bundle, **kwargs):
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -19,6 +19,7 @@ from django.contrib.sites.models import get_current_site
 from django.core.cache import cache
 from django.core.mail import EmailMessage
 from django.forms import ValidationError
+from django.db.models import Q
 from django.template import loader
 from django.utils import timezone
 
@@ -1852,9 +1853,8 @@ class ExtendedGroupResource(ModelResource):
                 'last_name': u.last_name,
                 'is_manager': self.user_authorized(u, ext_group)
             },
-            ext_group.user_set.all().exclude(
-                is_active=False, username='AnonymousUser'
-            )
+            ext_group.user_set.all().exclude(Q(is_active=False) |
+                                             Q(username='AnonymousUser'))
         )
 
     # Override ORM methods for customization.
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -19,7 +19,6 @@ from django.contrib.sites.models import get_current_site
 from django.core.cache import cache
 from django.core.mail import EmailMessage
 from django.forms import ValidationError
-from django.db.models import Q
 from django.template import loader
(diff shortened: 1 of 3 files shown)


refinery/ui/source/js/dashboard/partials/collaboration-card.html
```diff
@@ -68,7 +68,7 @@ <h3>
            aria-labelledby="headingOne"
            data-parent="#accordion">
             <ul class="list-unstyled dashboard-member-list">
-              <li ng-repeat="member in group.member_list">
+              <li ng-repeat="member in group.member_list | orderBy: last_name">
```
jkmarx: Typically I like the API to handle sorting (it's faster) but decided to it here until we decide what order we'd like. Last_name seems like a good default.

jkmarx: Never mind, after talking to Scott, it's really a client side filter.

refinery/core/api.py
```diff
@@ -1852,7 +1853,9 @@ def get_member_list(self, ext_group):
                 'last_name': u.last_name,
                 'is_manager': self.user_authorized(u, ext_group)
             },
-            ext_group.user_set.all().filter(is_active=True))
+            ext_group.user_set.all().exclude(Q(is_active=False) |
```
jkmarx: https://docs.djangoproject.com/en/2.0/ref/models/querysets/#methods-that-return-new-querysets

jkmarx: "If you need to execute more complex queries (for example, queries with OR statements), you can use Q objects."

refinery/ui/source/js/dashboard/partials/collaboration-card.html
```diff
@@ -32,43 +32,46 @@ <h3>
...
+             <span ng-if="group.name!=='Public'">
+                &nbsp;{{ group.name }}
              </span>
-             &nbsp;{{ group.name }}
            </a>
             <div class="col-md-4 col-xs-4">
                <span ng-if="group.can_edit">Manager</span>
+               <span ng-if="!group.can_edit"> &nbsp;</span>
             </div>
             <div class="col-md-2 col-xs-3">
```
jkmarx: Just fixed some padding issues.

refinery/ui/source/js/dashboard/partials/collaboration-card.html
```diff
@@ -32,43 +32,46 @@ <h3>
...
+                     refinery-tooltip-container="body"
+                     refinery-tooltip-placement="right"
+                     title="Invite user to group">
+                     <i class="fa fa-plus-circle fa-stack-1x fa-stack-right-corner"></i>
+                     <i class="fa fa-user-o"></i>&nbsp;
+                   </span>
+                </a>
               </span>
-              </a>
+              <span ng-if="group.name==='Public'"> &nbsp;</span>
```
jkmarx: Issue with empty divs/grid https://stackoverflow.com/questions/35283053/empty-columns-in-responsive-grid

refinery/core/api.py
scottx611x: I think you could avoid the use of `Q()` objects with a query like: `ext_group.user_set.all().filter(is_active=True).exclude(id=settings.ANONYMOUS_USER_ID)`

They're not bad to use, but I've generally only seen them used when the standard Django ORM falls short.

jkmarx: @scottx611x What advantage is there to chaining filter & exclude vs Q? Skimming the docs it appears Q would work well for this OR situation.

PR #2859, opened 2018-07-12
Scottx611x/properly associate metadata file for tabular imports

### LINKED ISSUE #2669: FileStoreItem pointing to tabular metadata file has incorrect source
* Specific code commit: 4e4c04b
* Environment where the error occurred: any

### Steps to reproduce
Import a [data set](http://data.cloud.refinery-platform.org.s3.amazonaws.com/data/sample-data/human-chipseq/hg19-metadata-local.txt)

### Observed behavior
* Four FileStoreItem instances are created that correspond to the four files in the data set
* An additional FileStoreItem instance created with the field `source` set to `blob`

### Expected behavior
Only four FileStoreItem instances created that correspond to the four files in the data set

### Notes
See also #2547

scottx611x:
Fixes #2669 

UI: 
screen shot 2018-07-12 at 5 08 30 pm

FileStoreItem:
screen shot 2018-07-12 at 4 41 00 pm


refinery/data_set_manager/management/commands/process_metadata_table.py
```diff
@@ -127,24 +127,21 @@ def handle(self, *args, **options):
...
-                    is_public=options['is_public'],
-                    delimiter=options['delimiter'],
-                    custom_delimiter_string=options['custom_delimiter_string']
-                )
-        except IOError as exc:
-            raise CommandError("Could not open file '%s': %s" %
-                               (options['file_name'], exc))
+            dataset_uuid = process_metadata_table(
+                username=options['username'], title=options['title'],
+                metadata_file_path=options['file_name'],
```
hackdna: Providing a file path instead of a file object complicates the code and makes this function less testable.

refinery/data_set_manager/single_file_column_parser.py
```diff
@@ -181,12 +184,10 @@ def run(self):
         # import in file as "pre-isa" file
         logger.info("trying to add pre-isa archive file %s",
                     self.metadata_file.name)
-        # FIXME: this will not create a FileStoreItem if self.metadata_file
-        # does not exist on disk (e.g., a file object like TemporaryFile)
         file_store_item = \
-            FileStoreItem.objects.create(source=self.metadata_file.name)
+            FileStoreItem.objects.create(source=self.metadata_file_path)
```
hackdna: The only change that is needed:
```python
file_store_item = FileStoreItem.objects.create(source=metadata_file.name)
file_store_item.datafile.save(os.path.basename(metadata_file.name),
                              django.core.files.File(metadata_file))
```
https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.fields.files.FieldFile.save

refinery/data_set_manager/single_file_column_parser.py
```diff
@@ -181,12 +184,10 @@ def run(self):
...
         investigation.pre_isarchive_file = file_store_item.uuid
-        import_file(investigation.pre_isarchive_file, refresh=True)
+        import_file.delay(investigation.pre_isarchive_file, refresh=True)
```
hackdna: Not needed if file is already saved above.

refinery/data_set_manager/views.py
```diff
@@ -442,6 +416,33 @@ def import_by_url(self, url):
         }
 
 
+def import_by_file(file):
```
hackdna: If this is moved out of the view than why not `import_by_url`?

refinery/data_set_manager/views.py
```diff
@@ -475,8 +476,9 @@ def post(self, request, *args, **kwargs):
             if data_set_ownership_response is not None:
                 return data_set_ownership_response
         try:
-            metadata_file = request.FILES['file']
             title = request.POST.get('title')
+            metadata_file = request.FILES['file']
+            metadata_file.name = request.POST.get('file_name')
```
hackdna: `metadata_file.name` should already have the name of the uploaded file:
https://docs.djangoproject.com/en/1.8/ref/files/uploads/#django.core.files.uploadedfile.UploadedFile.name

refinery/data_set_manager/views.py
```diff
@@ -487,7 +489,13 @@ def post(self, request, *args, **kwargs):
...
             else:
                 return render(request, self.template_name, error)
-
+        else:
+            response = import_by_file(metadata_file)
+            if not response["success"]:
+                error_message = response["message"]
+                logger.error(error_message)
+                raise RuntimeError(error_message)
+            metadata_file_path = response["data"]["temp_file_path"]
```
hackdna: Not necessary if uploaded file is used directly.

refinery/data_set_manager/views.py
hackdna: Why not move out `import_by_url()` too?

refinery/data_set_manager/views.py
hackdna: `metadata_file.name` should already contain uploaded file name: https://docs.djangoproject.com/en/1.8/ref/files/uploads/#django.core.files.uploadedfile.UploadedFile.name

refinery/data_set_manager/views.py
hackdna: This should not be needed if a file object is used.

refinery/data_set_manager/views.py
scottx611x: `metadata_file.name` is: `u'blob'` before this assignment

scottx611x: `metadata_file.name` is: `u'blob'` before this assignment

PR #2913, opened 2018-07-30
Jkmarx/node api v2 unit test

jkmarx:
- Add unit tests for node v2 api - patch

### DIFF
--- a/refinery/data_set_manager/test_views.py
+++ b/refinery/data_set_manager/test_views.py
@@ -1,5 +1,6 @@
 import mock
 import uuid
+from urlparse import urljoin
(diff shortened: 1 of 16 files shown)


refinery/data_set_manager/test_views.py
```diff
@@ -1,5 +1,6 @@
 import mock
 import uuid
+from urlparse import urljoin
```
scottx611x: Alphabetical import ordering

jkmarx:
@scottx611x No, empty string is fine. I want to be explicit and avoid the null conversion.

PR #2926, opened 2018-08-02
Fix bug, not checking every data set for validiness.

jkmarx:
- Fix bug related to not checking every data set for validiness

### DIFF
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -784,6 +784,7 @@ class DataSetsViewSet(APIView):
                     "DataSet with UUID: {} is invalid, and most likely is "
                     "still being created".format(data_set.uuid)
                 )
+                continue
             elif filter_requested:
                 if self.is_filtered_data_set(data_set, filters):
                     filtered_data_sets.append(data_set)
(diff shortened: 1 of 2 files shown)


refinery/core/views.py
```diff
@@ -774,25 +774,26 @@ def get(self, request):
...
+        filtered_data_sets = []
+        filter_requested = filters.get('is_owner') \
+            or filters.get('is_public') \
+            or filters.get('group')
+        for data_set in user_data_sets:
+            if not data_set.is_valid:
+                logger.warning(
+                    "DataSet with UUID: {} is invalid, and most likely is "
+                    "still being created".format(data_set.uuid)
+                )
```
scottx611x: I think a `continue` after this logger would be ideal

jkmarx: @scottx611x Why?

scottx611x: I guess it won't change any functionality right now, but its a good visual indicator that the current iteration is complete. If any logic was ever added under the `else`, but still within the loop, it would be hit for a DataSet that is invalid without it.

PR #2927, opened 2018-08-02
Remove use of MetadataTableView in favor of utility functions

scottx611x:
`MetadataTableView` was not actually a `View`

### DIFF
--- a/refinery/data_set_manager/views.py
+++ b/refinery/data_set_manager/views.py
@@ -175,13 +175,8 @@ def import_by_file(_file):
         _handle_uploaded_file(file_obj, temp_file_path)
     except IOError as exc:
         error_msg = "Error writing file to disk"
-        logger.error(
-            "%s. IOError: %s, file name: %s, error: %s.",
-            error_msg,
-            e.errno,
-            e.filename,
-            e.strerror
-        )
+        logger.error("%s. IOError: %s, file name: %s, error: %s.",
+                     error_msg, exc.errno, exc.filename, exc.strerror)
         return _error_message(error_msg)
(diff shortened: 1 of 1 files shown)


refinery/data_set_manager/views.py
```diff
@@ -169,72 +169,65 @@ def clean(self):
                 "Please provide either a file or a URL")
 
 
-class MetaDataImportView(View):
+def import_by_file(_file):
```
hackdna: Something more explicit like `file_obj` instead of `_file`?

refinery/data_set_manager/views.py
```diff
@@ -169,72 +169,65 @@ def clean(self):
...
+    try:
+        _handle_uploaded_file(_file, temp_file_path)
+    except IOError as e:
+        error_msg = "Error writing file to disk"
+        logger.error(
+            "%s. IOError: %s, file name: %s, error: %s.",
+            error_msg,
+            e.errno,
+            e.filename,
+            e.strerror
```
hackdna: No extra line breaks necessary: `error_msg, exc.errno, exc.filename, exc.strerror`

refinery/data_set_manager/views.py
```diff
@@ -169,72 +169,65 @@ def clean(self):
                 "Please provide either a file or a URL")
 
 
-class MetaDataImportView(View):
+def import_by_file(_file):
+    temp_file_path = os.path.join(get_temp_dir(), _file.name)
+    try:
+        _handle_uploaded_file(_file, temp_file_path)
+    except IOError as e:
```
hackdna: `exc` instead of `e`

Issue #1936, opened 2017-07-21
What are "projects"?


mccalluc:
Looking through the code, I came across `Project`: There are URLs mapped, DB objects, and some html templates, but I have no idea what it is supposed to do. Any chance this is vestigial code which could be deleted?

mccalluc:
There are 33 private ones (no public) on https://test.stemcellcommons.org/statistics/

mccalluc:
To be clear, I think the conclusion is that we should get rid of them, unless the code is more deeply entangled than we think it is.

jkmarx:
Minimum we should get rid of the associated UI.

hackdna:
This a great reason to disentangle the code if it is.

scottx611x:
We also have `core.search_indexes.ProjectIndex` hanging around so we should check if the indexed data is used anywhere as well.

PR #2932, opened 2018-08-03
Jkmarx/remove projects template

jkmarx:
- Remove projects api, templates, and views.
- Ref #1936 
*Note the models and index remain. Removing them are more involved as they are entangled with other models.


jkmarx:
@mccalluc We do want to remove it as soon as the information is located elsewhere (dashboard or landing page.)

Issue #2909, opened 2018-07-27
Dashboard Data Set Performance Enhancement


jkmarx:
- Implement continous scrolling
    - [ ] Dashboard
    - [ ] History Panel
- [ ] Cache client side code (basic filters)

PR #2938, opened 2018-08-08
Jkmarx/data set v2 api filter enhancement

jkmarx:
https://django-guardian.readthedocs.io/en/stable/userguide/performance.html?highlight=performance

Ref #2909


refinery/core/serializers.py
```diff
@@ -24,17 +24,23 @@ class DataSetSerializer(serializers.ModelSerializer):
...
 
     def get_is_owner(self, data_set):
-        owner = data_set.get_owner()
         try:
-            user_request = self.context.get('request').user
-        except AttributeError as e:
-            logger.error("Request is missing a user: %s", e)
-            return False
-        return user_request == owner
+            data_set.is_owner
```
jkmarx: for the data set view api, it's faster to pass along a value because we are prefetching perms.

refinery/core/test_views.py
```diff
@@ -110,37 +110,6 @@ def test_get_data_set_with_auth(self):
     def test_get_data_set_with_anon_user(self):
         self.assertEqual(len(self.get_response.data.get('data_sets')), 0)
 
-    @mock.patch('core.views.DataSetsViewSet.is_filtered_data_set')
```
jkmarx: unit tests related to a helper method which has been removed

PR #2939, opened 2018-08-09
Default to primary group on loading.

jkmarx:
- Defaults to primary group when dashboard loads, if a primary group has been selected.


refinery/ui/source/js/dashboard/controllers/primary-group-button-ctrl.js
```diff
@@ -13,11 +13,13 @@
```
jkmarx: the child component is responsible for loading the initial dashboard's data sets

PR #2940, opened 2018-08-09
Disable buttons when user is not a member of a group.

jkmarx:
Bug - user who's not a member of any groups
screen shot 2018-08-09 at 7 51 15 am

Solution - Deactivate group buttons and expand popover
screen shot 2018-08-09 at 8 49 12 am

screen shot 2018-08-09 at 9 04 20 am

### DIFF
--- a/refinery/ui/source/js/dashboard/partials/primary-group-button.html
+++ b/refinery/ui/source/js/dashboard/partials/primary-group-button.html
@@ -13,18 +13,22 @@
       Primary Group: {{ $ctrl.primaryGroup.name }}
     </span>
   </button>
-  <button type="button" class="btn btn-primary btn-xs" uib-dropdown-toggle>
+  <button
+    type="button"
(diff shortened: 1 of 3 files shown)


refinery/ui/source/js/dashboard/partials/primary-group-button.html
```diff
@@ -34,7 +38,8 @@
     popover-placement="right"
     uib-popover-html="'This is your <strong>Primary Group</strong>. Once selected, data
     sets shared with this group will be displayed by default. To change your Primary Group
-    selection, use the drop down menu.'"
+    selection, use the drop down menu. To enable the group buttons, you must
+    be a member of a group.'"
```
mccalluc: This is getting long enough that I'm not sure how likely a user is to read the whole thing carefully... This part `To change your Primary Group selection, use the drop down menu.` seems kind of obvious: Perhaps it could be dropped to keep the whole thing more concise?

PR #2942, opened 2018-08-09
Scottx611x/configurable branch for vis tool registry

scottx611x:
@mccalluc 
`./manage.py load_tools --visualizations multiqc heatmap-scatterplot  --branch new_tool_versions`


refinery/tool_manager/tests.py
```diff
@@ -1281,6 +1281,27 @@ def test_load_tools_error_message_yields_vis_registry_info(
                 context.exception.message
             )
 
+    def test_load_tools_command_error_if_branch_and_workflows_specified(self):
+        with self.assertRaises(CommandError):
+            call_command("load_tools", workflows=True,
+                         branch="fake-branch-name")
```
mccalluc: And in a similar vein, if we were using more of the built-in machinery, I don't think we'd need explicit checks on their error handling.

refinery/tool_manager/management/commands/load_tools.py
```diff
@@ -41,6 +41,12 @@ def add_arguments(self, parser):
             help='Overwrite ToolDefinitions with the same names instead of '
                  'skipping their creation'
         )
+        parser.add_argument(
+            '--branch',
+            action='store',
+            dest='branch',
+            help='Branch from the registry to target for Vis tool installation'
+        )
```
mccalluc: Maybe make the "master" default explicit here, rather than the init? The help text will be more explicit, and you can get rid of one `if`

scottx611x:
@mccalluc sounds good to me!

PR #2944, opened 2018-08-10
Jkmarx/limit events api response

jkmarx:
- Limit the event api response to save time
- Fix bug in data set serializer (null response for is_owner and public)
- Update unit tests

Ref #2909

### DIFF
--- a/refinery/core/serializers.py
+++ b/refinery/core/serializers.py
@@ -25,7 +25,7 @@ class DataSetSerializer(serializers.ModelSerializer):
 
     def get_is_owner(self, data_set):
         try:
-            data_set.is_owner
+            return data_set.is_owner
         except:
             owner = data_set.get_owner()
             try:
@@ -34,17 +34,13 @@ class DataSetSerializer(serializers.ModelSerializer):
                 logger.error("Request is missing a user: %s", e)
                 return False
             return user_request == owner
-        else:
-            return data_set.is_owner
 
     def get_public(self, data_set):
         try:
-            data_set.public
+            return data_set.public
         except:
             is_public = data_set.is_public()
             return is_public
-        else:
-            return data_set.public
 
     def get_is_clean(self, data_set):
         return data_set.is_clean()
--- a/refinery/core/serializers.py
+++ b/refinery/core/serializers.py
@@ -34,6 +34,8 @@ class DataSetSerializer(serializers.ModelSerializer):
                 logger.error("Request is missing a user: %s", e)
                 return False
             return user_request == owner
+        else:
+            return data_set.is_owner
 
     def get_public(self, data_set):
         try:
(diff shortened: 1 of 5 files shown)


refinery/core/serializers.py
```diff
@@ -34,13 +34,17 @@ def get_is_owner(self, data_set):
                 logger.error("Request is missing a user: %s", e)
                 return False
             return user_request == owner
+        else:
```
jkmarx: fixes bug

refinery/core/test_views.py
```diff
@@ -952,16 +952,28 @@ def test_get_event_list(self):
 
         get_request = self.factory.get(urljoin(self.url_root, '/'))
         get_request.user = self.user
-        get_response = self.view(get_request).render()
-
+        force_authenticate(get_request, user=self.user)
+        get_response = self.view(get_request)
         self.assertEqual(
-            json.loads(get_response.content),
+            get_response.data,
```
jkmarx: reorder response (api now handles it)

refinery/core/serializers.py
mccalluc: Maybe just `return` in the first `try`? And you could do the same thing in the `except`:
```python
     def get_is_owner(self, data_set):
         try:
             return data_set.is_owner
         except:
             owner = data_set.get_owner()
             try:
                 return owner == self.context.get('request').user
             except AttributeError as e:
                logger.error("Request is missing a user: %s", e)
                return False
```

refinery/core/serializers.py
```diff
@@ -34,13 +34,17 @@ def get_is_owner(self, data_set):
...
+            return data_set.is_owner
 
     def get_public(self, data_set):
         try:
             data_set.public
         except:
             is_public = data_set.is_public()
             return is_public
+        else:
+            return data_set.public
```
mccalluc: ditto: You don't want to do too much in a `try`, but just returning isn't going cause a different exception

refinery/core/test_views.py
```diff
@@ -952,16 +952,27 @@ def test_get_event_list(self):
...
-                    'message': messages[0],
+                    'date_time': date_times[3],
+                    'message': messages[3],
                     'data_set': DataSetSerializer(
-                        data_sets[0], context={'request': get_request}
+                        data_sets[3], context={'request': get_request}
+                    ).data,
+                    'group': None,
+                    'user': UserSerializer(self.user).data,
+                    'type': 'UPDATE',
```
mccalluc: Originally the type was one of the CRUD verbs, so I'm surprised this isn't `CREATE`?

mccalluc: ... but it looks like it's just being copied from below, so not a new issue...

PR #2947, opened 2018-08-16
Bump django-docker version

mccalluc:


mccalluc:
I don't remember the details, but I think we might have been waiting to update this until HiGlass was sorted out?

PR #2955, opened 2018-08-28
Resolve bug and test due to solr format update.

jkmarx:
Bug due to the switching the solr request format


refinery/tool_manager/tests.py
```diff
@@ -290,12 +290,15 @@ def tearDown(self):
 
     def create_solr_mock_response(self, nodes):
         node_uuids = [n.uuid for n in nodes]
+        solr_params = _create_solr_params_from_node_uuids(node_uuids)
+        solr_params['json']['fields'] = solr_params['json']['filter']
+        solr_params['json'] = json.dumps(solr_params['json'])
```
mccalluc: This seems odd to me: Couldn't `_create_solr_params_from_node_uuids` just return data in the right shape? I should look at it in context and it might make more sense.

Issue #872, opened 2016-02-08
Remove django-celery


hackdna:
[Django supported out of the box](http://celery.readthedocs.org/en/latest/whatsnew-3.1.html#django-supported-out-of-the-box) in Celery 3.1.

hackdna:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;uPviGMWp&#x2F;199-improve-data-model-of-data-sets"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Improve data model of data sets</a>

ngehlenborg:
While this is fresh in your memory, can you please note here what the dependencies are that led you to update it in #871?

hackdna:
Basically, django-celery version must match Celery version.

mccalluc:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;bt60wOPA&#x2F;35-872"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> #872</a>

PR #2956, opened 2018-08-28
Update indexing logic to handle for unknown file_import tasks

scottx611x:
We should do something more elegant like utilizing celery's available signal: `task_published` signal, updating the state of published tasks to something we decide (`SENT`, `PUBLISHED` etc.) which would then allow us to treat the `PENDING` state that celery is assigning as a true "unknown" state. This would also lessen the `djcelery` package's grip on us.

To be able to achieve something like this, we first have to get our existing file import tasks in a good state (no pun intended) and properly categorized (Removal of false `PENDINGS`).

**Note: these changes warrant another run of: `./manage.py update_index`**

### DIFF
--- a/refinery/data_set_manager/search_indexes.py
+++ b/refinery/data_set_manager/search_indexes.py
@@ -197,11 +197,11 @@ def _get_download_url_or_import_state(file_store_item):
 
     # "PENDING" if an import_task_id doesn't exist and
     # there is no valid download_url
-    import_state = file_store_item.get_import_status()
     if not file_store_item.import_task_id:
         return celery.states.PENDING
 
     # "N/A" if the import_state is in a "READY_STATE" or "PENDING" with an
     # import_task_id and without a valid download_url
+    import_state = file_store_item.get_import_status()
     if import_state in {celery.states.PENDING} | celery.states.READY_STATES:
         return constants.NOT_AVAILABLE
--- a/refinery/data_set_manager/search_indexes.py
+++ b/refinery/data_set_manager/search_indexes.py
@@ -181,6 +181,17 @@ class NodeIndex(indexes.SearchIndex, indexes.Indexable):
 
 
 def _get_download_url_or_import_state(file_store_item):
+    """
+    Discerns the download url or file import state for a given FileStoreItem
+    :param file_store_item: A FileStoreItem instance
+    :returns <String>:
+        - a valid url pointing to the FileStoreItem's datafile
+        - constants.NOT_AVAILABLE if the FileStoreItem's import_state is in
+        celery.READY_STATES w/o a valid url available or if there is no
+        available task information for said FileStoreItem's import_file task
+        - celery.states.PENDING if a FileStoreItem's import_file task hasn't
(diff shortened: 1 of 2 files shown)


scottx611x:
@hackdna It hasn't properly worked since the that `TaskMeta` portion in the search_indexes code was removed. We just hadn't noticed the side effects until this morning.

The actual bug is due to the fact that celery will report a `PENDING` state for tasks (`import_file`) that are either truly `PENDING` or in an unknown state. @mccalluc noticed some Nodes that have been in the improper half of this `PENDING` state for awhile this morning.

I mention a better fix [here](https://github.com/refinery-platform/refinery-platform/pull/2956#issue-211465530), but that would still be dependent on this workaround getting applied first and getting our index back in the proper state.

hackdna:
It is still not clear to me what the bug is. What are the steps to reproduce, what is expected and what is actually happening?

scottx611x:
The bug:
Currently, without the `TaskMeta` check, the index will be improperly updated with `PENDING` for some`FileStoreItems` when it should be `N/A`.

Observed: 
https://test.stemcellcommons.org/data_sets/f27ce499-e6d7-494b-8190-48856c00be5c/#/files/

Celery will forget about task results after one day by default. Running an `AsyncResult()` after 24hrs will yield a `PENDING` state. 

```
>>> from core.models import DataSet
>>> d = DataSet.objects.get(uuid="f27ce499-e6d7-494b-8190-48856c00be5c")
>>> [f.get_import_status() for f in d.get_file_store_items()]
['PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING']
>>> d = DataSet.objects.get(uuid="7d51c027-e529-4c38-a8fb-8f1d46aad009")
>>> [f.get_import_status() for f in d.get_file_store_items()]
['PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING', 'PENDING']
```

We need some way to determine whether these `PENDING` states that are reported actually correspond to a known task or not.

Expected:
All of the "importing" (`"PENDING"`)  files from https://test.stemcellcommons.org/data_sets/f27ce499-e6d7-494b-8190-48856c00be5c/#/files/ should actually have lightning bolts (`"N/A"`) since they were never imported successfully.

scottx611x:
@hackdna Any thoughts on this?

hackdna:
Thanks for a detailed description. The problem is that even `TaskMeta` records are only kept for a month (not to mention that they won't even exist at all after [django-celery is removed](https://github.com/refinery-platform/refinery-platform/issues/872)), so a different solution is needed.

PR #2959, opened 2018-08-28
Jkmarx/fix perms api error

### LINKED ISSUE #2957: DataSet api v2 improperly yields `is_owner: true`
* Specific code commit: f6b5665211815395d5ee7246a3e3574b1692eb72
* Version of the web browser and OS: `Chrome OSX`
* Environment where the error occurred: https://test.stemcellcommons.org

### Steps to reproduce

1. Navigate to: https://test.stemcellcommons.org/api/v2/data_sets/

### Observed behavior
* For my user, the `is_owner` field in the response is `true` for all but 6 of the 147 `Datasets` I queried for

### Expected behavior
* For proper `DataSet` ownership to be reflected

### General Notes:
* This is currently affecting the new Dashboard
* The delete, share, and transfer ownership buttons are visible for `DataSet` as a side-effect of this

jkmarx:
Fix bug related to misunderstanding of perms, share vs add perms
@scottx611x I did confirm via locally and unit tests the perms change fixes the issue.

Resolves #2957

### DIFF
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -165,7 +165,7 @@ class SharableResourceAPIInterface(object):
             owned_res_set = Set(
                 get_objects_for_user(
                     user,
-                    'core.add_%s' %
+                    'core.share_%s' %
                     self.res_type._meta.verbose_name,
                     use_groups=False
                 ).values_list("id", flat=True))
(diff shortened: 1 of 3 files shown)


refinery/core/api.py
```diff
@@ -699,7 +699,7 @@ def get_by_db_id(self, request, **kwargs):
                 is_public = True
 
         # get_owner in core models uses add to distinguish owner
```
jkmarx: add_%s perms is used for checking ownership on non-shareable resources like analyses and workflows
data_set uses share_%s (only owners can share data_sets right now)

PR #2963, opened 2018-08-29
Fix sort bug due to syntax error.

jkmarx:
fix-sort

### DIFF
--- a/refinery/data_set_manager/utils.py
+++ b/refinery/data_set_manager/utils.py
@@ -711,6 +711,9 @@ def generate_solr_params(
             facet_fields_obj[facet]['excludeTags'] = facet.upper()
         filter_arr.extend(create_facet_filter_query(facet_filter))
 
+    if sort:
+        fixed_solr_params['sort'] = sort
+
     solr_params = {
         "json": {
             "query": 'django_ct:data_set_manager.node',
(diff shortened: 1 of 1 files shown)


refinery/data_set_manager/utils.py
```diff
@@ -711,6 +711,9 @@ def generate_solr_params(
             facet_fields_obj[facet]['excludeTags'] = facet.upper()
         filter_arr.extend(create_facet_filter_query(facet_filter))
 
+    if sort:
+        fixed_solr_params['sort'] = sort
```
mccalluc: Maybe just add another line when you define the dict up top? Hopefully nothing depends on the absence of the key in `fixed_solr_params`? So get rid of this `if` and instead:
```python
     fixed_solr_params = {
         "facet.limit": "-1",
         "fq": "is_annotation:{}".format(is_annotation),
         "rows": row,
         "start": start,
         "sort": params.get('sort'),
         "wt": "json"
}
```

Or maybe not: I might be missing something

jkmarx: @mccalluc I'd have to check again, but pretty sure there was weirdness if I included sort: None. Let's leave it for now, and I'll recheck later.

jkmarx: Note, this is the not fully supported Solr JSON REQUEST API in our version

mccalluc: Go with what you know works: Thanks!

PR #2966, opened 2018-08-29
Jkmarx/fix data upload

jkmarx:
- Semi solution to avoiding multiple files in file browsers add file modal

data-upload-bug

@mccalluc The add button looks disabled but it's not. There's a conflict btwn the angular directives and perhaps the library we're using. It'd require more time than I have at the moment to truly resolve. I believe this solution at least visually helps users.


jkmarx:
@mccalluc Only when the user is in the file-browser (hence the is_node_update). They are adding a file for a specific node.

PR #2971, opened 2018-08-29
Jkmarx/fix delete modal

jkmarx:
- Extend dataset v2 api return analyses and file_count 
- Fix hyper link
- Remove user icon


refinery/ui/source/js/dashboard/partials/dataset-delete-dialog.html
```diff
@@ -4,17 +4,11 @@ <h3>Delete Dataset</h3>
 <div class="modal-body">
   <span>
     <h5>
-       <i class="fa fa-user"
```
scottx611x: Why are we removing this?

jkmarx: I don't think there's a reason for the user icon. @mccalluc commented on it.   Did I miss it's purpose?

scottx611x: Ah I just looked and I think its redundant to have it. I just wasn't sure!

PR #2975, opened 2018-08-30
Avoid adding UUIDs of FileStoreItems that have already been imported

scottx611x:
Fixes the issue where [Analyses couldn't be run on FileStoreItems older than 1 day](https://github.com/refinery-platform/refinery-platform/issues/2791#issuecomment-417055639)

Towards a fix for: #2791

PR #2984, opened 2018-09-04
Scottx611x/add test coverage for refinery import tasks

scottx611x:
Some refactoring and added test coverage for the changes from: #2975

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -1524,7 +1524,7 @@ class Analysis(OwnableResource):
 
     def has_all_local_input_files(self):
         return all(file_store_item.is_local() for file_store_item in
-                   self._get_input_file_store_items())
+                   self.get_input_file_store_items())
 
     def _get_input_nodes(self):
         return [analysis_node_connection.node for analysis_node_connection in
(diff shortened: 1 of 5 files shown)


refinery/tool_manager/models.py
```diff
@@ -1153,6 +1153,20 @@ def get_galaxy_import_tasks(self):
             ) for file_store_item_uuid in self.get_input_file_uuid_list()
         ]
 
+    def get_refinery_import_tasks(self):
```
hackdna: This method should be a part of the `Analysis` model. Then you won't need to obtain a `Tool` object in `_refinery_file_import()`.
Also, technically, this method returns [task signatures](http://docs.celeryproject.org/en/3.1/userguide/canvas.html#signatures), so it should probably be called something like `get_refinery_import_task_signatures` to avoid ambiguity.

refinery/core/models.py
```diff
@@ -1682,6 +1683,20 @@ def _create_annotated_nodes(self):
...
 
+    def get_refinery_import_task_signatures(self):
+        """Create and return a list of import_file() task signatures for the
+        user-selected inputs of a WorkflowTool.launch()
+
+        NOTE: We avoid adding UUIDs of already imported FileStoreItem's as
+        to not re-import them without need.
+        """
+        return [
+            analysis_manager.tasks.import_file.subtask((file_store_item.uuid,))
```
hackdna: `import_file` is defined in `file_store.tasks`

PR #2985, opened 2018-09-04
Trim Tool's creation date in `display_name` to the second

scottx611x:


refinery/tool_manager/utils.py
```diff
@@ -222,8 +222,10 @@ def create_tool(tool_launch_configuration, user_instance):
 
     display_name = tool_launch_configuration.get("display_name")
     if not display_name:
-        display_name = "{} {} {}".format(tool.name, tool.creation_date,
-                                         tool.get_owner().username)
+        display_name = "{} {} {}".format(
+            tool.name, tool.creation_date.strftime('%m/%d/%Y %H:%M:%S'),
+            tool.get_owner().username
+        )
```
mccalluc: Consider just using one format string, and perhaps naming the slots, too:
```
>>> from datetime import datetime
>>> '{tool} {dt:%m/%d/%Y %H:%M:%S} {user}'.format(tool='tool', dt=datetime.now(), user='user')
'tool 09/04/2018 12:59:34 user'
```

... and maybe use the three-letter month name? and maybe some linking words? But maybe we want this to look more like an ID and less like a name?
```
>>> '{tool} started by {user} on {datetime:%b %d, %Y at %H:%M:%S}'.format(tool='tool', datetime=datetime.now(), user='user')
'tool started by user on Sep 04, 2018 at 13:08:25'
```

refinery/tool_manager/tests.py
```diff
@@ -3690,19 +3690,11 @@ def assertions(tool):
...
-            )
-            with self.assertRaises(VisualizationToolError) as context:
-                self._start_visualization(
-                    'hello_world.json',
-                    "https://www.example.com/file.txt",
-                )
-        self.assertIn("Max containers", context.exception.message)
+        with self.assertRaises(VisualizationTool.DoesNotExist):
+            self.create_tool(ToolDefinition.VISUALIZATION)
+        self.assertIn("Max containers", self.post_response.content)
```
mccalluc: Looks like the test is being weakened?

refinery/tool_manager/tests.py
```diff
@@ -1653,7 +1653,7 @@ def test_tool_launch_has_resonable_default_display_name(self):
         self.create_tool(tool_type=ToolDefinition.VISUALIZATION)
         self.assertEqual(self.tool.display_name, "{} {} {}".format(
```
mccalluc: I would prefer to have static string here, if possible, rather than recapitulating the logic from the code.

PR #2991, opened 2018-09-06
Scottx611x/tool manager views cleanup

scottx611x:
I came across this split (extra logic for handling `data_set_uuid` vs. `dataSetUuid`) the other day, and figured it would be good to take care of.

I've also gone ahead and refactored to remove some redundancies and unclear code placement.

### DIFF
--- a/refinery/tool_manager/views.py
+++ b/refinery/tool_manager/views.py
@@ -63,6 +63,8 @@ class ToolDefinitionsViewSet(ToolManagerViewSetBase):
             return ToolDefinition.objects.filter(
                 tool_type=ToolDefinition.VISUALIZATION
             )
+        else:
+            return []  # get_queryset should return an iterable
 
 
 class ToolsViewSet(ToolManagerViewSetBase):
--- a/refinery/tool_manager/views.py
+++ b/refinery/tool_manager/views.py
@@ -59,7 +59,7 @@ class ToolManagerViewSetBase(ModelViewSet):
             return Response(
                 "User is not authorized to access DataSet with UUID: {}"
                 .format(self.data_set.uuid),
-                status.HTTP_401_UNAUTHORIZED
+                status.HTTP_403_FORBIDDEN
             )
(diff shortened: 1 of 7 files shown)


refinery/tool_manager/views.py
```diff
@@ -80,69 +63,32 @@ def get_queryset(self):
...
-        return Response("User is not authorized to view visualizations.",
-                        status=status.HTTP_401_UNAUTHORIZED)
+        visualization_tools = list(
+            VisualizationTool.objects.filter(dataset=self.data_set)
+        )
+        workflow_tools = list(
+            WorkflowTool.objects.filter(dataset=self.data_set)
+        )
+        tool_type = self.request.query_params.get("tool_type")
+        if tool_type is None:
```
jkmarx: @scottx611x Any user can hit this endpoint and grab analyses or visualizations with a dataSetUuid. Thoughts?

refinery/ui/source/js/commons/services/tool-definitions.js
```diff
@@ -14,7 +14,7 @@
         query: {
           method: 'GET',
           isArray: true,
-          params: { dataSetUuid: '' }
+          params: { data_set_uuid: '' }
```
jkmarx: Sure, since it's a url query. Worth noting for json it's camelCase styling. https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format

refinery/tool_manager/views.py
scottx611x: Ah, this is just the `get_queryset` method. The permission logic happens in the respective `list()` methods

refinery/ui/source/js/commons/services/tool-definitions.js
scottx611x: Yeah I wasn't sold on one way vs. the other, but I can change the backend code to work with `dataSetUuid`

PR #2996, opened 2018-09-11
Scottx611x/fix add users to public group timeout

scottx611x:
While trying to deploy an AWS instance for `hotfix-1.6.6.1` I consistently ran into a timeout that didn't allow for the Puppet manifest to apply completely. 

This is due to the fact that we have a bunch of spam `User` instances hanging around and for each `user.save()` being called within the `add_users_to_public_group` mgmt command `_add_user_to_neo4j` is being run unnecessairily.

This PR only adds a user to neo4j on a `User.post_save` signal being hit if we have activated said user prior.

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -133,15 +133,16 @@ post_save.connect(create_user_profile, sender=User)
 
 
 @receiver(post_save, sender=User)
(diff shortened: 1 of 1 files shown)


scottx611x:
@mccalluc I've now been able to bring up a stack with this fix introduced

PR #3005, opened 2018-09-17
Hackdna/rewrite import file

### LINKED ISSUE #2359: Rewrite import_file task

hackdna:
Close #2359


refinery/file_store/tasks.py
```diff
@@ -1,371 +1,256 @@
...
+        file_store_name = storage.get_name(os.path.basename(source_path))
+        file_store_path = storage.path(file_store_name)
+
+        if source_path.startswith((settings.REFINERY_DATA_IMPORT_DIR,
+                                   get_temp_dir())):
+            move_file(source_path, file_store_path)
+        else:
+            if symlink:
+                symlink_file(source_path, file_store_path)
+            else:
```
jkmarx: Can we combine these into one if/elif/else? if source.....elif symlink....else

PR #3006, opened 2018-09-18
Scottx611x/add recaptcha to registration form

scottx611x:
This adds a Google reCAPTCHA v2 to our user registration form to combat the spam accounts we've been accumulating.

screen shot 2018-09-18 at 4 00 21 pm

TODO:

- [x] Provide instructions as to how one should acquire the API keys and configure Refinery to use them for local/AWS deployments -> **See [instructions here](https://github.com/refinery-platform/refinery-platform/wiki/Google-reCAPTCHA#acquiring-the-keys)**
- [x] The Cypress test for user-registration fails now with these changes, so we'll need to figure out a fix for that


refinery/core/test_views.py
```diff
@@ -1089,3 +1092,47 @@ def test_get_event_list(self):
...
+                "username": username,
+                "email": "test@example.com",
+                "first_name": "test",
+                "last_name": "user",
+                "affiliation": "Test Users",
+                "password1": password,
+                "password2": password
+            }
+        )
+        self.assertTrue(response.wsgi_request.recaptcha_is_valid)
```
hackdna: Why wsgi_request and not request?

refinery/core/test_views.py
```diff
@@ -1089,3 +1092,47 @@ def test_get_event_list(self):
...
+        self.assertIsNotNone(User.objects.get(username=username))
+
+    @override_settings(REFINERY_GOOGLE_RECAPTCHA_SITE_KEY="invalid_site_key")
+    @override_settings(
+        REFINERY_GOOGLE_RECAPTCHA_SECRET_KEY="invalid_secret_key"
+    )
```
hackdna:
```python
@override_settings(REFINERY_GOOGLE_RECAPTCHA_SITE_KEY="invalid_site_key",
                   REFINERY_GOOGLE_RECAPTCHA_SECRET_KEY="invalid_secret_key")
```

refinery/core/test_views.py
```diff
@@ -1089,3 +1092,47 @@ def test_get_event_list(self):
...
+                "last_name": "user",
+                "affiliation": "Test Users",
+                "password1": password,
+                "password2": password
+            }
+        )
+        self.assertEqual(400, response.status_code)
+        self.assertFalse(response.wsgi_request.recaptcha_is_valid)
+        with self.assertRaises(User.DoesNotExist):
+            User.objects.get(username=username)
```
hackdna: Why not just:
```python
self.assertRaises(User.DoesNotExist, User.objects.get, username=username)
```

refinery/templates/registration/registration_form.html
```diff
@@ -101,12 +101,19 @@ <h4>Terms of Service</h4>
 			<p>
         {{ form.spam_filter }}
       </p>
-
+      <script src='https://www.google.com/recaptcha/api.js'></script>
```
hackdna: Is this the right place to insert this script?

refinery/templates/registration/registration_form.html
```diff
@@ -101,12 +101,19 @@ <h4>Terms of Service</h4>
...
+      {% if recaptcha_error_message %}
+			  <p class="errors">
+          {{ recaptcha_error_message }}
+        </p>
+			{% endif %}
+      <div class="g-recaptcha" data-sitekey={{ REFINERY_GOOGLE_RECAPTCHA_SITE_KEY }}></div>
+      <br>
```
hackdna: Uneven indentation and formatting in this block.

refinery/core/test_views.py
scottx611x: When we add the attribute `recapthca_is_valid` [here](https://github.com/refinery-platform/refinery-platform/pull/3006/files?utf8=%E2%9C%93&diff=unified&w=1#diff-f3493f14bf81ad37fdc507fe58c10e29R1011 ), we are adding it to the `WSGIRequest` instance. The Django test client only returns this through [`response.wsgi_request`](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#django.test.Response.wsgi_request)

refinery/templates/registration/registration_form.html
scottx611x: The formatting looks fine in PyCharm for me. I think this is just how GitHub is rendering it.

hackdna: This is because some of the whitespace chars are actually tabs (rendered in PyCharm as two spaces and in Github as eight). They should all be spaces.

refinery/templates/registration/registration_form.html
```diff
@@ -114,6 +114,7 @@ <h4>Terms of Service</h4>
       </p>
 	</form>
 </div>
+</div>
```
hackdna: The code above should be indented accordingly. This will also help identifying missing tags later.

PR #3009, opened 2018-09-20
Data set manager utils cleanup

hackdna:
Remove unused functions


jkmarx:
I'm all set and wouldn't use that method. Thanks though.

PR #3011, opened 2018-09-21
Mccalluc/lru tool management

mccalluc:
The idea here is to pull out the old limit of 10 containers at a time, and rely instead on logic in the latest django_docker which will stop the oldest container if the newest requires more memory than we have available.


refinery/tool_manager/tests.py
```diff
@@ -3656,45 +3642,14 @@ def test_IGV(self):
         )
 
     def test_HiGlass(self):
+        # TODO: Is this necessary? It's a big image, it's not used on SCC.
```
jkmarx: Turn into an issue and remove.

deployment/manifests/aws.pp
```diff
@@ -22,6 +22,10 @@
 $solr_data_set_manager_data = "${data_dir}/solr/data_set_manager"
 $solr_core_data = "${data_dir}/solr/core"
 $django_docker_engine_data_dir = "${data_dir}/django-docker-engine-data"
+$django_docker_engine_mem_limit_mb = 3584
```
hackdna: I'd drop 'mb' from name since changing the units would require changing the code everywhere this variable is used.

hackdna: The standard practice is to add a comment specifying the units. Compare:
$django_docker_engine_mem_limit_mb = 3584
vs
$django_docker_engine_mem_limit = 3584  # size in MB

See also:
https://github.com/galaxyproject/galaxy/blob/dev/config/galaxy.yml.sample#L853
https://github.com/galaxyproject/galaxy/blob/dev/config/galaxy.yml.sample#L1586

deployment/manifests/default.pp
```diff
@@ -17,6 +17,7 @@
 $ui_app_root = "${django_root}/ui"
 $data_dir = '/data'
 $django_docker_engine_data_dir = "${data_dir}/django-docker-engine-data"
+$django_docker_engine_mem_limit_mb = 20
```
scottx611x: It seems like we wouldn't be able to test any of the beefier tools with a local Refinery after this change

refinery/tool_manager/schemas/ToolDefinition.json
```diff
@@ -37,7 +37,8 @@
         "file_relationship",
         "parameters",
         "extra_directories",
-        "image_name"
+        "image_name",
+        "mem_reservation_mb"
```
scottx611x: I'm wondering if it makes sense to not require this as part of the schema, but still require the type check if its supplied. Workflow ToolDefinitions will never use this, but it would be required to provide it which could get confusing.

deployment/manifests/aws.pp
mccalluc: I really don't believe there is a single universal standard here, but thank you for looking at it.

PR #3034, opened 2018-09-26
Mccalluc/lru tool management 0.1.2

mccalluc:
This rebranches from just before I added the `sleep` and bumps the version number. I'll close the other one, and reference this.

Tests still failing the same way as before. 😞 

----

Passing now! Yay!


refinery/tool_manager/tests.py
```diff
@@ -1627,7 +1613,7 @@ def test_visualization_is_running(self):
             start_vis_container=True
         )
         self.assertTrue(self.tool.is_running())
-        self._django_docker_engine_cleanup_wrapper()
+        DockerClientWrapper().purge_inactive(0)
```
scottx611x: Try without purging in these tests but don't block for now

refinery/config/config.json.erb
```diff
@@ -32,6 +32,7 @@
   },
   "DEFAULT_FROM_EMAIL": "<%= @default_from_email || "webmaster@localhost" %>",
   "DJANGO_DOCKER_ENGINE_DATA_DIR": "<%= @django_docker_engine_data_dir || "/data" %>",
+  "DJANGO_DOCKER_ENGINE_MEM_LIMIT_MB": "<%= @django_docker_engine_mem_limit_mb || 20 %>",
```
scottx611x: Could this be just an int to remove the int() cast in config.settings.base.py

Issue #3025, opened 2018-09-25
Update FileStoreItem.get_absolute_path() to support S3 storage backend


hackdna:
S3 objects do not have absolute file system path

Example error message:
```
2018-09-25 12:31:08 ERROR    django.request:256 handle_uncaught_exception() - Internal Server Error: /api/v2/file_store_items/c5508e0c-66c9-4f55-bf72-2ec4301e9d39/
Traceback (most recent call last):
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/rest_framework/views.py", line 466, in dispatch
    response = self.handle_exception(exc)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/rest_framework/views.py", line 463, in dispatch
    response = handler(request, *args, **kwargs)
  File "/vagrant/refinery/file_store/views.py", line 46, in get
    return Response(serializer.data)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/rest_framework/serializers.py", line 503, in data
    ret = super(Serializer, self).data
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/rest_framework/serializers.py", line 239, in data
    self._data = self.to_representation(self.instance)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/rest_framework/serializers.py", line 463, in to_representation
    attribute = field.get_attribute(instance)
  File "/home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/rest_framework/fields.py", line 405, in get_attribute
    return get_attribute(instance, self.source_attrs)
[...]

Issue #2718, opened 2018-04-11
Rewrite generate_bam_index()


hackdna:
* FIXME refers to standalone IGV that is no longer used (v1.6.3)
* `pysam.index()` generates `.bai` file inside `FILE_STORE_BASE_DIR` which will not work with S3 storage backend since it circumvents file_store API (https://github.com/refinery-platform/refinery-platform/blob/v1.6.3/refinery/data_set_manager/tasks.py#L471)
* getting `bai` extension from FileExtension object is redundant

See also #2210

PR #3037, opened 2018-09-27
Remove generate_bam_index()

### LINKED ISSUE #2713: Avoid generating BAM index file under FILE_STORE_BASE_DIR tree
* Specific code commit: v1.6.3
* Environment where the error occurred: any

### Steps to reproduce
Run a workflow that generates a BAM file

### Observed behavior
A `.bai` file is [generated directly in the same directory as the BAM file](https://github.com/refinery-platform/refinery-platform/blob/v1.6.3/refinery/data_set_manager/tasks.py#L471)

### Expected behavior
A `.bai` file should be added using the file_store API or in a temporary directory

### Notes
This functionality will break when file_store is updated to use S3 as a storage backend
See also #2718

hackdna:
Close #2713 and #2718. Required for #3025. This function was only necessary for standalone IGV which is no longer used. Also, remove related code that is no longer necessary.


hackdna:
So the [comment in generate_bam_index()](https://github.com/refinery-platform/refinery-platform/blob/44d7e1799de21e5822ad9af741be86568f35f02b/refinery/data_set_manager/tasks.py#L451) is incorrect?

scottx611x:
@hackdna No, it should still be refactored/rewritten Ref https://github.com/refinery-platform/refinery-platform/issues/2718. When that code was introduced we only had Standalone IGV implemented.

hackdna:
Strange, "standalone" implies that it would not be needed with a different type of IGV.

Issue #2977, opened 2018-08-31
Meta-data update - modify existing node's content (back-end)


jkmarx:
- Backend
    - [ ] Non-derived cells and non-names

Some cells require multiple places to be updated. Questionable urls, name, etc updates.
- [ ] Validation
- [ ] Disallow editing names

PR #3039, opened 2018-09-28
Jkmarx/update node api patch edit

jkmarx:
- Ref #2977 

- add dataset hg-19 mocker for testing
- add dataset isa-9909 mocker for testing
- add logic to node api to edit non-derived cells

Note: Still need to add text validation and also disallow name (or other fields).

### DIFF
--- a/refinery/data_set_manager/test_views.py
+++ b/refinery/data_set_manager/test_views.py
@@ -22,7 +22,7 @@ from core.models import DataSet, InvestigationLink
 from core.test_views import APIV2TestCase
 from file_store.models import FileStoreItem
 
-from .models import (Attribute, AnnotatedNode, Assay, AttributeOrder,
+from .models import (AnnotatedNode, Assay, Attribute, AttributeOrder,
                      Investigation, Node, Study)
 from .tests import MetadataImportTestBase
 from .views import (AddFileToNodeView, Assays, AssaysAttributes, NodeViewSet,
--- a/refinery/data_set_manager/test_views.py
+++ b/refinery/data_set_manager/test_views.py
@@ -10,6 +10,7 @@ from django.http import QueryDict
 from django.test import LiveServerTestCase, override_settings
 
 from factory_boy.utils import (create_dataset_with_necessary_models,
+                               create_hg_19_data_set,
                                make_analyses_with_single_dataset)
 from guardian.shortcuts import assign_perm
 from mock import ANY
@@ -20,10 +21,12 @@ from core.models import DataSet, InvestigationLink
 from core.test_views import APIV2TestCase
 from file_store.models import FileStoreItem
 
-from .models import AnnotatedNode, Assay, AttributeOrder, Investigation, Study
+from .models import (AnnotatedNode, Assay, AttributeOrder, Investigation,
+                     Node, Study)
 from .tests import MetadataImportTestBase
 from .views import (AddFileToNodeView, Assays, AssaysAttributes, NodeViewSet,
                     import_file)
+
 
 TEST_DATA_BASE_PATH = "data_set_manager/test-data/"
 
@@ -789,6 +792,7 @@ class NodeViewAPIV2Tests(APIV2TestCase):
             view=NodeViewSet.as_view()
         )
         self.data_set = create_dataset_with_necessary_models(user=self.user)
+        self.hg_19_data_set = create_hg_19_data_set(user=self.user)
         self.node = self.data_set.get_nodes()[0]
 
     @mock.patch('data_set_manager.models.Node.update_solr_index')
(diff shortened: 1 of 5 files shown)


refinery/data_set_manager/views.py
```diff
@@ -1170,11 +1182,93 @@ def patch(self, request, uuid):
...
+        :param checked_nodes: str, value for attribute's value
+        :return: start_nodes
+        """
+        node_attributes = Attribute.objects.filter(node=node)
+        # splits solr name into type and subtype
+        attribute_obj = customize_attribute_response([solr_name])[0]
+        attribute_type = attribute_obj.get('attribute_type')
+        attribute_subtype = attribute_obj.get('display_name').lower()
+
+        if node_attributes.exists():
```
scottx611x: For both of the filtering tasks here (`node_attributes ` & `annotated_nodes` below) I think you'd be fine attempting to iterate over the resulting querysets if they ended up being empty.  I think it may end up being clearer as to what we're trying to filter for if we include the desired fields in a single query.

```
node_attributes = Attribute.objects.filter(node=node, type=attribute_type, subtype=attribute_subtype)
for attribute in node_attributes:
    attribute.value = new_value
    attribute.save()
```

refinery/factory_boy/utils.py
```diff
@@ -212,3 +212,362 @@ def create_tool_with_necessary_models(tool_type, user=None):
     if user is not None:
         tool.set_owner(user)
     return tool
+
+
+def create_hg_19_data_set(user=None):
```
scottx611x: For these two: `create_hg_19_data_set` & `create_isatab_9909_data_set` maybe call them:`create_mock_*_data_set`

refinery/factory_boy/utils.py
```diff
@@ -212,3 +212,362 @@ def create_tool_with_necessary_models(tool_type, user=None):
...
+    # Generally mocks the hg_19 local data set's study, assay, nodes,
```
scottx611x: Maybe adding in tests to assert against the results of the parsed versions of these respective datasets would be good since these structures being correct are important to this new functionality

refinery/data_set_manager/test_views.py
```diff
@@ -846,6 +851,230 @@ def test_patch_edit_field_405_status(self):
...
+                                             annotated_node.attribute_type)
+        patch_request = self.factory.patch(urljoin(self.url_root, node.uuid),
+                                           {"attribute_solr_name": solr_name,
+                                            "attribute_value": new_value})
+        force_authenticate(patch_request, user=self.user)
+        self.view(patch_request, node.uuid)
+        self.assertEqual(annotated_node.attribute.value, new_value)
+
+    @mock.patch('data_set_manager.models.Node.update_solr_index')
+    def test_patch_edit_updates_attribute_value(self, update_solr_index_mock):
```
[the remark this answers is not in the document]
jkmarx: yes, but the above checks the source attribute gets updated while this checks the hardcoded (field on the annotated_node) attribute_value gets updated.

refinery/data_set_manager/views.py
jkmarx: can update for node attributes, but want to keep it for annotated_nodes as we don't want to update solr_index if there's no annotated_nodes

refinery/data_set_manager/test_views.py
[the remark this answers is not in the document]
scottx611x: Ah just saw the slight difference in the assertions. I missed the: `annotated_node.attribute.value` vs. `annotated_node.attribute_value`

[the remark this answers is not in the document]
jkmarx: will add a comment for clarity

refinery/data_set_manager/views.py
```diff
@@ -1170,11 +1182,97 @@ def patch(self, request, uuid):
...
+                    NodeSerializer(node).data, status=status.HTTP_200_OK
+                )
+
+            return Response('Currently, you can only remove node files or '
+                            'edit non-derived files',
                             status=status.HTTP_405_METHOD_NOT_ALLOWED)
 
         return Response(uuid, status=status.HTTP_401_UNAUTHORIZED)
 
+    def find_root_attributes(self, parent_nodes, attribute_types,
```
scottx611x: I think that this would still be more appropriate as a method on the `Node` model.

Consider its usage above: 
```
root_node, root_attributes = self.find_root_attributes(
     [node], attribute_obj
)
```

We already are operating on a single `Node` here, the wrapping in a list was a code smell to me.

I think the signature could be changed to: 
`def find_root_attributes(self, attribute_types, parent_nodes=None, checked_nodes=None):`

And the inner logic to:
```
        if checked_nodes is None:
            checked_nodes = []
        if parent_nodes is None:
            parent_nodes = self.parents.all()

        checked_nodes += [
            node for node in parent_nodes
            if node not in checked_nodes
        ]

        for parent_node in parent_nodes:
            node_attributes = Attribute.objects.filter(
                node=parent_node,
                type=attribute_types.get('attribute_type'),
                subtype=attribute_types.get('display_name').lower()
            )
            if node_attributes.exists():
                return parent_node, node_attributes
            else:
                return self.find_root_attributes(
                    attribute_types,
                    parent_nodes=parent_node.parents.all(),
                    checked_nodes=checked_nodes
                )
```

refinery/data_set_manager/views.py
```diff
@@ -1170,7 +1181,41 @@ def patch(self, request, uuid):
...
+                    node=node,
+                    attribute_type=attribute_obj.get('attribute_type'),
+                    attribute_subtype__icontains=attribute_obj.get(
+                        'display_name'
+                    )
+                )
+
+                # update the source attribute
+                try:
+                    source_attribute = annotated_nodes_query[0].attribute
```
scottx611x: Do we know that the first item in this queryset is the AnnotatedNode that we want? Also, remember that the order of the `filter()` isn't guaranteed.

jkmarx: Yes, it should only be one item unless there duplicates. If there duplicates, it doesn't matter as they will be updated below as long as the source attribute gets updated. Of course, if you can think of user edge cases you'd like for me to add as a unit test, please let me know.

PR #3047, opened 2018-10-02
Remove references to DJANGO_DOCKER_ENGINE_DATA_DIR

mccalluc:
This had been one way of providing input to containers, but is no longer actually a parameter of DockerClientSpec.

(The value we were passing in was being cast to a Boolean, and causing the INPUT_JSON envvar to be set in the container environment.)

When "extra_directories" are passed in to DockerContainerSpec, Docker is creating directories on the host for storage, but the exact location is an internal detail we do not need to worry about.


mccalluc:
I merged [this fix to Intervene](https://github.com/refinery-platform/intervene-refinery-docker/commit/5b2ccb768268bbe36260ccd802ba56c579acb130) to master in that repo by mistake... but that should also fix the immediate problem, so this PR is not urgent.

PR #3048, opened 2018-10-02
Enable tagging of all resources managed by Terraform

hackdna:


deployment/terraform/modules/docker_host/main.tf
```diff
@@ -33,6 +31,13 @@ resource "aws_instance" "docker_host" {
   root_block_device {
     volume_size = "20"
   }
+  tags                   = "${merge(
+    var.tags, map("Name", "${terraform.workspace} docker host")
+  )}"
+  volume_tags            = "${merge(
+    var.tags, map("Name", "${terraform.workspace} docker host")
+  )}"
```
mccalluc: Sorry for just not getting the syntax, but `tags` had been a map type, and is it now a string. Or after interpolation is it transformed into a map?

mccalluc: Would it be worth defining a variable rather than repeating these lines?

deployment/terraform/modules/vpc/main.tf
```diff
@@ -2,31 +2,26 @@ resource "aws_vpc" "vpc" {
...
 resource "aws_subnet" "public_subnet" {
   vpc_id                  = "${aws_vpc.vpc.id}"
   cidr_block              = "${var.public_cidr_block}"
   availability_zone       = "${var.availability_zone_a}"
   map_public_ip_on_launch = true
-
-  tags {
-    Name = "${terraform.workspace} (terraform public subnet)"
-  }
+  tags              = "${merge(
```
mccalluc: It's been a while, but I think you prefer the "=" to be aligned in each section, and I think there's a tool that does that for you?

deployment/terraform/live/main.tf
```diff
@@ -29,25 +37,27 @@ module "identity_pool" {
...
-
 module "vpc" {
   source               = "../modules/vpc"
   vpc_cidr_block       = "${var.vpc_cidr_block}"
   public_cidr_block    = "${var.public_cidr_block}"
   private_cidr_block_a = "${var.private_cidr_block_a}"
   private_cidr_block_b = "${var.private_cidr_block_b}"
-  availability_zone    = "${var.availability_zone_a}"
+  availability_zone_a  = "${var.availability_zone_a}"
+  availability_zone_b  = "${var.availability_zone_b}"
```
mccalluc: Is there a correspondence between the cidr_blocks and the availability_zones? Would this be a good place for a comment to explain what's going on here, or that might go somewhere else?

deployment/terraform/modules/vpc/vars.tf
```diff
@@ -2,10 +2,9 @@ variable "vpc_cidr_block" {}
...
 variable "private_cidr_block_b" {}
 variable "public_cidr_block" {}
-variable "availability_zone" {}
-variable "availability_zone_a" {
-  default = "us-east-1a"
-}
-variable "availability_zone_b" {
-  default = "us-east-1b"
+variable "availability_zone_a" {}
+variable "availability_zone_b" {}
```
mccalluc: Is the availability zone now something that gets specified on the command line when invoking terraform, or does it come from somewhere else?

deployment/terraform/modules/docker_host/main.tf
hackdna: Interpolation works fine:
```hcl
$ terraform console
> map("a", "b")
{
  "a" = "b"
}
> "${map("a", "b")}"
{
  "a" = "b"
}
```
This expression is used only here. Declaring it as a variable would require putting it in a  separate `local` block outside of this resource which would reduce readability.

deployment/terraform/modules/vpc/main.tf
hackdna: Pushed a fix. HCL plugin for PyCharm does have formatting functionality but it doesn't work very well.

deployment/terraform/live/main.tf
hackdna: You might want to ask @scottx611x since this is his code :) I've simply removed an unused variable `availability_zone` and added variables that were [declared](https://github.com/refinery-platform/refinery-platform/blob/95025f0a8d1bbf2e90fbc4094928989c3e3279a6/deployment/terraform/live/vars.tf) but not assigned: `availability_zone_a` and `availability_zone_b`.

scottx611x: The relationship between the private cidr blocks a & b and availability zones a & b is that to be able to put an RDS instance inside of a VPC there must be two subnets in two different Availability Zones in the region where we deploy. Since we don't have the RDS hotswapping enabled, `private_cidr_block_b` and `availability_zone_b` don't matter other than to satisfy that requirement. I thought I had some comment about this in the TF code let me check

hackdna:
@mccalluc is this good to merge?

Issue #2602, opened 2018-02-16
Migrate stack.py to Terraform


hackdna:
This will provide a simple, unified process for deployment on AWS.

See also: #1837, #2307 and #2597 

New features: deployment to non-default VPC

Existing features:
- [ ] ELB
  - [ ] listeners
  - [ ] SSL termination
  - [ ] S3 log bucket and policy
- [ ] `config.yaml` to `vars.tf`
- [ ] WebInstance
  - [ ] user data script
  - [ ] data volume and attachment
  - [ ] security group
- [ ] RDSInstance
  - [ ] security group
- [ ] SMTP user
- [ ] Backups (RDS and EC2 instance tag)

PR #3049, opened 2018-10-05
Migrate S3 log bucket code to Terraform

hackdna:
Toward #2602


hackdna:
`S3_LOG_BUCKET` is set automatically in `load_config()`, so no changes.

PR #3054, opened 2018-10-11
Add name variable to all Terraform modules

hackdna:
Make Terraform modules self-contained


scottx611x:
@hackdna What are your thought on having a `terraform validate` As part of our CI process?

hackdna:
I'm not sure it is worth additional complexity of adding Terraform support to Travis (writing code to install the binary and AWS provider, initializing the environment, making sure versions are up-to-date, installation taking potentially a long time, etc). I'd also expect developers at least to run `terraform plan` before committing Terraform code which already includes the validation step.

PR #3055, opened 2018-10-11
Jkmarx/fix transfer ownership bug

### LINKED ISSUE #3045: Transfer Data Set Bug
@jkmarx 

On the HSCI site:
old owners have a lingering 'share-data_set' perms

Should not have any lingering perms after transfer.

jkmarx:
- Resolves #3045 
- Removes a custom data_set get_owner method which used add_ perms

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -313,13 +313,16 @@ class OwnableResource(BaseResource):
         if self._meta.verbose_name == 'dataset':
             assign_perm("read_meta_%s" % self._meta.verbose_name, user, self)
 
-    def transfer_ownership(self, user, new_owner):
+    def remove_owner(self, user):
         remove_perm("add_%s" % self._meta.verbose_name, user, self)
         remove_perm("read_%s" % self._meta.verbose_name, user, self)
         remove_perm("delete_%s" % self._meta.verbose_name, user, self)
         remove_perm("change_%s" % self._meta.verbose_name, user, self)
         if self._meta.verbose_name == 'dataset':
             remove_perm("read_meta_%s" % self._meta.verbose_name, user, self)
+
+    def transfer_ownership(self, user, new_owner):
+        self.remove_owner(user)
         self.set_owner(new_owner)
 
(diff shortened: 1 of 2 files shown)


refinery/core/models.py
```diff
@@ -364,10 +367,36 @@ class SharableResource(OwnableResource):
     def __unicode__(self):
         return self.name
 
+    def get_owner(self):
```
jkmarx: Moved from the data_set class. Fritz noted this method improved performance 10%-20% percent.

refinery/core/models.py
```diff
@@ -364,10 +367,36 @@ class SharableResource(OwnableResource):
...
+        )
+
+        if perms.count() > 0:
+            try:
+                owner = User.objects.get(id=perms[0].user_id)
+            except User.DoesNotExist:
+                pass
+
+        return owner
+
```
mccalluc: Feel free to ignore, since this is just a move, but:
- There is an [`exists()`](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#exists), but I'd be surprised if it's much of an improvement over `count()`.
- Maybe just return in the try? (and get rid of `owner = None` and `return owner`)
- I can see that it's tested below, but I can't remember what it would mean to be ownerless, and how that would be useful.

refinery/core/test_models.py
```diff
@@ -890,6 +897,106 @@ def test_get_details_as_dict(self):
...
+
+    def test_set_owner_data_sets_read_perms(self):
+        self.data_set.set_owner(self.user)
+        user_perms = get_perms(self.user, self.data_set)
+        self.assertTrue('read_dataset' in user_perms)
+
+    def test_set_owner_data_sets_read_meta_perms(self):
+        self.data_set.set_owner(self.user)
+        user_perms = get_perms(self.user, self.data_set)
+        self.assertTrue('read_meta_dataset' in user_perms)
```
mccalluc: I might have done all of these at once with `issubset`, but this works, too.

PR #3056, opened 2018-10-15
Migrate RDS instance to Terraform

hackdna:
Toward #2602. Add `RDS_ENDPOINT_ADDRESS` and `RDS_SUPERUSER_PASSWORD` variables to `config.yaml` until migration to TF is complete. Remove `RDS_DB_SUBNET_GROUP_NAME` and `RDS_NAME` from `config.yaml`.


hackdna:
We have not had an outage yet due to running RDS in a single AZ. Also, if there is a problem in that AZ then the app server instance would be likely affected as well and it does not have a failover configuration, so the site would be down anyway (and this would double the spend on RDS).

PR #3057, opened 2018-10-15
Scottx611x/fix duplicate facet fields

scottx611x:
Fixes issue with initialization/display of `FileBrowser` when duplicate `facet_fields` are present in the `Solr` response.

The issue currently affects `DataSets` that have had a metadata revision performed

**Before:**
screen shot 2018-10-15 at 3 26 19 pm

**After:**
screen shot 2018-10-15 at 3 26 29 pm


jkmarx:
@scottx611x It doesn't actually resolve the issue correct? It band-aids the solr response but in the db/solr index there are duplicates correct?

scottx611x:
@jkmarx I think the duplicates existing within Solr are fine in this case. We're re-parsing a new metadata file and duplicates of all of these objects are going to be created as long as the metadata is the same/similar enough. 

The dupes are still properly attached to nodes that are associated with the latest `Investigation` version.

I don't think we would want to remove the older version of the `DataSet's` information from Solr though. I could imagine wanting to resort to or view prior versions of a DataSet in the future and this would make that problematic.

I think a better fix would be rooted in understanding why the Solr query is returning information for `facet_fields` across multiple versions of a DataSet, but I was unable to track down why this is happening and this route was the least invasive one I could find.

scottx611x:
Closing as more investigation into the root cause is needed

scottx611x:
After talking to Geoff, this issue is still affecting users so an approach like this may be a good intermediary until the root cause is found and addressed.

PR #3060, opened 2018-10-17
Scottx611x/integrate trevors observable

scottx611x:
oct-17-2018 16-13-29

### DIFF
--- a/refinery/core/admin.py
+++ b/refinery/core/admin.py
@@ -151,7 +151,7 @@ class SiteProfileAdmin(GuardedModelAdmin):
 
 
 class SiteStatisticsAdmin(AdminFieldPopulator):
-    pass
+    change_list_template = "admin/core/sitestatistics/change_list.html"
 
 
 class EventAdmin(AdminFieldPopulator):
--- a/refinery/core/urls.py
+++ b/refinery/core/urls.py
@@ -10,7 +10,8 @@ from constants import UUID_RE
 from rest_framework.routers import DefaultRouter
 
 from .views import (AnalysesViewSet, DataSetsViewSet, EventViewSet,
-                    OpenIDToken, UserProfileViewSet, WorkflowViewSet)
+                    OpenIDToken, UserProfileViewSet, WorkflowViewSet,
+                    site_statistics)
 
 urlpatterns = patterns(
     'core.views',
@@ -76,5 +77,6 @@ core_router.urls.extend([
     url(r'^analyses/(?P<uuid>' + UUID_RE + r')/$',
         AnalysesViewSet.as_view()),
     url(r'^openid_token/$',
-        OpenIDToken.as_view(), name="openid-token")
+        OpenIDToken.as_view(), name="openid-token"),
+    url(r'^sitestatistics/(?P<type>(deltas)|(totals)).csv', site_statistics)
 ])
--- a/refinery/core/urls.py
+++ b/refinery/core/urls.py
@@ -61,7 +61,8 @@ urlpatterns = patterns(
 
     url(r'^neo4j/annotations/$',
         'neo4j_dataset_annotations', name="neo4j_dataset_annotations"),
-    url(r'^auto_login/$', 'auto_login', name='auto_login')
+    url(r'^auto_login/$', 'auto_login', name='auto_login'),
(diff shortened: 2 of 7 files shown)


refinery/core/views.py
```diff
@@ -1091,3 +1093,23 @@ def patch(self, request, uuid):
         return Response(
             serializer.errors, status=status.HTTP_400_BAD_REQUEST
         )
+
+
+@api_view(['GET'])
+@staff_member_required
+def site_statistics(request, **kwargs):
+    site_statistics_type = kwargs.get("type")
+    response = HttpResponse()
```
hackdna: Should this be `rest_framework.response.Response` if it's a DRF API view?

refinery/core/admin.py
```diff
@@ -151,7 +151,7 @@ class SiteProfileAdmin(GuardedModelAdmin):
 
 
 class SiteStatisticsAdmin(AdminFieldPopulator):
-    pass
+    change_list_template = "admin/core/sitestatistics/change_list.html"
```
hackdna: Is folder necessary for a single file?

refinery/core/views.py
scottx611x: No, it doesn't have to be. In this case, it's simpler without. To return csv data with a DRF `Response` we would actually have to add a [third party library](https://www.django-rest-framework.org/api-guide/renderers/#csv) to get & use a `CSVRenderer`.

refinery/core/admin.py
scottx611x: I thought it made sense for organizational purposes

refinery/core/views.py
hackdna: OK then does it need to be an API view?

scottx611x: The decorator ended up being superfluous

refinery/core/urls.py
```diff
@@ -76,5 +77,6 @@
     url(r'^analyses/(?P<uuid>' + UUID_RE + r')/$',
         AnalysesViewSet.as_view()),
     url(r'^openid_token/$',
-        OpenIDToken.as_view(), name="openid-token")
+        OpenIDToken.as_view(), name="openid-token"),
+    url(r'^sitestatistics/(?P<type>(deltas)|(totals)).csv', site_statistics)
```
hackdna: If site_statistics is no longer a DRF API view then should it be managed by DRF URL config?

refinery/templates/admin/core/sitestatistics/change_list.html
```diff
@@ -0,0 +1,21 @@
+{% extends "admin/change_list.html" %}
+{% load i18n admin_urls %}
+
+{% block extrahead %}
+  {{ block.super }}
+  <script type="module">
+    // Load the Observable runtime and inspector.
+    import {Runtime, Inspector} from "https://unpkg.com/@observablehq/notebook-runtime?module";
+
+    import notebook from "https://api.observablehq.com/@scottx611x/refinery-stats.js";
```
hackdna: Shouldn't this live in something like https://api.observablehq.com/@refinery-platofrm instead?

scottx611x: @hackdna I don't believe it should. That would require us to create another github user: `@refinery-platform`. I'd rather have the link tied to a real person than an umbrella user account.

hackdna: So, what would happen if we lost access to https://api.observablehq.com/@scottx611x and needed to make changes to refinery-stats.js?

scottx611x: @hackdna I mean, you could say the same about the refinery-platform github user you are suggesting... Who would have access/administrate that account? Would we be sharing credentials amongst ourselves? If changes really need to be made, the Observable is public and can be forked.

hackdna: Is there no way to provide access to refinery-platform organization?

scottx611x: No, at least not at the moment.

https://beta.observablehq.com/terms-of-service

```
1. Required Information
You must log in with a valid GitHub account, linked to a valid email address in order to complete the signup process. Any other information requested, such as your real name, is optional.
```

PR #3065, opened 2018-10-23
Jkmarx/non edit background internal attributes

jkmarx:
- For internal attributes add custom css class via ui-grid colProperties 

non-edit-attributes


refinery/ui/source/js/file-browser/services/factory.spec.js
```diff
@@ -176,34 +176,72 @@
...
       });
 
+      it('sets correct name', function () {
+        expect(response[0].name).toEqual(tempAssayAttributes[0].display_name);
+        expect(response[1].name).toEqual(tempAssayAttributes[1].display_name);
+      });
+
+      it('sets custom url name', function () {
+        expect(response[2].name).toEqual(tempAssayAttributes[2].internal_name);
+      });
```
mccalluc: Should this be display_name too? It seems redundant, as-is.
```
0 + 1: display_name
2: internal_name
0 + 1: internal_name
2: internal_name
```

jkmarx: No, the url is a custom method which grabs from the internal name and not the display name.

PR #3068, opened 2018-10-24
Scottx611x/postgresql upgrade

### LINKED ISSUE #2707: Upgrade PostgreSQL to version 10
* Currently used [version 9.3 will reach EOL in September 2018](https://www.postgresql.org/support/versioning/)
* [Amazon RDS for PostgreSQL now supports PostgreSQL major version 10](https://aws.amazon.com/about-aws/whats-new/2018/02/postgresql-10-now-supported-in-amazon-rds/)
* Version 10 is the new default when installing PostgreSQL using Homebrew in OS X
* A dump/reload of the database or use of the pg_upgrade module is required for major upgrades
* May require [upgrade to Ubuntu 18.04 LTS](https://github.com/refinery-platform/refinery-platform/issues/2616)
* Currently used version of PostgreSQL Puppet module (4.9.0) [works with PostgreSQL versions through 9.5 only](https://forge.puppet.com/puppetlabs/postgresql/4.9.0/readme#limitations), so will probably require an upgrade.
* [Upgrading the PostgreSQL DB Engine](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html)

Notice from AWS in July 2018:

This notification is being sent to you to as a reminder that Amazon RDS will be discontinuing support for version 9.3 of the PostgreSQL database in RDS, starting on September 5th, 2018. 

Amazon RDS has been supporting the PostgreSQL major version 9.3 since 2013. That release has since been replaced by subsequent major versions that contain significant improvements in functionality, security, reliability and performance. The PostgreSQL community has announced the end of life date for PostgreSQL 9.3 as September 2018 [1].  Aligning with the community support model, AWS will be ending support for the PostgreSQL major version 9.3 in Amazon RDS, including its minor engine versions including 9.3.10, 9.3.12, 9.3.14, 9.3.16, 9.3.17, 9.3.19, 9.3.20, and 9.3.22. We will continue to support 9.4 and more recent versions of the PostgreSQL database on Amazon RDS.

We recommend that you upgrade your Amazon RDS PostgreSQL database instances to either version 9.6 or version 10 at your earliest convenience. You can use the Major Version Upgrade feature of RDS to upgrade databases to the next higher version. You can initiate an upgrade by going to the Modify DB Instance page in the AWS Management Console and changing the database version setting to a newer major version of PostgreSQL. If you choose the "Apply Immediately" option, the upgrade will be initiated immediately after exiting the Modify page. If you choose not to apply the change immediately, the upgrade will be performed during your subsequent maintenance window.  To learn more about upgrading PostgreSQL major versions in RDS, review the Upgrading Database Versions page [2].

As part of the retirement plan for Amazon RDS PostgreSQL 9.3, starting the week of August 6th 2018, you will no longer be able to create any new PostgreSQL 9.3 instances using the AWS console. **In November 2018, your remaining 9.3 instances will be force upgraded to the most recent minor version of PostgreSQL 9.4. Once this occurs, any instance restored from a snapshot of a deprecated version will automatically and immediately be upgraded to a currently supported version.** The exact dates and details on force upgrades will be communicated in September 2018.

scottx611x:
Fixes #2707


hackdna:
What would be the step by step process to upgrade a production RDS instance using this code?

scottx611x:
@hackdna:

1. Have some planned downtime and alert users.
2. Take the production site down.
3. Restore the latest snapshot of the production site's RDS instance and create an RDS instance to perform the major engine version upgrades on.
4. Upgrade said instance's engine version to `PostgreSQL 10.5` 
    - This requires following an upgrade path like so: `9.3.x -> 9.4.x -> 9.5.x - 9.6.x -> 10.5`
5. After the upgrading is complete a new db snapshot should be created from this instance
    - This snapshot will have the latest production site data, but just with Postgres engine version 10.5
6. We'll be able to then use said db snapshot along with the latest data snapshot to bring up the production instance again.

.travis.yml
```diff
@@ -36,13 +38,18 @@ env:
     - secure: oGdTYh3rnPdG7qVGYq3TgVESypir1btO2kmafumtKR6o+FYmMtb09pe+HtiT6Qfv4pDwnYZ2RTYqsTY4h6POigU5j7qQPSsZ5WkhHt9ybFyt/vfo2B95GaAIFt+8LVjTX5TItyw0/PPUCCUluQ1n8k/NCtSfMb0kIT/CuLrg7N8=
 
 before_script:
+  # PostgreSQL-10 is still not fully supported by Travis.
```
hackdna: This post (https://github.com/travis-ci/travis-ci/issues/8537#issuecomment-399738282) says that the approach described in Travis docs does indeed work (https://docs.travis-ci.com/user/database-setup/#using-a-different-postgresql-version). All you have to do is use a port other than 5432.

deployment/refinery-modules/refinery/manifests/postgresql.pp
```diff
@@ -1,5 +1,7 @@
 class refinery::postgresql {
-  $server_version = '9.3'
+  # Postgres versioning scheme provides latest point releases when
+  # specifying major version: https://www.postgresql.org/support/versioning/
+  $server_version = '10'
```
hackdna: The minor version should be specified explicitly. If a new instance gets a new minor version which does not match the version of the snapshot then restore will fail.

deployment/terraform/live/vars.tf
```diff
@@ -57,3 +57,8 @@ variable "rds_master_user_password" {
   description = "Password for the master DB user"
   default     = ""  # will be autogenerated if left blank
 }
+
+variable "rds_engine_version" {
+  description = "PostgreSQL engine verision to target for RDS instances"
+  default     = "10.5"
+}
```
hackdna: Not needed.

.travis.yml
scottx611x: Yes, I saw this. It came down to these few lines in .travis.yml, or providing a travis-specific Django settings file pointing to the non-standard port, and I chose this workaround.

Presumably when https://github.com/travis-ci/travis-ci/issues/8537 is closed, the travis-provided postgresql will also operate on the default 5432 port.

deployment/refinery-modules/refinery/manifests/postgresql.pp
scottx611x: Specifying the `10` here isn't any different than how we were specifying the `9.3` before. Those are both major versions of PostgreSQL. It just happens to be that the Postgres folks have changed their versioning scheme starting with the [`10` series of releases](https://www.postgresql.org/support/versioning/).

deployment/terraform/live/vars.tf
scottx611x: If I hardcode this value in rds/main.tf someone will have to edit tracked files to bring up an instance targeting a non-`10.5` RDS Snapshot. I thought that this would still provide some flexibility in the iterim until we know that all deployed instances of refinery are up to date.

.travis.yml
hackdna: OK but this adds to technical debt of having separate/duplicated provisioning code on Travis (see #822).

deployment/refinery-modules/refinery/manifests/postgresql.pp
hackdna: OK, fair enough. This is not the RDS version actually.

deployment/terraform/live/vars.tf
hackdna: The upgrade operation is atomic 9.3 -> 10.5. Both versions were tested, as well as the upgrade procedure, so there is no need to bother with intermediate versions.
Actually, setting a value here that is different from the one specified in `postgres.pp` will probably result in error anyway.
So, what would be useful is to leave the precise RDS version hardcoded in `rds/main.tf` then pass it to Puppet like `RDS_ENDPOINT_ADDRESS` but for now it can just stay hardcoded in those two places.

scottx611x: It is my understanding that the packages provided by the [PostgreSQL Apt repository](https://wiki.postgresql.org/wiki/Apt) don't allow for the installation of specification of minor versions. Currently when one does a: `apt-get install postgresql-<major_version>`, the latest minor version is installed. I think it just happens to be that we have the ability to be more specific when creating an RDS instance.

I will move the variable back to a hardcoded value in `postgres.pp`

deployment/aws.sh
```diff
@@ -14,7 +14,7 @@ env
 GIT_BRANCH=${GIT_BRANCH:-develop}
 
 /usr/bin/apt-get -q -y install htop
-/usr/bin/apt-get -q -y install awscli jq postgresql-client-9.3
+/usr/bin/apt-get -q -y install awscli jq postgresql-client-10
```
hackdna: I don't think this will work because the apt repo that provides postgresql-client-10 is not configured at this point.

scottx611x: @hackdna You're right, but after looking into it I think the only reason we had to do the `postgresql-client` install in `aws.sh` was due to the [`ensure-postgresql-database` script](https://github.com/refinery-platform/refinery-platform/pull/3061/files#diff-69a92f04c759417a5a4dc7c0821e122cL89).

Since that was removed in #3061, I don't think we need the `postgresql-client` in this script anymore.

hackdna: OK then remove it and check if the stack still comes up?

hackdna: Great. This should probably be the last PR merged before the release to make bringing up stacks using RDS snapshots easier in the meantime.

PR #3069, opened 2018-10-24
Update versions of installation prerequisites

hackdna:


hackdna:
2.17.1 is the version of git that comes with OS X 10.13.6 and the one I have tested with. If you can confirm that v2.19.0 works then I can update the PR.

scottx611x:
@hackdna I've been using `git v2.19.0` for the past 5 days and haven't had any issues

Issue #2982, opened 2018-08-31
Meta-data update - modify existing node's content - UI


jkmarx:
- Modify existing cell in file browser (UI)
    - [x] UI Grid implemented
    - [x] Basic Text Edit a cell with Double Click (utilizing the library)
    - [ ] Add ui validation

scottx611x:
https://github.com/refinery-platform/refinery-platform/issues/2896 Will come into play. We may not want to allow for such an edit to happen if a given `Node` and any of it's parent/descendants were used as Tool inputs

PR #3070, opened 2018-10-24
Jkmarx/file browser disable derived rows

jkmarx:
Ref #2982 
- Disable and add css to non-derived cell. The plan is to enhance at some point and allow for editing of these.

screen shot 2018-10-24 at 3 12 57 pm

### DIFF
--- a/refinery/ui/source/js/file-browser/services/factory.js
+++ b/refinery/ui/source/js/file-browser/services/factory.js
@@ -84,6 +84,16 @@
       return null;
     }
 
+    // helper method to check if cell is derived
+    function isCellDerived (row) {
+      for (var cellName in row) {
+        if (cellName.includes('REFINERY_TYPE') && row[cellName].includes('Derived')) {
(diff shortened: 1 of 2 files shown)


refinery/ui/source/js/file-browser/services/factory.js
```diff
@@ -79,6 +79,16 @@
...
 
+    // helper method to check if cell is derived
+    function isCellDerived (row) {
+      for (var cellName in row) {
+        if (cellName.includes('REFINERY_TYPE') && row[cellName].includes('Derived')) {
+          return true;
+        }
+      }
+      return false;
+    }
```
mccalluc: consider using `some()`:
```javascript
function isNameDerived(cellName) {
  return cellName.includes('REFINERY_TYPE') && row[cellName].includes('Derived')
};

function isCellDerived(row) {
  return row.some(isNameDerived);
}
```

PR #3074, opened 2018-10-29
Scottx611x/properly reflect admin dataset ownership

### LINKED ISSUE #825: Admin owns everything?
I notice that the API returns all datasets as owned by the admin user when being logged in as the admin.

http://192.168.50.50:8000/api/v1/data_sets/?format=json&limit=50&offset=0&order_by=-modification_date&is_owner=True

This is a very minor issue but it's a bit confusing as the admin might have some test datasets or sometimes are an actual user at the same time and then it would be helpful to be able to distinguish between truly owned and read access datasets.

scottx611x:
Fixes #825 

Take note of the `Owner` entry in the preview on the right, and the presence of a `person icon` next to the DataSet modification date on the left.

> **Note:** This change also has the added benefit of allowing the admin user to import other user's ISATab-based datasets into their own space allowing for easier debugging of analysis runs, metadata revisions etc.

Before:
screen shot 2018-10-29 at 2 10 46 pm

After:
screen shot 2018-10-29 at 2 17 54 pm

### DIFF
--- a/refinery/core/api.py
+++ b/refinery/core/api.py
@@ -161,15 +161,6 @@ class SharableResourceAPIInterface(object):
                 )
 
         if cache_check is None:
-            # for ownership, don't check group perms
-            owned_res_set = Set(
-                get_objects_for_user(
-                    user,
-                    'core.share_%s' %
-                    self.res_type._meta.verbose_name,
-                    use_groups=False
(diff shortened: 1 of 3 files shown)


jkmarx:
@scottx611x Did a quick is_owner check and line ~701 core/api and core/views line 699

PR #3075, opened 2018-10-31
Jkmarx/highlight related nodes

### LINKED ISSUE #2977: Meta-data update - modify existing node's content (back-end)

### LINKED ISSUE #2982: Meta-data update - modify existing node's content - UI

jkmarx:
Resolves #2982 
Resolves #2977 

- Resolves any outstanding minor bugs due to no subtypes
- Show related nodes

complete-feature

### DIFF
--- a/refinery/data_set_manager/test_views.py
+++ b/refinery/data_set_manager/test_views.py
@@ -861,8 +861,9 @@ class NodeViewAPIV2Tests(APIV2TestCase):
         force_authenticate(get_request, user=self.user)
         get_response = self.view(get_request, file_node.uuid)
         self.assertEqual(get_response.status_code, 200)
-        derived_nodes_uuid = [node.uuid for node in Node.objects.filter(
-            assay=file_node.assay) if node.is_derived()]
+        assay_nodes = Node.objects.filter(assay=file_node.assay)
+        derived_nodes_uuid = [node.uuid for node in assay_nodes if
+                              node.is_derived() and not node == file_node]
         self.assertItemsEqual(get_response.data, derived_nodes_uuid)
 
     @mock.patch('data_set_manager.models.Node.update_solr_index')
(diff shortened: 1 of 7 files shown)


refinery/data_set_manager/views.py
```diff
@@ -1206,69 +1213,65 @@ def patch(self, request, uuid):
...
-                        uuid=node.file_uuid
-                    )
-                except (FileStoreItem.DoesNotExist,
-                        FileStoreItem.MultipleObjectsReturned) as e:
-                    logger.error(e)
-                    return Response('Missing file store item.',
-                                    status=status.HTTP_400_BAD_REQUEST)
-                else:
-                    file_store_item.delete_datafile()
+        if not data_set.get_owner() == request.user:
```
jkmarx: avoid indent

refinery/data_set_manager/test_views.py
```diff
@@ -861,9 +861,9 @@ def test_get_returns_derived_nodes_for_dataset_with_related_nodes(self):
         force_authenticate(get_request, user=self.user)
         get_response = self.view(get_request, file_node.uuid)
         self.assertEqual(get_response.status_code, 200)
-
-        derived_nodes_uuid = [node.uuid for node in Node.objects.filter(
-            assay=file_node.assay) if node.is_derived()]
+        assay_nodes = Node.objects.filter(assay=file_node.assay)
```
scottx611x: You could think about updating the query with an `exclude()` to make the list comprehension below simpler.
But this isn't crucial

```suggestion
        assay_nodes = Node.objects.filter(assay=file_node.assay).exclude(id=file_node.id)
        derived_nodes_uuid = [node.uuid for node in assay_nodes if node.is_derived()]
```

refinery/data_set_manager/views.py
```diff
@@ -1148,6 +1148,23 @@ def get_object(self, uuid):
             logger.error(e)
             raise APIException("Multiple objects returned.")
 
+    def get_first_annotated_node_from_solr_name(self, solr_name, node):
```
scottx611x: I think this would be more appropriate outside of the View class, maybe in a `utils.py`. Having this logic within the view makes it harder to test, and it doesn't utilize the reference to `self`.

For example, you wouldn't have to instantiate `NodeViewSet()` and give it a `request` just to test it anymore.

jkmarx: Sure.

Issue #2680, opened 2018-03-20
Home Page Rev 1


jkmarx:
- [x] Mock up home page
- [x] Investigate video placement and layout
- [x] Screen casts to start?
- [x] Carousel
- [x] Custom components 
    - [x] Platform, what am I
    - [x] Events
- [x] Buttons

-------- Alpha version --------
- [x] User testing

jkmarx:
@gmnelson Assigning you also based on our conversation.

PR #3117, opened 2018-11-20
Jkmarx/add chart component

jkmarx:
Ref #2680


refinery/core/views.py
```diff
@@ -69,6 +69,14 @@ def home(request):
     )
 
 
+def home_v2(request):
+    return render_to_response(
```
scottx611x: [The `context_instance` kwarg has been deprecated in Django 1.8](https://docs.djangoproject.com/en/1.10/releases/1.8/#dictionary-and-context-instance-arguments-of-rendering-functions)

I don't think you need the context of the request passed along to the template either. The following should be fine:

```suggestion
    return render('core/home_v2.html')
```

PR #3122, opened 2018-11-28
Jkmarx/add tools list landing page

jkmarx:
Ref #2680 

- Added tools list
Note: Just getting the components in and will focus on styling once all the pieces are in place.
screen shot 2018-11-28 at 2 17 40 pm


refinery/ui/source/js/home/ctrls/home-ctrl.spec.js
```diff
@@ -1,24 +0,0 @@
-(function () {
```
jkmarx: unused

refinery/tool_manager/serializers.py
```diff
@@ -68,6 +68,12 @@ class Meta:
 class ToolDefinitionSerializer(serializers.ModelSerializer):
     file_relationship = FileRelationshipSerializer()
     parameters = ParameterSerializer(many=True, allow_null=True)
+    workflow = serializers.SerializerMethodField()
+
+    def get_workflow(self, tool):
+        if tool.workflow:
+            return tool.workflow.uuid
+        return tool.workflow
```
scottx611x: Wouldn't the 2nd `return tool.workflow` always be `None`here?

jkmarx: Yes, am I being redundant?

scottx611x: Maybe a little. How about something like
```
def get_workflow(self, tool):
    return tool.workflow.uuid if tool.is_workflow() else None
```
or
```
def get_workflow(self, tool):
    if tool.is_visualization():
        return None  # Vis Tools have no associated Workflows
    return tool.workflow.uuid
```

jkmarx: I don't see how that improves anything.

PR #3177, opened 2019-01-17
Jkmarx/add config model

jkmarx:
- Add model and api for getting and updating site profiles.


refinery/core/models.py
```diff
@@ -2221,6 +2221,10 @@ class SiteProfile(models.Model):
 
     site = models.OneToOneField(Site, related_name='profile')
     repo_mode_home_page_html = models.TextField(blank=True)
+    about_markdown = models.TextField(blank=True)
+    intro_markdown = models.TextField(blank=True)
+    twitter_username = models.CharField(max_length=100, blank=True)
+    yt_videos = models.TextField(blank=True)
```
scottx611x: Maybe `youtube_videos` here instead?

Also rather than a `TextField` it looks like `yt_videos` could be better expressed through a relationship to another model? 

Alternatives to that ^^^ would be:
- a"pickle-able" field (requires a [third party library](https://pypi.org/project/django-picklefield/)) 
- a [JSONField](https://docs.djangoproject.com/en/1.9/ref/contrib/postgres/fields/#jsonfield) (requires django >= 1.9 and is PostgreSQL specific)

refinery/core/serializers.py
```diff
@@ -82,6 +82,31 @@ def partial_update(self, instance, validated_data):
...
+                                                     instance.about_markdown)
+        instance.intro_markdown = validated_data.get('intro_markdown',
+                                                     instance.intro_markdown)
+        instance.twitter_username = validated_data.get(
+            'twitter_username',  instance.twitter_username
+        )
+        instance.yt_videos = validated_data.get('yt_videos',
+                                                instance.yt_videos)
+        instance.save()
+        return instance
```
scottx611x: I think the shorthand for this would look like:  `instance.update(**validated_data)`

refinery/core/views.py
```diff
@@ -1040,6 +1041,68 @@ def post(self, request):
...
+    """
+    http_method_names = ["get", "patch"]
+
+    def get(self, request):
+        site_profile = SiteProfile.objects.get(site=get_current_site(request))
+        serializer = SiteProfileSerializer(site_profile)
+        return Response(serializer.data)
+
+    def patch(self, request):
+        if not request.user.is_superuser():
```
scottx611x: I'm sure you're thinking of adding this, but a test to ensure admin-only access would be great

refinery/core/models.py
```diff
@@ -2340,6 +2343,13 @@ def get_aggregate_sum(field_name):
         ]
 
 
+class SiteVideo(models.Model):
+    caption = models.TextField(blank=True)
+    site_profile = models.ForeignKey(SiteProfile)
+    source = models.CharField(max_length=100, blank=True)
```
scottx611x: One thing that could be useful for the `source` field is the use of the `choices` arg.

```
class SiteVideo(models.Model):
    YOUTUBE = 'YT'

    VIDEO_SOURCE_CHOICES = (
        (YOUTUBE, 'Youtube'),
        ...
    )
    ...
    source = models.CharField(choices=VIDEO_SOURCE_CHOICES, max_length=100, blank=True)
```

Even if we don't envision videos coming from too many different sources, this would still solve the problem of potentially having many differing entries like: `youtube`, `Youtube`, `YouTube` etc.

refinery/core/views.py
```diff
@@ -1040,6 +1043,104 @@ def post(self, request):
...
+        if site_videos is not None:
+            db_site_videos = SiteVideo.objects.filter(
+                site_profile=site_profile
+            )
+            new_video_list = json.loads(request.data.getlist('site_videos')[0])
+            new_video_list_ids = [vid.get('id') for vid in new_video_list]
+            # delete unused videos
+            for video in db_site_videos:
+                if video.id not in new_video_list_ids:
+                    video.delete()
```
scottx611x: If I'm reading this correctly an existing video will be deleted if its `id` isn't included in a given PATCH payload? So one would have to include all existing `site_videos` in a patch to preserve them even if they aren't being updated?

refinery/core/views.py
```diff
@@ -1040,6 +1043,104 @@ def post(self, request):
...
+            # add new videos or update exisiting videos
+            for new_video_data in new_video_list:
+                try:
+                    db_video = SiteVideo.objects.get(
+                        id=new_video_data.get('id')
+                    )
+                except:
```
scottx611x: It would be good to explicitly catch the `DoesNotExist` exception here

refinery/core/admin.py
```diff
@@ -154,6 +154,10 @@ class SiteStatisticsAdmin(AdminFieldPopulator):
     change_list_template = "admin/core/sitestatistics/change_list.html"
 
 
+class SiteVideoAdmin(AdminFieldPopulator):
+    list_display = ['site_profile']
```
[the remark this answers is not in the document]
jkmarx: @scottx611x It's displaying as I would expect below. Can you clarify?
screen shot 2019-01-25 at 9 33 00 am

PR #3194, opened 2019-02-01
Show only non-manager groups on user page.

### LINKED ISSUE #1307: Created manager group names are cryptic
Commit: 33de42c5dcd1e22fe17e4
### Steps to reproduce
- Create a new group
### Observed behavior

An admin group, titled something like `.Managers 8c74c9f0-01c2-4c85-bdb1-6d963e8e508c` is created.

[My account](https://test.stemcellcommons.org/users/1c5f033e-6546-46c6-937f-eccf6245dfc2/) lists two groups:
- [`Public2`](https://test.stemcellcommons.org/groups/8c74c9f0-01c2-4c85-bdb1-6d963e8e508c/), which I created
- and [`.Managers 8c74c9f0-01c2-4c85-bdb1-6d963e8e508c`](https://test.stemcellcommons.org/groups/81785b8e-a95e-48b5-bb4b-e21293396870/), whose name matches the ID of the former.
### Expected behavior

Perhaps it should be called something like "[new group name] Managers", or something else more human readable?

jkmarx:
Resolves #1307
Issue:
screen shot 2019-02-01 at 2 27 08 pm

Fixed:
screen shot 2019-02-01 at 2 59 23 pm

### DIFF
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -180,9 +180,9 @@ def user(request, query):
 
     # return all non-manager groups in profile
     groups = user.groups.exclude(name__contains='Managers')
-    return render_to_response('core/user.html',
-                              {'profile_user': user, 'user_groups': groups},
-                              context_instance=RequestContext(request))
+    return render(request,
+                  'core/user.html',
+                  {'profile_user': user, 'user_groups': groups})
 
 
 @login_required()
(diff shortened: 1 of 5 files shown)


refinery/core/views.py
```diff
@@ -177,7 +178,10 @@ def user(request, query):
     except User.DoesNotExist:
         user = get_object_or_404(UserProfile, uuid=query).user
 
-    return render_to_response('core/user.html', {'profile_user': user},
+    # return all non-manager groups in profile
+    groups = user.groups.exclude(name__contains='Managers')
```
scottx611x: How about `[group for group in user.groups if not group.is_manager_group()]`

I doubt it would change anytime soon, but relying on the name containing `Managers` could eventually fail without us knowing

jkmarx: I'll add a unit test

jkmarx: @scottx611x I moved it to a utils method due to issues accessing context in unit tests.

PR #3202, opened 2019-02-06
Jkmarx/fix dashboard search bugs

### LINKED ISSUE #2967: Dashboard text search doesn't interact with paging
* Specific code commit: f6b5665211815395d5ee7246a3e3574b1692eb72
* Version of the web browser and OS: FF on Max
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): Production Test

### Steps to reproduce

1. On the new Dashboard, do a text search

### Observed behavior

The found set hasn't actually changed: We still have the multiple paging buttons below, even though there was just one match. This is confusing, and really won't work if there is more than one page of search results.

screen shot 2018-08-29 at 10 41 30 am

### Expected behavior
* Probably the paging should correspond to the results found with the text search.

jkmarx:
Resolves #2967 
Also fixes this new bug with highlight
screen shot 2019-02-06 at 11 10 43 am

Fixed:
screen shot 2019-02-06 at 12 46 07 pm

### DIFF
--- a/refinery/ui/source/js/dashboard/controllers/data-sets-card-ctrl.js
+++ b/refinery/ui/source/js/dashboard/controllers/data-sets-card-ctrl.js
@@ -236,7 +236,7 @@
       if (query && query.length > 1) {
         vm.loadingDataSets = true;
         var apiRequest = new DataSetSearchApi(query);
-        apiRequest(100).then(function (response) {
+        apiRequest(200).then(function (response) {
           vm.dataSets = response.data;
           angular.copy(response.data, dataSetCardFactory.dataSets);
           dataSetCardFactory.dataSetStats.totalCount = response.data.length;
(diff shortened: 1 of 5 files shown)


refinery/ui/source/js/dashboard/controllers/data-sets-card-ctrl.js
```diff
@@ -229,13 +232,20 @@
     function searchDataSets (query) {
       // reset perm filter until we can search & check perms
       vm.groupFilter = { selectedName: 'All' };
+      vm.params = { limit: vm.itemsPerPage, offset: vm.pageStartOffset };
       if (query && query.length > 1) {
         vm.loadingDataSets = true;
         var apiRequest = new DataSetSearchApi(query);
-        apiRequest(200).then(function (response) {
+        apiRequest(100).then(function (response) {
```
scottx611x: Not understanding the need for a `100` instead of a `200` here

jkmarx: Oops, good catch. It's the number of ds returned.

PR #3246, opened 2019-03-07
Jkmarx/remove explore satori

jkmarx:
- Remove explore page (aka launchpad / satori)
    - Includes migration (remove tutorial for launchpad)
    - Remove unused common services

Resolves #3162 
These related issues were closed because of the feature removal.
Ref #643 
Ref #1015 
Ref #1530 
Ref #1540 
Ref #1549 
Ref #1551 
Ref #1587 
Ref #1773 
Ref #1888 
Ref #2137
Ref #2143 
Ref #2340 
Ref #2363 
Ref #2673 
Ref #2674 
Ref #2832 

Left dashboard.less as is, #3247

Issue #2618, opened 2018-02-22
RemovedInDjango19Warning for Tastypie


hackdna:
* Specific code commit: ae273b9
* Version of the web browser and OS: N/A
* Environment where the error occurred: Vagrant

### Steps to reproduce
```bash
(refinery-platform)vagrant@refinery:/vagrant/refinery$ ./manage.py test data_set_manager
```

### Observed behavior
```
2018-02-22 18:37:21 WARNING  py.warnings:9 <module>() - /home/vagrant/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/tastypie/fields.py:7: RemovedInDjango19Warning: django.utils.importlib will be removed in Django 1.9.
  from django.utils import importlib
```

### Expected behavior
No warning

### Notes
Likely requires removal of Tastypie

#### Remove the following API/Resources -> Add DRF APIs for those which require it and add unit tests
data_set_manager
- [x] Attribute Resource
- [x] Node Resource
- [x] Investigation Resource
- [x] Protocol Resource
- [x] Protocol Reference Resource
- [x] Protocol Reference Parameter Resource
- [x] Publication Resource
- [x] Study Resource
- [x] Assay Resource

core
- [x] Sharable Resource API (base class)
- [x] User Resource
- [x] User Profile Resource
- [x] Data Set Resource
- [x] Workflow Resource
- [x] Analysis Resource
- [x] Group Management
- [x] User Authentication
- [x] Invitation Reosurce
- [x] Extended Group Resource

docs
- [ ] Update, https://github.com/refinery-platform/refinery-platform/wiki/API

jkmarx:
Resolved by #3318

PR #3249, opened 2019-03-08
Remove assays api.

jkmarx:
Ref #2618 
~Need to merge post #3246~
Only used for Satori which was removed in #3246


refinery/core/api.py
```diff
@@ -38,9 +38,7 @@
 from tastypie.resources import ModelResource, Resource
 from tastypie.utils import trailing_slash
 
-from data_set_manager.api import (AssayResource, InvestigationResource,
-                                  StudyResource)
-from data_set_manager.models import Study
+from data_set_manager.api import (InvestigationResource, StudyResource)
```
hackdna: No need for parens for imports on a single line

refinery/core/test_api.py
```diff
@@ -407,14 +407,6 @@ def test_get_node_list_count(self):
...
         self.assertEqual(len(node_list_dict['objects']), 1)
 
-    def test_get_node_detail_fields(self):
-        response = self.api_client.get(api_uri(NodeResource, self.node.uuid))
-        node_fields = self.deserialize(response).keys()
-        expected_fields = NodeResource._meta.fields + [
-            'file_import_status', 'resource_uri'
-        ]
-        self.assertItemsEqual(node_fields, expected_fields)
-
```
hackdna: Should this be removed or just modified?

jkmarx: I'd say just remove it. In this milestone, all these tests will be deleted and new tests will be written for any new DRF APIs or have been written. We already have a Node API V2 which will need to be extended for more detail fields. There shouldn't be any roadblocks removing the node tasty pie API.

Issue #3250, opened 2019-03-11
Dashboard Enhance DS Details


jkmarx:
- [ ] Add DS Perms Icons
- [ ] Analyses Count
- [ ] DS Technology & Measurement
- [ ] DS Sources
- [ ] DS File Count

PR #3254, opened 2019-03-11
Remove unused annotations.

jkmarx:
Ref #2618
Annotations currently not used by client since Satori was removed.
In the future, we may need to add an API to support dashboard enhancements, Ref #3250


jkmarx:
@hackdna It's on the agenda for tomorrow, but the plan is to remove Neo4j this milestone. I plan to look into it tomorrow for an idea of how involved it'd be. This should also help with some performance issues.

PR #3257, opened 2019-03-12
Jkmarx/switch study resource v2

jkmarx:
Ref #2618

- Create Study API V2 to get specific studies based on DataSetUUID
- Add unit tests
- Update client to use Study API V2

** Keep study resource until provenance graph uses Node V2 API.**


refinery/data_set_manager/test_views.py
```diff
@@ -1615,3 +1616,96 @@ def test_metadata_revision_is_only_allowed_if_data_set_owner(self):
             "Metadata revision is only allowed for Data Set owners",
             response.content
         )
+
+
+class StudiesViewAPIV2Tests(APIV2TestCase):
+    def setUp(self, **kwargs):
+        super(StudiesViewAPIV2Tests, self).setUp(
+            api_base_name="studies/",
+            view=StudiesView.as_view()
```
hackdna: Lines 1624 and 1625 can be combined to fit on one line.

refinery/data_set_manager/urls.py
```diff
@@ -59,4 +59,5 @@
         AddFileToNodeView.as_view()),
     url(r'^nodes/(?P<uuid>' + UUID_RE + r')/$',
         NodeViewSet.as_view()),
+    url(r'^studies/$', StudiesView.as_view()),
```
hackdna: Why not use the DefaultRouter like other instances?

jkmarx: Routers require viewSets, and this api uses the APIView class. It's a limited API which only allows get lists, so I like to be explicit and use the APIView class.

PR #3258, opened 2019-03-13
Extend node v2 service.

jkmarx:
Ref #2618

- Extend Node API V2
- Add unit tests
- Update client (provenance graph) to use Node API V2

**Removing unused nodeResource and studyResource in separate (next) pull request.**


refinery/ui/source/js/file-browser/services/factory.js
```diff
@@ -20,7 +20,6 @@
     'assayFiltersService',
     'dataSetPropsService',
     'fileBrowserSettings',
-    'nodeService',
```
jkmarx: Unused dependency

PR #3261, opened 2019-03-14
Remove unused service.

jkmarx:
Ref #2618 

- Remove unused client-side service


jkmarx:
Mistake searching for wrong service name.

PR #3276, opened 2019-03-22
Hackdna/vagrantfile cleanup

hackdna:
Remove local_download field from Galaxy Instance model


hackdna:
No. I'll update the wiki.

PR #3286, opened 2019-03-26
Refactor other view sets to use helper.

jkmarx:
- Refactor to have viewsets use a helper method for grabbing data sets

**Branched off jkmarx/group-patch-perms**


refinery/core/views.py
```diff
@@ -593,16 +593,6 @@ class DataSetsViewSet(viewsets.ViewSet):
     http_method_names = ['get', 'delete', 'patch']
     lookup_field = 'uuid'
 
-    def get_object(self, uuid):
```
hackdna: This appears to be the correct method to use (https://www.django-rest-framework.org/api-guide/generic-views/#get_objectself). Why change?

hackdna: This appears to be the correct method to use (https://www.django-rest-framework.org/api-guide/generic-views/#get_objectself). Why remove?

jkmarx: We use data sets to check perms for many view sets and therefore need to retrieve it. The helper method keeps the api response the same across the board when looking up a data set. Not a necessary change, but it's redundant to keep this helper method when we have a utils. Also the api response will be the same when a data set has an invalid uuid or multiple objects are found. @hackdna

hackdna: I see. Since `get_object()` provides the canonical way to get object details in the API I would keep this method, and just call `get_data_set_for_view_set()` from it.

PR #3301, opened 2019-04-01
Jkmarx/extend group api to delete

jkmarx:
Ref #2618 
- Extend API to delete groups for managers only

**Branched off jkmarx/extend-group-api-return-all-user-groups**

### DIFF
--- a/refinery/core/views.py
+++ b/refinery/core/views.py
@@ -908,7 +908,7 @@ class AnalysisViewSet(APIView):
 
 class GroupViewSet(viewsets.ViewSet):
     """API endpoint for viewing groups."""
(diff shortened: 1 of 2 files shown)


refinery/core/views.py
```diff
@@ -939,6 +939,15 @@ def create(self, request):
             serializer.errors, status=status.HTTP_400_BAD_REQUEST
         )
 
+    def destroy(self, request, uuid):
+        user = request.user
+        group = self.get_object(uuid)
+        if not self.is_user_a_group_manager(user, group):
```
hackdna: I'd avoid a negation operator in the conditional for better readability.

PR #3302, opened 2019-04-01
Jkmarx/update client with group api enhancements

jkmarx:
Ref #2618
- Extend API to return can_edit field
- Update client to use Group V2 API for adding and deleting group and retrieving all user's groups.

**Branched off jkmarx/extend-group-api-to-delete**


refinery/core/serializers.py
```diff
@@ -121,12 +121,23 @@ def partial_update(self, instance, validated_data):
...
     member_list = serializers.SerializerMethodField()
     perm_list = serializers.SerializerMethodField()
+    can_edit = serializers.SerializerMethodField()
     uuid = serializers.SerializerMethodField()
     name = serializers.CharField(
         min_length=3,
         validators=[UniqueValidator(queryset=ExtendedGroup.objects.all())]
     )
 
+    def get_can_edit(self, group):
```
hackdna: Shouldn't this be a part of the user API?

jkmarx: It's part of the extended Group model.

refinery/core/serializers.py
```diff
@@ -121,12 +121,23 @@ def partial_update(self, instance, validated_data):
...
+        user = self.context.get('user')
+        if user is None:
+            return ''
```
hackdna: The name `get_can_edit()` suggests a boolean return value but it can also return an empty string?

PR #3309, opened 2019-04-03
Jkmarx/add group member api

jkmarx:
Ref #2618 
- Add groupMember resource to promote/demote/remove users
- Add unit tests


refinery/core/models.py
```diff
@@ -1863,6 +1863,12 @@ def get_managed_group(self):
         except:
             return None
 
+    def is_user_a_group_manager(self, user):
```
hackdna: Sounds like it should be a method on the User model.

refinery/core/serializers.py
```diff
@@ -128,6 +129,11 @@ class ExtendedGroupSerializer(serializers.ModelSerializer):
         validators=[UniqueValidator(queryset=ExtendedGroup.objects.all())]
     )
 
+    def get_manager_group_uuid(self, group):
```
hackdna: Seems redundant. Doesn't appear to be used anywhere.

refinery/core/views.py
```diff
@@ -1015,11 +1015,82 @@ def partial_update(self, request, uuid, format=None):
...
+        if group.is_user_a_group_manager(edit_user):
+            return HttpResponseBadRequest(
+                content="Managers can not leave group. Demote user first."
+            )
+        if len(group.user_set.all()) > 0:
+            group.user_set.remove(edit_user)
+            return Response(uuid)
+        return HttpResponseBadRequest(content="No users left in group.")
+
+    def is_user_authorized_to_edit(self, group, request_user, edit_user):
```
hackdna: Sounds like it should be a method either on the Group or the User model. This would help with reuse and simplify the function signature.

refinery/core/serializers.py
jkmarx: It's a field the client expects.

refinery/core/views.py
```diff
@@ -1015,11 +1015,82 @@ def partial_update(self, request, uuid, format=None):
...
+        try:
+            return ExtendedGroup.objects.get(uuid=uuid)
+        except ExtendedGroup.DoesNotExist as e:
+            logger.error(e)
+            raise Http404
+        except ExtendedGroup.MultipleObjectsReturned as e:
+            logger.error(e)
+            raise APIException("Multiple groups returned for this request.")
+
+    def get_user(self, id):
```
hackdna: Why use Django internal ID instead of UUID like everywhere else?

jkmarx: Legacy of previous api. url /group/:uuid/member/:id
I can update it to both use uuid after merging the other two branches into here.

refinery/core/views.py
jkmarx: It's only used for this delete request which is dependent on both the member and request user. I don't imagine it being used anywhere else, so it's fine here.

refinery/core/views.py
hackdna: Yes, it would be great to fix this.

refinery/core/views.py
```diff
@@ -1015,11 +1015,82 @@ def partial_update(self, request, uuid, format=None):
...
+                )
+        # not a group manager
+        return Response(uuid, status=status.HTTP_403_FORBIDDEN)
+
+    def delete(self, request, uuid, id):
+        # if group is a manager_group, user is demoted by removal
+        group = self.get_object(uuid)
+        edit_user = self.get_user(id)
+        if self.is_user_authorized_to_edit(group, request.user, edit_user):
+            return Response(uuid, status=status.HTTP_403_FORBIDDEN)
```
hackdna: If user is authorized to edit then return HTTP 403?

jkmarx: Forgot the un, will update the method name.

hackdna: I'd rather keep the name and simplify the logic.

refinery/core/views.py
jkmarx: @hackdna Users don't have UUID. User profiles have UUIDs. Simplicity wise, it would make sense to use IDs verses having to grab users through their profiles.

hackdna: I see. It would be great to switch to a custom User model as Two Scoops and [Django docs](https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#specifying-a-custom-user-model) suggest but it's out of the scope of this PR.

PR #3323, opened 2019-04-11
Fix bug with incorrect logic.

jkmarx:
- Resolves a bug with launching tools for logged in users.
- Mis-logic when updating to new apis https://github.com/refinery-platform/refinery-platform/pull/3290/files#diff-18b5350f22686e5318a05e9076918d3dR44

### DIFF
--- a/refinery/ui/source/js/tool-launch/ctrls/tool-launch-button-ctrl.js
+++ b/refinery/ui/source/js/tool-launch/ctrls/tool-launch-button-ctrl.js
@@ -80,7 +80,6 @@
    * Watchers
    * ---------------------------------------------------------
    */
-
     vm.$onInit = function () {
       $scope.$watchCollection(
         function () {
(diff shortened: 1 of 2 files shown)


refinery/ui/source/js/tool-launch/ctrls/tool-launch-button-ctrl.js
```diff
@@ -35,13 +36,12 @@
...
     $uibModal,
     visualizationService,
-    $rootScope,
     $window
   ) {
     var vm = this;
     vm.launchTool = launchTool;
     vm.needMoreNodes = needMoreNodes;
-    vm.userIsAnonymous = settings.djangoApp.userId !== undefined;
+    vm.userIsAnonymous = settings.djangoApp.userId === undefined;
```
hackdna: Would it make more sense to follow the existing usage of `isLoggedIn` (https://github.com/refinery-platform/refinery-platform/search?q=isLoggedIn&unscoped_q=isLoggedIn) and perhaps factor this out into a re-usable function or service?

Issue #2593, opened 2018-02-14
Update Django URL Patterns


jkmarx:
https://docs.djangoproject.com/en/1.8/releases/1.8/#django-conf-urls-patterns

- [ ] url patterns
- [ ] Passing a string as view to url()
Related to the previous item, referencing views as strings in the url() function is deprecated. Pass the callable view as described in the previous section instead.

Deprecated since version 1.8:
urlpatterns should be a plain list of django.conf.urls.url() instances instead.
https://docs.djangoproject.com/en/1.8/ref/urls/

PR #3341, opened 2019-04-24
Hackdna/update urls

hackdna:
Toward #2593

### DIFF
--- a/refinery/user_files_manager/urls.py
+++ b/refinery/user_files_manager/urls.py
@@ -6,11 +6,5 @@ from .views import UserFiles
 
 user_files_router = DefaultRouter()
 user_files_router.urls.extend([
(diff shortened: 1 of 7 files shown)


refinery/user_files_manager/urls.py
```diff
@@ -6,11 +6,5 @@
 
 user_files_router = DefaultRouter()
 user_files_router.urls.extend([
-    url(r'^files/$', UserFiles.as_view())  # Rooted under /api/v2
+    url(r'^files/$', UserFiles.as_view())
 ])
-user_files_url = url(r'^files/$',
```
jkmarx: Are these unused? They are still being imported above in config/urls

hackdna: Good catch, I'll update import statements before merging. I wonder why didn't flake8 pick this up.

Issue #3366, opened 2019-05-01
Newly Registered Users added to group cannot be demoted or deleted


jkmarx:
Specific code commit: fdf35ba
Version of the web browser and OS: Firefox 66.0.3
Environment where the error occurred (Vagrant VM and site conf mode or AWS instance):
https://test.stemcellcommons.org/   Also reproducible locally

### Steps to reproduce
Please list all the actions and the input data used:

1. In the admin interface, add a new users and add them to the test group
2. Promote
3. Demote

### Observed behavior
```angular.js:12185 DELETE http://192.168.50.50:8000/api/v2/groups/6997f024-3c93-4ad3-a820-7bb3840xxxxx/members/10/?format=json 404 (NOT FOUND).```

### Expected behavior
* Should be able to remove/delete

Not an issue with older existing users.

PR #3369, opened 2019-05-03
Jkmarx/fix routing issue for demoting users

jkmarx:
#3366 
- Turns out DRF doesn't fully support nested resources /groups/{{uuid}}/members/{{id}}
The bugs were inconsistent with new user ids not being found (via routing).
- ~/groups/{{uuid}}/members works so post handles updating the member_list~
- Remove group member api and extended the groups api to edit it's member list
- update unit tests and client side


refinery/core/views.py
```diff
@@ -1109,11 +1102,44 @@ def get_user(self, id):
             raise APIException("Multiple users returned for this request.")
 
     def post(self, request, uuid):
-        # if group is a manager_group, user is promoted by adding to user set
+        ''' handles promoting, demoting, and removing users. Note, adding
+        members to a group handled to Invitations API.'''
```
hackdna: Updates to an existing resource should be made via a PUT or a PATCH (https://restfulapi.net/rest-put-vs-post/).

refinery/ui/source/js/commons/services/group-member.js
```diff
@@ -5,17 +5,14 @@ angular
   .factory('groupMemberService', ['$resource', 'settings',
     function ($resource, settings) {
       return $resource(
-        settings.appRoot + settings.refineryApiV2 + '/groups/:uuid/members/:id/',
+        settings.appRoot + settings.refineryApiV2 + '/groups/:uuid/members/',
```
hackdna: Simply: `'/groups/:uuid/'` since it's the group resource that is getting updated.

PR #3401, opened 2019-06-19
Display the tools display name for analysis model.

### LINKED ISSUE #3379: Analyses Show Different Names Depending on Location
* Specific code commit: c0b2669c3fcd89b300209c48be80caf122b87730
* Version of the web browser and OS: Any
HSCI site (also an issue with the 1.6.9 test site)

### Steps to reproduce

1. Launch an analysis with a custom name

### Observed behavior
The name on the alert message and events card are correct. (Shows tool display_name)
The name on the analysis tab and analysis global popover is the default workflow name.

### Expected behavior
Custom name should show, so the tool's display_name vs the analysis name.

jkmarx:
Resolves #3379

Global
Screen Shot 2019-06-19 at 11 50 07 AM

File Browser
Screen Shot 2019-06-19 at 11 50 01 AM


jkmarx:
What do you mean by correctly set? It was a design decision to keep tool_display_name and analysis_name separate.

hackdna:
`Analysis.name` should be set to a custom name then displayed everywhere, correct?

jkmarx:
The tool.display_name is set to analysis.name if the user does not customize it. For any tool (vis and analysis) the display_name should be displayed in the UI. The analysis.name is automatic can be different than display_name and is used for other purposes beyond a name display.

hackdna:
There is a [one-to-one relationship](https://github.com/refinery-platform/refinery-platform/blob/develop/refinery/tool_manager/models.py#L294) between Tool and Analysis, so it sounds like `Tool.display_name` is redundant with respect to `Analysis.name`. Is there a way to simplify? If not, why not simply pull `display_name` from the Tool API directly?

jkmarx:
@hackdna We need an extra api call just to supply one field in the analysis monitoring UIs.

refinery/core/serializers.py
```diff
@@ -22,13 +24,22 @@ def to_representation(self, utc_time):
...
         model = Analysis
         fields = ('data_set_uuid', 'facet_name', 'name', 'owner', 'status',
-                  'summary', 'time_start', 'time_end', 'uuid', 'workflow')
+                  'summary', 'time_start', 'time_end', 'tool_display_name',
+                  'uuid', 'workflow')
+
+    def get_tool_display_name(self, analysis):
+        try:
+            tool = Tool.objects.get(analysis_id=analysis.id)
+        except:
```
hackdna: Bare `except:` should not be used.

Issue #2592, opened 2018-02-14
Move from to django_extensions UUIDField to Django UUIDField


jkmarx:
Note: switching from `UUIDField` provided by `django_extensions` to the native one may not trivial and will require a data migration since hyphenated UUIDs currently stored in `varchar(36)` columns will be truncated when moved to `char(32)` columns.
- [ ] [How to change text fields to a real UUID type for Django and PostgreSQL](http://baltaks.com/2015/08/how-to-change-text-fields-to-a-real-uuid-type-for-django-and-postgresql)
[Unexpected UUIDField upset](http://tech.yunojuno.com/unexpected-uuidfield-upset)

PR #3424, opened 2019-07-18
Migrate UUIDField uuid in Parameter model

hackdna:
Toward #2592

### DIFF
--- a/refinery/tool_manager/models.py
+++ b/refinery/tool_manager/models.py
@@ -508,8 +508,7 @@ class VisualizationTool(Tool):
         )
 
     def get_container_input_dict(self):
-        """
-        Create a dictionary containing information that Dockerized
+        """Create a dictionary containing information that Dockerized
         Visualizations will have access to
         """
         return {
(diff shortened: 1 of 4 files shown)


refinery/tool_manager/models.py
```diff
@@ -509,8 +508,7 @@ class Meta:
         )
 
     def get_container_input_dict(self):
-        """
-        Create a dictionary containing information that Dockerized
+        """Create a dictionary containing information that Dockerized
```
jkmarx: Why the format change?

refinery/tool_manager/tests.py
```diff
@@ -255,10 +255,8 @@ def setUp(self):
         self.tool_defs_url_root = '/api/v2/tool_definitions/'
 
         self.mock_parameter = ParameterFactory(
-            name="Test Param",
-            description="Test Param Description",
-            value_type=Parameter.STRING,
-            default_value="Coffee"
+            name="Test Param", description="Test Param Description",
```
jkmarx: The left format is easier to read. Why the change?  https://github.com/refinery-platform/refinery-platform/wiki/coding-style-guide#source-formatting

refinery/tool_manager/models.py
hackdna: Unnecessary newline (https://www.python.org/dev/peps/pep-0257/#id17)

jkmarx: It'd be great to automate any rules we want to enforce. I personally find the left easier to read.

refinery/tool_manager/tests.py
hackdna: It's customary in Python to list function arguments with no extra new lines (example: https://www.python.org/dev/peps/pep-0008/#indentation). Also, unnecessary newlines make functions/methods longer and more difficult to read (i.e., to fit on one screen).

PR #3426, opened 2019-07-22
Jkmarx/move dependencies to npm

### LINKED ISSUE #2605: Migrating away from Bower
https://bower.io/blog/2017/how-to-migrate-away-from-bower/

jkmarx:
Resolves #2605

- [ ] After merging into upgrade-node-js will spin up test instance.


jkmarx:
@hackdna Yes, we discussed this move to NPM vs something like Yarn months+ ago. https://bower.io/blog/2017/how-to-migrate-away-from-bower/  

Bower recommended yarn (2017) but since then NPM has updated to allow pinning. We discussed this ages ago in a previous Dev Meetings and settled on NPM instead of Yarn.

fabfile.py
```diff
@@ -150,13 +148,13 @@ def update_refinery():
...
     of grunt tasks
-    dependencies: update bower and pip dependencies
+    dependencies: update npm and pip dependencies
     migrations: apply migrations
     """
     puts("Relaunching Refinery")
     with cd(os.path.join(env.refinery_app_dir, "ui")):
         if dependencies:
-            run("bower update --config.interactive=false")
+            run("npm update --config.interactive=false")
```
hackdna: This should probably be `npm update --progress false`. However, I would just remove `relaunch_refinery()` - I think it has not been used in a long time.

jkmarx: I've never used it, so I can remove it.

Issue #2636, opened 2018-02-28
Update get_absolute_url() to raise an exception for empty input


hackdna:
Currently, it returns an empty string for empty string and `None` for `None` (see `test_get_absolute_url_with_empty_input()`). This means that the result of `get_absolute_url()` does not always pass `is_absolute_url()`.

`get_absolute_url()` should either return an absolute URL or raise an exception. Note: this requires exception handling for all usage instances of `get_absolute_url()`.

Reference: https://github.com/refinery-platform/refinery-platform/pull/2634#discussion_r171136928

Also, it might be worth renaming `get_absolute_url()` to `build_absolute_url()` for clarity.

PR #3448, opened 2019-09-10
Get Absolute URL Always Returns Absolute URL or Raises Error

ilan-gold:
This PR closes out #2636 by forcing get_absolute_url() to always return an absolute url or raise an error

### DIFF
--- a/refinery/core/utils.py
+++ b/refinery/core/utils.py
@@ -116,7 +116,7 @@ def invalidate_cached_object(instance, is_test=False):
         return mc
 
 
-def get_absolute_url(string):
+def build_absolute_url(string):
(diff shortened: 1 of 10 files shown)


refinery/core/utils.py
```diff
@@ -116,19 +116,20 @@ def invalidate_cached_object(instance, is_test=False):
         return mc
 
 
-def get_absolute_url(string):
+def build_absolute_url(string):
```
jkmarx: Why the name change?

ilan-gold: ilya mentioned it in the ticket and it is more descriptive of what the function does.

ilan-gold:
closes #2636

Issue #3455, opened 2019-10-01
Pass in Context for ToolSerializer


ilan-gold:
Noticed while working on #2619 
* Version of the web browser and OS: 
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): Any

### Steps to reproduce
Please list all the actions and the input data used:

1.  Launch an analysis and look in the logs for a caught error about the user being None https://github.com/refinery-platform/refinery-platform/blob/37a93141a777a2ae61e21050982dd72af8d28a9c/refinery/core/serializers.py#L78

### Observed behavior
* AttribtueError

### Expected behavior
* Our handling of context should be {'request':request} here to prevent the error

jkmarx:
@ilan-gold How are you/user hitting this error? Could you give a specific example. What do the DRF docs say? In general, I thought we wanted to minimize passing the context to the serializer but we have some use cases.

ilan-gold:
@jkmarx https://github.com/refinery-platform/refinery-platform/blob/develop/refinery/tool_manager/views.py#L118 calls DatSetSerializer (via ToolSerliazer) without context.  When I was looking in to it, I noticed that the convention was inconsistent - the example on DRF's website uses context={'request' = request} and i think this makes sense.  but we, for example, do something different sometimes (like in https://github.com/refinery-platform/refinery-platform/blob/0683eba983b771f3941c7c13a1a2e0b9e316fd76/refinery/core/views.py#L1056)

jkmarx:
@ilan-gold For the data_set example above, the context required more than the request in the context. 

https://www.django-rest-framework.org/api-guide/serializers/#including-extra-context

From the docs, "You can provide arbitrary additional context by passing a context argument when instantiating the serializer. " The purpose of context is to provide additional information not necessarily the request.

Issue #2619, opened 2018-02-23
Upgrade to Python 3.x


hackdna:
Required for #2918 

evaluate existing codebase for Python 3 compatibility:

* 3rd party modules (Run Python 3 travis builds)
* unicode string handling
* print statements
* urlparse -> urllib.parse (https://pymotw.com/3/urllib.parse/)

create a plan to make it compatible along with developing guidelines for writing new code

Tools:

* Six (https://pythonhosted.org/six/)
* Automated Python 2 to 3 code translation (https://docs.python.org/2/library/2to3.html)
* PyCharm can highlight problematic code (https://superuser.com/questions/920962/pycharm-turn-on-python-3-code-checking-in-a-python-2-project)

Reference:

* Two Scoops book, appendix F
* https://docs.python.org/3/howto/pyporting.html
* https://docs.djangoproject.com/en/1.8/topics/python3/
* http://python-future.org/compatible_idioms.html
* http://www.diveintopython3.net/porting-code-to-python-3-with-2to3.html
* https://medium.com/@anthonypjshaw/oh-no-this-package-is-python-2-only-8e6316f9a02
* https://blogs.dropbox.com/tech/2019/02/incrementally-migrating-over-one-million-lines-of-code-from-python-2-to-python-3/

Unicode string handling (http://python-future.org/unicode_literals.html); for example:

`updated_dict[curr_id]['uuid'] = unicode(str(uuid.uuid4()))`
or
`from django.utils.encoding import python_2_unicode_compatible`

3rd party modules:

* virtualenvwrapper depends on python 2.7 but can be used to create python 3 envs
* evaluate pyenv

Proposed Plan:

- [x] Update broken/missing Python code ahead of time 

Two phases (each roughly corresponding to a separate branch/merge):
**Python2 In-Place Updates**
- [x] Remove django-smtp-ssl
- [x] Remove djangular (#3452)
- [x] Replace librabbitmq with py-amqp

**Python3 Update**
- [ ] Syntax 

- dict iteration
- `unicode` -> `str`
- `raw_input` -> `input`
- `urlparse` -> `urllib`
- `urllib2` -> `urllib`
- getting rid of `from __future__ import unicode_literals`
- assertItemsEqual -> assertCountEqual (https://hg.python.org/cpython/rev/d9921cb6e3cd)
- content in responses needs to be cast to `str` where appropriate https://docs.djangoproject.com/en/1.8/ref/request-response/#django.http.HttpResponse.content
[...]

PR #3457, opened 2019-10-03
Update Buggy Code in Advance of Python 3 Update

ilan-gold:
There are a few things/bugs here related to #2619 that should be fixed before updating to Python3.  

1) We don't save() the status during the `test_tasks.py` test which means we are comparing None and int when `_check_galaxy_history_state` runs, which is problematic in Python3 and not exactly what we want these tests to do anyway in Python2.
2) If a use logs out in another webpage and then tries to access a dataset in another page, I noticed we were returning a 500 error, so I wrapped that in a try-catch.
3) We don't pass context into ToolSerializer, which we should since it raises an exception that is caught (#3455).


refinery/core/views.py
```diff
@@ -207,8 +207,16 @@ def data_set_slug(request, slug):
 def data_set(request, data_set_uuid, analysis_uuid=None):
     data_set = get_object_or_404(DataSet, uuid=data_set_uuid)
     public_group = ExtendedGroup.objects.public_group()
-
-    if not request.user.has_perm('core.read_meta_dataset', data_set):
+    try:
+        perms = request.user.has_perm('core.read_meta_dataset', data_set)
+    except User.DoesNotExist:
```
jkmarx: If the user is not logged in, they are the anon user. What's throwing the exception exactly?

Issue #3435, opened 2019-08-28
Analysis import fails due to celery SoftTimeLimitExceeded 


gmnelson:
* Specific code commit: 3abb3aa
* Version of the web browser and OS: chrome, macOS
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): hsci.stemcellcommons.org 

### Steps to reproduce
Please list all the actions and the input data used:

1. Launch analysis that completes successfully in Galaxy

### Observed behavior
* Analysis still fails Refinery-side, **without email notification**
* Logs suggest celery SoftTimeLimitExceeded while importing outputs back into Refinery

### Expected behavior
* Outputs should import successfully and analysis should register as successful in Refinery

jkmarx:
Traceback from celery:
```2019-08-29 21:15:20,461 ERROR    analysis_manager.tasks:51 on_failure[....4856fa] - Setting status of analysis 'ChIP-seq Peak Calling - Human - 2019/08/29 15:50:17 - Gnelson - gnelson - Galaxy workflow execution for: ChIP-seq Peak Calling - Human' to failure
2019-08-29 21:15:20,484 ERROR    celery.worker.job:282 log() - Task analysis_manager.tasks.run_analysis[.....4856fa] raised unexpected: SoftTimeLimitExceeded()
Traceback (most recent call last):
  File "/home/ubuntu/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/ubuntu/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/srv/refinery-platform/refinery/analysis_manager/tasks.py", line 303, in run_analysis
    _galaxy_file_export(analysis_uuid)
  File "/srv/refinery-platform/refinery/analysis_manager/tasks.py", line 184, in _galaxy_file_export
    galaxy_export_tasks = _get_galaxy_download_task_ids(analysis)
  File "/srv/refinery-platform/refinery/analysis_manager/tasks.py", line 462, in _get_galaxy_download_task_ids
    tool.create_analysis_output_node_connections()
  File "/srv/refinery-platform/refinery/tool_manager/models.py", line 818, in create_analysis_output_node_connections
    for galaxy_dataset in self._get_galaxy_history_dataset_list():
[...]

PR #3461, opened 2019-10-08
Restructure Analysis Metadata Fetching for Celery

ilan-gold:
Speed up Celery by storing results of multiple API calls (that don't change) in a table / using existing stored information instead of making the request.  This addresses both #2642 and #3435.  The speed up is drastic, and gets larger (in absolute terms) the more requests that are made.  Benchmarking is in #3435 but it appears that the speed up is on the magnitude of 5x or greater for large calls.


refinery/tool_manager/models.py
```diff
@@ -793,8 +794,11 @@ def create_analysis_output_node_connections(self):
         """Create the AnalysisNodeConnection objects corresponding to the
         output Nodes (Derived Data) of a WorkflowTool launch
         """
-        exposed_workflow_outputs = self._get_exposed_galaxy_datasets()
-        for galaxy_dataset in self._get_galaxy_history_dataset_list():
+        exposed_dataset_list = self._get_galaxy_history_dataset_list()
+        exposed_workflow_outputs = self._get_exposed_galaxy_datasets(
+            exposed_dataset_list=exposed_dataset_list
+        )
```
ilan-gold: @hackdna these two functions are only every used here, so my question is similar to my other comment.  not being familiar enough with the system, i wonder if they should be removed and all their functionality be wrapped up into create_analysis_output_node_connections() ?  is there a reason to keep around these functions that specifically make the set of API calls/data filtering they make?  i aimed for maximizing speed with this PR rather than refactoring completely in this PR.

refinery/analysis_manager/tasks.py
```diff
@@ -473,7 +473,11 @@ def _get_galaxy_download_task_ids(analysis):
     tool.create_analysis_output_node_connections()
     galaxy_instance = analysis.workflow.workflow_engine.instance
     try:
-        download_list = tool.get_galaxy_dataset_download_list()
+        download_list = tool_manager.models.AnalysisNodeConnection.\
+            objects.filter(
+                is_refinery_file=True, analysis=analysis,
+                direction=OUTPUT_CONNECTION
+            )
```
ilan-gold: @hackdna I'm not familiar with the history of this enough to make this call, but the get_galaxy_dataset_download_list() function is used nowhere else.  can it be removed?  would it be better to replace it with the output of the the AnalysisNodeConnection.objects.filter() operation (maybe along with some checking to make sure the list is non-empty)?

refinery/core/models.py
```diff
@@ -1730,6 +1729,12 @@ class AnalysisNodeConnection(models.Model):
                                            default=False)
     galaxy_dataset_name = models.CharField(null=True, blank=True,
                                            max_length=250)
+    # state from galaxy
+    state = models.CharField(null=True, blank=True, max_length=100)
+    # size of the file
+    file_size = models.IntegerField(null=True, blank=True)
+    # galaxy dataset id
+    dataset_id = models.CharField(null=True, blank=True, max_length=100)
```
hackdna: Are these really necessary? AnalysisNodeConnection already has a lot of fields that appear to be used only once (during analysis).

refinery/tool_manager/models.py
hackdna: I am not familiar with this part of the system either but I suspect the whole `tool_manager` app probably requires a re-write.

ilan-gold: I think that gets to the heart of why this issue has taken so long to resolve - I am walking a line between solving the bug at hand and refactoring large swaths of code, which was not what this was originally supposed to be.  We can do that, but I was always under the impression that was going to be the next step as it might take some time to do that.

refinery/core/models.py
ilan-gold: we could change the return value of the function create_analysis_output_node_connections() https://github.com/refinery-platform/refinery-platform/pull/3461/files#diff-49577ffb85a5a9e95447df9c8ccc34eeR473 to be a dict that represents the objects of the filter operation on the AnalysisNodeConnection at https://github.com/refinery-platform/refinery-platform/pull/3461/files#diff-49577ffb85a5a9e95447df9c8ccc34eeR476.  Then we wouldn't have to store the data in the database for re-use, although it would make the create_analysis_output_node_connections() a little opaque in terms of what it does.

refinery/tool_manager/models.py
hackdna: One of the reasons this code needs cleanup is that band-aid solutions were often used. Milestone 1.7.2 has just started, so might be a good time to refactor or rewrite.

refinery/core/models.py
hackdna: I would say it's a better choice than making database schema even more complicated.

refinery/tool_manager/models.py
ilan-gold: ok @hackdna i think then we should leave this alone in this pull request since it's not really relevant to the solution - i'll update the above create_analysis_output_node_connections() function to return the values mentioned above (and remove the database schema change).  my next pull request on this matter will then be the refactor; that way, i can separate the more cosmetic change from this, which is a bug-fix.  does that sound like a good plan?

hackdna: OK, as long as you follow up :)

refinery/config/settings/base.py
```diff
@@ -315,7 +315,7 @@ def get_setting(name, settings=local_settings, default=None):
 CELERYD_MAX_TASKS_PER_CHILD = get_setting("CELERYD_MAX_TASKS_PER_CHILD")
 CELERY_ROUTES = {"file_store.tasks.FileImportTask": {"queue": "file_import"}}
 CELERY_ACCEPT_CONTENT = ['pickle']
-CELERYD_TASK_SOFT_TIME_LIMIT = 60  # seconds
+CELERYD_TASK_SOFT_TIME_LIMIT = 180  # seconds
```
hackdna: You've mentioned in the description that the new changes provide a "5x or greater" speed up, so is this really necessary?

refinery/tool_manager/models.py
```diff
@@ -673,6 +673,7 @@ class WorkflowTool(Tool):
     REVERSE = "reverse"
     TOOL_ID = "tool_id"
     WORKFLOW_OUTPUTS = "workflow_outputs"
+    invocation = models.TextField(null=True, blank=True)
```
hackdna: Would it be possible to pass around a variable instead?

refinery/analysis_manager/tasks.py
ilan-gold: i use the term "filter" loosely to just mean that we do some subsetting and transformation

refinery/tool_manager/models.py
```diff
@@ -1095,7 +1102,7 @@ def _get_galaxy_history_dataset_list(self):
         ]
         return retained_datasets
 
-    def _get_exposed_galaxy_datasets(self):
+    def _get_exposed_galaxy_datasets(self, exposed_dataset_list=None):
```
hackdna: Is this argument necessary? I only see this function called with `exposed_dataset_list = self._get_galaxy_history_dataset_list()` in `create_analysis_output_node_connections()`

ilan-gold: it is needed to prevent repeated calls to _get_galaxy_history_dataset_list, which is a big API call.

refinery/tool_manager/models.py
ilan-gold: this is a serious pain-point in terms of usage - everytime `_get_galaxy_workflow_invocation` is called by `_get_workflow_step`, this function is called.  `_get_workflow_step` is used quite a few times.  this is the sort of thing that would get fixed in a refactor.  unlike the other fields i added in the `AnalysisNodeConnection` this one is used repeatedly and makes sense as part of the tool as it is a clear property of it.

refinery/analysis_manager/tasks.py
hackdna: OK but I was referring to the replacement code here which is just filtering (vs a bunch of operations in `get_galaxy_dataset_download_list()`). If this is equivalent you can remove the function. Another question is: will this help you with refactoring later?

refinery/tool_manager/models.py
hackdna: Right but that call is made anyway (https://github.com/refinery-platform/refinery-platform/pull/3461/files#diff-8beee8c9215c07566345b13603f3b9c6R797) and the result is passed as an argument on the next line.

ilan-gold: ah i see what you are saying - will update

refinery/config/settings/base.py
ilan-gold: I suppose not

refinery/tool_manager/models.py
hackdna: OK then just drop `null=True` - it is not recommended for `TextField` (see Two Scoops book, 6.2.4)

refinery/analysis_manager/tasks.py
ilan-gold: likely, yes. it depends on how i refactor, but i assume this replacement will be helpful. removing the get_galaxy_dataset_download_list() sped things up considerably. in this case i'll remove the function altogether.

refinery/tool_manager/models.py
```diff
@@ -1164,10 +1175,19 @@ def _get_galaxy_workflow_invocation(self):
         """
         Fetch our Galaxy Workflow's invocation data.
         """
-        return self.galaxy_connection.workflows.show_invocation(
-            self.galaxy_workflow_history_id,
-            self.get_galaxy_dict()[self.GALAXY_WORKFLOW_INVOCATION_DATA]["id"]
-        )
+        # separate if-then assignment needed to avoid using the dict stored
+        # in self.invocation before .save() is called
+        if self.invocation == '' or self.invocation is None:
```
hackdna: Comparing to `None` is not necessary if `invocation` cannot be `null` in the database (it shouldn't be).

refinery/tool_manager/models.py
ilan-gold: ok, yea, jennifer and i talked about that `blank=True` and `null=True` coexistence at length (since it appears all over the code base) and i remember the outcome of the conversation being to wait for your opinion since so much of it went against convention but we had it nonetheless.  good to know :)

refinery/tool_manager/models.py
ilan-gold: ah, @hackdna, i stand corrected.  the AnalysisNodeConnection objects correspond to the output of `_get_galaxy_history_dataset_list` and not `_get_exposed_galaxy_datasets` (which is a subset) - therefore it is necessary actually, as far as i can tell, as both calls need to be made in `create_analysis_output_node_connections`

ilan-gold: barring the refactor, i would just leave this.

refinery/analysis_manager/tasks.py
```diff
@@ -470,29 +470,20 @@ def _get_galaxy_download_task_ids(analysis):
...
     galaxy_instance = analysis.workflow.workflow_engine.instance
-    try:
-        download_list = tool.get_galaxy_dataset_download_list()
-    except galaxy.client.ConnectionError as exc:
-        error_msg = \
-            "Error downloading Galaxy history files for analysis '%s': %s"
-        logger.error(error_msg, analysis.name, exc.message)
-        analysis.set_status(Analysis.FAILURE_STATUS, error_msg)
-        analysis.galaxy_cleanup()
-        return task_id_list
```
hackdna: Doesn't `create_analysis_output_node_connections()` still raise exceptions?

ilan-gold: yes - i will update

refinery/tool_manager/models.py
hackdna: I see, in that case I think making this argument required would be a better choice: it would simplify the logic here and it would be a step toward refactoring.

ilan-gold: if it's all the same to you, i would just leave it.  i am not sure what the refactor is going to look like yet and it might involve changing these functions and their names altogether.  on top of that i would have to fix the tests

refinery/core/models.py
```diff
@@ -1067,7 +1066,7 @@ def __str__(self):
 
     def get_expanded_workflow_graph(self):
         return tool_manager.utils.create_expanded_workflow_graph(
-            ast.literal_eval(self.workflow_copy)
+            json.loads(self.workflow_copy)
```
hackdna: How will this work with existing analyses? The value stored in `workflow_copy` was a Python dict which is not going to be processed correctly by `json.loads()`.

ilan-gold: yes, this fell by the wayside through testing but i remember this now.  do you think we should check type here or remove the json.dumps() i added around self.workflow_copy? i think the later.

hackdna: I would guess a bunch of other code is relying on the existing serialization format, so it'd be safer to avoid making this change in this PR.

refinery/tool_manager/tests.py
```diff
@@ -1709,7 +1699,7 @@ def test_workflow_tool_launch_valid_workflow_object(self):
         self.assertEqual(self.tool.get_owner(), self.user)
         self.assertTrue(self.tool.is_workflow())
         self.assertEqual(
-            ast.literal_eval(self.tool.analysis.workflow_copy),
+            json.loads(self.tool.analysis.workflow_copy),
```
hackdna: This should be reverted too.

Issue #2539, opened 2018-01-25
Separate Django unit tests from integration and functional tests


hackdna:
Running the whole test suite takes about 15 minutes. It should be possible to run just the unit tests quickly (ideally, under one minute) for easy verification of changes and troubleshooting.

Example:
https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/unit-tests/#running-the-selenium-tests

PR #3464, opened 2019-10-10
Python2 Package Updates/Changes

ilan-gold:
1) Python3 uses a different hashing method than Python2 altogether so I decided to get rid of the `hash` function altogether and use a reliable library
2) Remove `librabbitmq` (which is Python2 only) and rely solely on `ampq`
3) Add testing of the `pysam` module in advance of its upgrade in Python3


requirements.txt
```diff
@@ -1,5 +1,8 @@
-# workaround for https://github.com/celery/celery/issues/3225
-amqp>1.4,<2.0
+# recommended broker is py-amqp(amqp) for celery (librabbitmq is python2 only):
+# https://github.com/celery/librabbitmq/issues/118#issuecomment-385461136
+# https://github.com/celery/librabbitmq/issues/119#issuecomment-388327748
+# http://docs.celeryproject.org/en/3.1/userguide/optimizing.html?highlight=pyamqp#librabbitmq-python-2-only
+amqp==1.4.9
```
hackdna: This is not needed: `amqp` is installed by `kombu` which is installed by `celery`. Listing `amqp` explicitly was only necessary as a workaround for a bug in `librabbitmq`.

ilan-gold: that makes sense - i will remove then

refinery/config/config.json.erb
```diff
@@ -6,7 +6,7 @@
   "ALLOWED_HOSTS": ["localhost", "127.0.0.1", "192.168.50.50"
     <%= @site_url ? ', "' + @site_url +  '"' : "" %>
     ],
-  "BROKER_URL": "amqp://guest:guest@localhost:5672//",
+  "BROKER_URL": "pyamqp://guest:guest@localhost:5672//",
```
hackdna: Not necessary: `amqp://` will use the correct transport automatically

refinery/data_set_manager/test_tasks.py
```diff
@@ -0,0 +1,32 @@
...
+            FileStoreItem.objects.get(
+                uuid=self.file_store_item.uuid).source,
+            'data_set_manager/test-data/test.bam.bai'
+        )
+
+    def test_index_bam_success_file_creation(self):
+        generate_bam_index(self.file_store_item.uuid, self.test_bam_file)
+        self.assertTrue(
+            os.path.exists(self.test_bam_file + '.bai')
+        )
```
hackdna: This is unnecessary because code under test is broken: (#2713 and #2718).

Also, for future reference:
* this is not a unit test (it requires a file system) and we've decided to keep those separate from integration/functional tests (#2539)
* we don't need to test 3rd party code (pysam) - mocking those calls should be used instead

ilan-gold: I was aware at the time I wrote this that it was not a unit test - I wanted nonetheless to make sure this function was working through the upgrade of it/python, especially since its documentation is not great.  Should I get rid of the test all together or keep it and mock the pysam function (which would make it unit but also then not give me the test of functionality I intended)?

hackdna: I would remove the test because the code under test is broken and will require a significant rewrite.

refinery/file_store/utils.py
```diff
@@ -74,7 +75,7 @@ def get_available_name(self, name, max_length=None):
         # TODO: remove this when get_directory_name() is removed in Django 2.0
         name = os.path.normpath(name)
         # create a hashed directory structure
-        hashcode = hash(name)
+        hashcode = int(hashlib.md5(name).hexdigest(), 16)
```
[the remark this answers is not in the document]
ilan-gold: yes you're right about the bytes encoding - surprised that didn't break the python3 tests, but maybe it casts under the hood.  will simplify in any case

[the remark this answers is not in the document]
hackdna: Thanks. Looks like the tests ran under Python 2: https://api.travis-ci.org/v3/job/598333510/log.txt

PR #3471, opened 2019-10-18
Rename Files in Flight

ilan-gold:
In order to solve #3435 we need to be able to rename things in flight so that we do not synchronously copy things in s3 and timeout with that

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -1440,17 +1440,28 @@ class Analysis(OwnableResource):
 
     def _prepare_annotated_nodes(self, node_uuids):
         """
-        Wrapper method to ensure that `rename_results` is called before
-        index_annotated_nodes_selection.
-
-        If `rename_results` isn't executed before
-        `index_annotated_nodes_selection` we end up indexing incorrect
-        information.
+        Wrapper method to ensure that auxiliary nodes are generated before
+        indexing annotated nodes
 
         Call order is ensured through:
         core.tests.test__prepare_annotated_nodes_calls_methods_in_proper_order
         """
-        self.rename_results()
+        for result in self.results.all():
+            try:
+                item = FileStoreItem.objects.get(uuid=result.file_store_uuid)
+            except (FileStoreItem.DoesNotExist,
+                    FileStoreItem.MultipleObjectsReturned) as exc:
+                logger.error("Error renaming analysis result '%s': %s",
+                             result, exc)
+                break
+            try:
+                node = Node.objects.get(file_item=item)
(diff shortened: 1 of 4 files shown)


refinery/core/models.py
```diff
@@ -1346,43 +1346,6 @@ def send_email(self):
...
-                try:
-                    zipfile = FileType.objects.get(name='ZIP')
-                except (FileType.DoesNotExist,
-                        FileType.MultipleObjectsReturned) as exc:
-                    logger.error("Error renaming HTML to zip: %s", exc)
-                else:
-                    if item.filetype == zipfile:
-                        item.rename_datafile(''.join([root, '.zip']))
-            else:
-                item.rename_datafile(result.file_name)
```
hackdna: It looks like you've lost this functionality. Have you tested this branch with any workflows?

refinery/file_store/tasks.py
```diff
@@ -27,7 +27,7 @@ class FileImportTask(celery.Task):
 
     soft_time_limit = 3600  # 1 hour
 
-    def run(self, item_uuid):
+    def run(self, item_uuid, target_name=None):
```
hackdna: If you are changing the API then it should work for all cases. Right now it only works for two out of six which is confusing and would be hard to troubleshoot.

refinery/core/models.py
ilan-gold: i don't think the functionality is not lost - this appears to be redundant data processing given that something identical happens here: https://github.com/refinery-platform/refinery-platform/blob/develop/refinery/analysis_manager/tasks.py#L498.  the only difference is that this pull request now sets the name in flight: https://github.com/refinery-platform/refinery-platform/pull/3471/files/6e8fc3b2b1047e3429121406983fe438f88cce8a#diff-49577ffb85a5a9e95447df9c8ccc34eeR523

refinery/file_store/tasks.py
ilan-gold: that's fair, i can update that

refinery/core/models.py
ilan-gold: yes, i can do that

refinery/file_store/tasks.py
hackdna: Thanks, could you also test all these cases (manual testing is fine)? File import is a core functionality.

ilan-gold: yes will do

refinery/core/models.py
hackdna: Also, can `FileStoreItem.rename_datafile()` be removed now?

ilan-gold: We can remove it, yes

PR #3474, opened 2019-10-22
Update to Python3.5

ilan-gold:
This pull request contains the update to python3.5 outlined here: #2619


deployment/puppet/refinery/manifests/apache2.pp
```diff
@@ -32,10 +32,11 @@
   }
 
   class { 'apache::mod::wsgi':
-    mod_path     => 'mod_wsgi.so',
-    package_name => 'libapache2-mod-wsgi',
+    mod_path         => 'mod_wsgi.so',
+    package_name     => 'libapache2-mod-wsgi-py3',
+    wsgi_python_home => "${virtualenv}",
+    wsgi_python_path => "${django_root}",
```
hackdna: Are these options necessary? They already exist in `apache::vhost`

deployment/puppet/refinery/manifests/python.pp
```diff
@@ -7,16 +7,21 @@
   $django_root         = $refinery::params::django_root,
 ) inherits refinery::params {
 
+
   class { '::python':
+    version    => 'python3.5',
```
hackdna, line 11: Why not use the latest version 3.8?

deployment/puppet/refinery/manifests/apache2.pp
```diff
@@ -132,7 +133,8 @@
       'refinery' => {
         'user'        => $app_user,
         'group'       => $app_group,
-        'python-path' => "${django_root}:${virtualenv}/lib/python2.7/site-packages",
+        'python-home' => "${virtualenv}",
+        "python-path" => "${django_root}"
```
hackdna: It's best to use double quotes for interpolated strings and single quotes for regular strings. Also, add the trailing comma.
`"python-path" => "${django_root}",`
https://puppet.com/docs/puppet/5.5/style_guide.html

deployment/puppet/refinery/manifests/python.pp
```diff
@@ -41,17 +46,20 @@
   }
   ->
   python::virtualenv { $virtualenv:
-    ensure  => present,
-    owner   => $app_user,
-    group   => $app_group,
-    require => [
+    version    => '3.5',
```
hackdna: Is this not inherited from `::python` class?

deployment/puppet/refinery/manifests/python.pp
```diff
@@ -41,17 +46,20 @@
...
+    ensure     => present,
+    owner      => $app_user,
+    group      => $app_group,
+    distribute => false,
```
hackdna: Is this necessary?

deployment/puppet/refinery/manifests/python.pp
```diff
@@ -72,12 +80,18 @@
...
   }
 
-  package { 'virtualenvwrapper': }
+  exec { 'get virtualenvwrapper':
+    command     => "python3 -m pip install virtualenvwrapper",
+    path        => [ '/bin/', '/sbin/' , '/usr/bin/', '/usr/sbin/' ],
+    user        => $app_user,
+    group       => $app_group,
+    require     => Python::Virtualenv[$virtualenv],
+  }
```
hackdna: There should be no need for this since `virtualenvwrapper` package is available.

Btw, `python::pip` installs and manages packages from pip
https://forge.puppet.com/puppet/python/reference#pythonpip
In general, `exec` resource type should be used as a last resort

deployment/puppet/refinery/manifests/python.pp
```diff
@@ -72,12 +80,18 @@
...
   ->
   file_line { "virtualenvwrapper_config":
-    path    => "/home/${app_user}/.profile",
-    line    => "source /etc/bash_completion.d/virtualenvwrapper",
-    require => Python::Virtualenv[$virtualenv],
+    path        => "/home/${app_user}/.profile",
+    line        => "export VIRTUALENVWRAPPER_PYTHON=/usr/bin/python3 && export PROJECT_HOME=/vagrant/refinery && export VIRTUALENVWRAPPER_VIRTUALENV=/usr/bin/virtualenv && export WORKON_HOME=/home/${app_user}/.virtualenv/ && source ./.local/bin/virtualenvwrapper.sh",
```
hackdna: This should suffice:
```shell
VIRTUALENVWRAPPER_PYTHON='/usr/bin/python3'
source /usr/local/bin/virtualenvwrapper.sh
```
https://askubuntu.com/questions/974511/ubuntu-16-04-virtualenvwrapper-python3-6-path

refinery/analysis_manager/models.py
```diff
@@ -43,15 +43,15 @@ class AnalysisStatus(models.Model):
...
+    def __str__(self):
         return self.analysis.name
 
     def set_galaxy_history_state(self, state):
         """
         Set the `galaxy_history_state` of an analysis
         :param state: a valid GALAXY_HISTORY_STATE
         """
-        if state in dict(self.GALAXY_HISTORY_STATES).keys():
+        if state in list(dict(self.GALAXY_HISTORY_STATES).keys()):
```
hackdna: `list()` should not be necessary since `keys()` returns an iterable

refinery/analysis_manager/models.py
```diff
@@ -62,7 +62,7 @@ def set_galaxy_import_state(self, state):
         Set the `galaxy_import_state` of an analysis
         :param state: a valid GALAXY_HISTORY_STATE
         """
-        if state in dict(self.GALAXY_HISTORY_STATES).keys():
+        if state in list(dict(self.GALAXY_HISTORY_STATES).keys()):
```
hackdna: `list()` should not be necessary since `keys()` returns an iterable

deployment/puppet/refinery/manifests/python.pp
ilan-gold: the ubuntu version we're on comes with 3.5 - otherwise we'd have to pull from [here](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa).  also, i don't know if django 1.8 supports 3.8 officially - the website was somewhat unclear.

deployment/puppet/refinery/manifests/python.pp
ilan-gold: what do you mean that it is "available?"

deployment/puppet/refinery/manifests/python.pp
ilan-gold: related to the above, if we were to use a different version of python from deadsnakes, it was not my experience that this would inherit.

ilan-gold: thus i left it alone since it could be helpful to have around in the future and i prefer the specificity

refinery/analysis_manager/models.py
ilan-gold: i am aware - this was from 2to3, whose changes i did not touch.  i can update these if you'd prefer though

deployment/puppet/refinery/manifests/python.pp
hackdna: Good point, Django 1.8 - 1.10 support only up to 3.5 (https://docs.djangoproject.com/en/1.11/faq/install/#what-python-version-can-i-use-with-django)

However, 3.5 end-of-life is 2020-09-13 (https://devguide.python.org/#status-of-python-branches) - that should motivate upgrade to Ubuntu 18.04 (it comes with 3.6) and to Django 1.11+

ilan-gold: yup, i filed that away in the back of my head while doing this :)

deployment/puppet/refinery/manifests/apache2.pp
ilan-gold: i will look into this - it might be touching a different .conf file than the apache::vhost but i can't remember.  i've looked at the puppet source code for this ( while going down the rabbit hole that led me to change these similar settings in apache::vhost) and it's very opaque so i would not be surprised if it were necessary.  also this is another thing to watch out for when we  upgrade to python3.8.  i don't think there a version of libapache2-mod-wsgi-py3 out for python3.8 so we'll need to switch over to pip-installing it - i think the writer of the package suggests doing it this way going forward anyway which might explain its absence.  in any case, the way to do this was not simple to figure out, but i know how now :)

deployment/puppet/refinery/manifests/python.pp
ilan-gold: in short, though, this is a rabbit hole i chose not to go down any further.  i think this is a known issue with pip and python and ubuntu: https://askubuntu.com/a/1106328

deployment/puppet/refinery/manifests/python.pp
ilan-gold: we need the `WORKON_HOME` as well but otherwise we got rid of the other fields

deployment/puppet/refinery/manifests/apache2.pp
ilan-gold: i have removed these now - i appear to have forgotten reading this: https://stackoverflow.com/a/50165569

essentially, we have no choice but to download mod_wsgi from PyPi unless we want to have a broken version and nothing works

deployment/puppet/refinery/manifests/python.pp
hackdna: `virtualenvwrapper` is available as an APT package

refinery/analysis_manager/tasks.py
```diff
@@ -475,7 +478,7 @@ def _get_galaxy_download_task_ids(analysis):
     except galaxy.client.ConnectionError as exc:
         error_msg = \
             "Error downloading Galaxy history files for analysis '%s': %s"
-        logger.error(error_msg, analysis.name, exc.message)
+        logger.error(error_msg, analysis.name, str(exc))
```
hackdna: I don't think explicit conversion to string is required

deployment/puppet/refinery/manifests/python.pp
```diff
@@ -7,16 +7,21 @@
   $django_root         = $refinery::params::django_root,
 ) inherits refinery::params {
 
+
```
hackdna: Unnecessary blank line

deployment/puppet/refinery/manifests/python.pp
```diff
@@ -7,16 +7,21 @@
...
   class { '::python':
+    version    => 'python3.5',
+    ensure     => 'latest',
     dev        => 'present',
     gunicorn   => 'absent',
     pip        => 'latest',
-    virtualenv => 'present',
+    virtualenv => 'present'
```
hackdna: Need a trailing comma

.travis.yml
```diff
@@ -17,6 +17,9 @@ addons:
       - postgresql-client-10
       - libgconf-2-4  # for Cypress
 
+before_install:
+  - sudo apt-get install apache2-dev
```
hackdna: What is this for?

ilan-gold: mod_wsgi: https://modwsgi.readthedocs.io/en/develop/user-guides/quick-installation-guide.html

ilan-gold: (since we install it via pip now and thus it gets installed on travis too)

requirements.txt
```diff
@@ -28,16 +29,19 @@ flower==0.9.1
 jsonschema==2.6.0
 Markdown==2.2.1
 mock==2.0.0
-mockcache==1.0.2
+mod_wsgi==4.6.7
```
hackdna: `mod_wsgi` should be installed using `apt` as before

requirements.txt
```diff
@@ -6,9 +6,10 @@ cffi==1.9.1
 codecov==2.0.11
 coverage==4.0.3
 cryptography==2.5
+Cython==0.29.13
```
hackdna: Why is this needed?

requirements.txt
```diff
@@ -28,16 +29,19 @@ flower==0.9.1
 jsonschema==2.6.0
 Markdown==2.2.1
 mock==2.0.0
-mockcache==1.0.2
+mod_wsgi==4.6.7
+git+git://github.com/lunant/mockcache@master
+ndg-httpsclient==0.3.2
 networkx==1.7
 psycopg2-binary==2.7.4
+pyasn1==0.1.9
```
hackdna: Why is this required?

requirements.txt
```diff
@@ -28,16 +29,19 @@ flower==0.9.1
 jsonschema==2.6.0
 Markdown==2.2.1
 mock==2.0.0
-mockcache==1.0.2
+mod_wsgi==4.6.7
+git+git://github.com/lunant/mockcache@master
+ndg-httpsclient==0.3.2
```
hackdna: Why is this required?

requirements.txt
```diff
@@ -28,16 +29,19 @@ flower==0.9.1
 jsonschema==2.6.0
 Markdown==2.2.1
 mock==2.0.0
-mockcache==1.0.2
+mod_wsgi==4.6.7
+git+git://github.com/lunant/mockcache@master
```
hackdna, line 31: The last commit to that repo was in 2016. I looks like this module is no longer maintained, so need to find a replacement or remove it.

ilan-gold: Nils and I both felt comfortable pinning it to master (we can also do a specific commit) but removing it should be a separate issue independent of this pull request if you feel strongly.

deployment/puppet/refinery/manifests/python.pp
hackdna: The problem is that this creates two sources of truth for Python version which can lead to errors. We are using the OS-provided Python now, so this is not needed.

ilan-gold: That sounds right, good catch - I hadn't thought that putting that there with the OS version would download an extra version like it does for a different number with deadsnakes i.e 3.7

requirements.txt
hackdna: Does Nils know that this package is no longer maintained?

deployment/puppet/refinery/manifests/python.pp
```diff
@@ -72,12 +79,18 @@
...
+    group       => $app_group,
+    require     => Python::Virtualenv[$virtualenv],
+  }
   ->
   file_line { "virtualenvwrapper_config":
-    path    => "/home/${app_user}/.profile",
-    line    => "source /etc/bash_completion.d/virtualenvwrapper",
-    require => Python::Virtualenv[$virtualenv],
+    path        => "/home/${app_user}/.profile",
+    line        => "export VIRTUALENVWRAPPER_PYTHON=/usr/bin/python3 && export WORKON_HOME=/home/${app_user}/.virtualenvs/ && source ./.local/bin/virtualenvwrapper.sh",
```
hackdna: I don't think this line should be changed at all if virtualenvwrapper is installed by APT.

refinery/analysis_manager/tasks.py
```diff
@@ -3,7 +3,7 @@
 
 @author: nils
 '''
-import urlparse
+import urllib.parse
```
hackdna: just `import urllib` is sufficient

refinery/core/models.py
```diff
@@ -899,15 +899,15 @@ def get_node_collection(self):
                 NodeCollection.MultipleObjectsReturned) as exc:
             logger.error('Failed to get NodeCollection for '
                          'Investigation %s: %s',
-                         unicode(self.investigation), exc)
+                         str(self.investigation), exc)
```
hackdna: Explicit cast to string is probably unnecessary. Also, this whole statement can be compacted to two lines with sacrificing readability.

refinery/core/models.py
```diff
@@ -2120,9 +2118,9 @@ def get_csv_row(self, aggregates=False):
         def get_aggregate_sum(field_name):
             if not aggregates:
                 return getattr(self, field_name)
-            return SiteStatistics.objects.filter(
+            return list(SiteStatistics.objects.filter(
```
hackdna, line 2124: Probably unnecessary

refinery/core/views.py
```diff
@@ -2,7 +2,9 @@
 from datetime import timedelta
 import json
 import logging
-import urllib
+import urllib.request
+import urllib.parse
+import urllib.error
```
hackdna: `import urllib` should be sufficient

refinery/data_set_manager/isa_tab_parser.py
```diff
@@ -1173,7 +1168,7 @@ def get_dataset_name(self, path):
         investigation_title = True
         title = None
 
-        with open(investigation_file_name, 'r') as f:
+        with open(investigation_file_name, 'r', encoding='latin-1') as f:
```
hackdna, line 1171: Why 'latin-1'?

refinery/data_set_manager/test_models.py
```diff
@@ -216,48 +216,54 @@ def test_get_assay_count(self):
...
-             u'rfc134.txt', u'rfc174.txt', u'rfc177.txt', u'rfc178.txt',
-             u'rfc86.txt', u'rfc94.txt']
+            ['rfc-test.zip', 'rfc111.txt', 'rfc125.txt', 'rfc126.txt',
+             'rfc134.txt', 'rfc174.txt', 'rfc177.txt', 'rfc178.txt',
+             'rfc86.txt', 'rfc94.txt']
         )
 
     def test_get_datafile_names_local_only(self):
-        with open(os.path.join(TEST_DATA_BASE_PATH, "rfc-test.zip")) as isatab:
+        with open(os.path.join(TEST_DATA_BASE_PATH, "rfc-test.zip"), 'rb') \
```
hackdna: may be:
```python
isa_archive_path = os.path.join(TEST_DATA_BASE_PATH, 'rfc-test.zip')
```

refinery/data_set_manager/test_utils.py
```diff
@@ -448,11 +449,11 @@ def test_is_field_in_hidden_list(self):
     def test_generate_solr_params_no_params_returns_obj(self):
         # empty params
         query = generate_solr_params_for_assay(QueryDict({}), self.valid_uuid)
-        self.assertItemsEqual(sorted(query.keys()), ['json', 'params'])
+        self.assertCountEqual(sorted(query.keys()), ['json', 'params'])
```
hackdna, line 452: This looks like a much weaker test. Is there anything similar to `assertItemsEqual`?

refinery/data_set_manager/views.py
```diff
@@ -367,7 +369,11 @@ def post(self, request, *args, **kwargs):
                 try:
                     response = import_by_file(f)
                 except Exception as e:
-                    logger.error(traceback.format_exc(e))
+                    etype, value, tb = sys.exc_info()
+                    logger.error("{} {}".format(
+                        PARSER_UNEXPECTED_ERROR_MESSAGE,
+                        "".join(traceback.format_exception(etype, value, tb))
+                    ))
```
hackdna, line 376: Why not just `logger.error(e)`?

refinery/data_set_manager/test_utils.py
ilan-gold: my understanding is that they are identical - i have a link to a reference in the issue for this but here is another: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertItemsEqual

refinery/data_set_manager/isa_tab_parser.py
ilan-gold: i think this is rectifying another python hiccup: http://python-notes.curiousefficiency.org/en/latest/python3/text_file_processing.html.  seems to best to just leave it as close to 2 as possible

refinery/ui/cypress/support/commands.js
```diff
@@ -36,7 +36,7 @@ function shell_quote(str) {
 
 Cypress.Commands.add('wrap_cmd',
   function(manage_cmd) {
-    var cd_cmd = "cd .. && " + manage_cmd;
+    var cd_cmd = "cd .. && source ~/virtualenv/python3.5.6/bin/activate && " + manage_cmd;
```
hackdna, line 39: Appears to be redundant with `workon refinery-platform`

ilan-gold: this is specifically for travis - the `workon` command is not run because `vagrant` is not run

refinery/analysis_manager/tasks.py
ilan-gold: i tried this:
```import urllib
urllib.parse('foo')
  File "<input>", line 1, in <module>
TypeError: 'module' object is not callable
```

i also don't want to `from urllib import parse` (or its ilk in `urllib` since those words are so common)

requirements.txt
ilan-gold: yes - we only use this in our unit testing

refinery/ui/cypress/support/commands.js
hackdna: It was working previously, so why wouldn't `workon` run now? What is the connection between virtualenvwrapper and Vagrant exactly?

ilan-gold: The `workon_cmd` is not run in travis, but the `cd_cmd` is.  I have not tried it for a bit, and so I could be wrong, but it appeared that the virtualenv was not persisting to the cypress test for some reason.  Thus it needed to be reactivated.

requirements.txt
ilan-gold: `pysam` doesn't pin its `Cython` dependency well so I figured it safest for us to pin ours (I did have some trouble with incorrect `Cython` versions locally but didn't spend too much time digging into this)

hackdna: Pinning 3rd party dependencies is problematic because it can cause resolution conflicts with other packages and also, it is easy to forget to remove them if the parent package is removed.

refinery/data_set_manager/views.py
ilan-gold: I just tried to replicate the behavior - I did not try to change what it did.

refinery/analysis_manager/tasks.py
```diff
@@ -3,7 +3,7 @@
 
 @author: nils
 '''
-import urlparse
+import urllib
```
hackdna: It looks like this kind of import is not going to work after all in this case, so I'd do `from urllib.parse import urljoin`, etc.

refinery/data_set_manager/utils.py
```diff
@@ -951,13 +950,13 @@ def get_file_url_from_node_uuid(node_uuid, require_valid_url=False):
             return core.utils.build_absolute_url(url) if url else None
         except ValueError as e:
             logger.error('URL {} is not a valid relative url'.format(str(url)))
-            raise type(e)(e.message)
+            raise type(e)(str(e))
```
hackdna: Simply `raise`?

refinery/data_set_manager/utils.py
```diff
@@ -951,13 +950,13 @@ def get_file_url_from_node_uuid(node_uuid, require_valid_url=False):
...
         except RuntimeError as e:
             logger.error('Could not build absolute URL for {}'.format(
                     str(url)
                 )
             )
-            raise type(e)(e.message)
+            raise type(e)(str(e))
```
hackdna: Simply `raise`?

refinery/data_set_manager/utils.py
ilan-gold: I wanted to keep the message

refinery/data_set_manager/utils.py
ilan-gold: I wanted to keep the message

refinery/file_store/tasks.py
```diff
@@ -1,10 +1,9 @@
-from __future__ import division
+
 import contextlib
 import os
 import tempfile
 import threading
-import urllib2
-import urlparse
+import urllib
```
hackdna:
```python
from urllib.parse import urlparse
from urllib.request import urlopen
```

refinery/file_store/tasks.py
```diff
@@ -234,16 +233,16 @@ def import_url_to_path(self, source_url, target_name=None):
         storage = SymlinkedFileSystemStorage()
         # remove query string from URL before extracting file name
         if target_name is None:
-            target_name = os.path.basename(urlparse.urlparse(source_url).path)
+            target_name = os.path.basename(urllib.parse(source_url).path)
```
hackdna: `urlparse`

refinery/file_store/tasks.py
```diff
@@ -234,16 +233,16 @@ def import_url_to_path(self, source_url, target_name=None):
...
         file_store_name = storage.get_name(target_name)
         file_store_path = storage.path(file_store_name)
 
         logger.debug("Transferring from '%s' to '%s'",
                      source_url, file_store_path)
         make_dir(os.path.dirname(file_store_path))
         try:
-            with contextlib.closing(urllib2.urlopen(source_url,
-                                                    timeout=30)) as response, \
+            with contextlib.closing(urllib.request.urlopen(source_url,
```
hackdna: `urlopen`

refinery/file_store/tasks.py
```diff
@@ -262,14 +261,15 @@ def import_url_to_s3(self, source_url, target_name=None):
         storage = S3MediaStorage()
         # remove query string from URL before extracting file name
         if target_name is None:
-            target_name = os.path.basename(urlparse.urlparse(source_url).path)
+            target_name = os.path.basename(urllib.parse(source_url).path)
```
hackdna: `urlparse`

refinery/file_store/tasks.py
```diff
@@ -262,14 +261,15 @@ def import_url_to_s3(self, source_url, target_name=None):
...
         file_store_name = storage.get_name(target_name)
 
         logger.debug("Transferring from '%s' to 's3://%s/%s'",
                      source_url, settings.MEDIA_BUCKET, file_store_name)
         try:
-            with contextlib.closing(urllib2.urlopen(source_url, timeout=30)) \
-                    as response:
+            with contextlib.closing(
+                    urllib.request.urlopen(source_url, timeout=30)
```
hackdna: `urlopen`

refinery/file_store/utils.py
```diff
@@ -4,9 +4,7 @@
 import shutil
 import stat
 import hashlib
-import urllib2
-import urlparse
-
+import urllib
```
hackdna:
```python
from urllib.request import urlopen
from urllib.parse import urlparse
```

refinery/file_store/utils.py
```diff
@@ -158,9 +156,9 @@ def get_file_size(file_location):
             return UNKNOWN_FILE_SIZE
     else:
         try:
-            with contextlib.closing(urllib2.urlopen(file_location,
-                                                    timeout=30)) as response:
-                return int(response.info().getheader('Content-Length'))
+            with contextlib.closing(urllib.request.urlopen(file_location,
```
hackdna: `urlopen`
Also, `contextlib.closing` is no longer necessary: https://docs.python.org/3/howto/urllib2.html#fetching-urls

refinery/file_store/utils.py
```diff
@@ -158,9 +156,9 @@ def get_file_size(file_location):
...
+                                    timeout=30)) as response:
+                return int(response.info().get('Content-Length'))
```
hackdna: Should this be: `return int(response.getheader('Content-Length'))`?
https://docs.python.org/3/library/http.client.html#http.client.HTTPResponse.getheader

refinery/file_store/utils.py
```diff
@@ -201,7 +199,7 @@ def parse_s3_url(url):
     """Return a tuple containing S3 bucket name and object key given S3
     protocol URL (s3://bucket-name/key)
     """
-    result = urlparse.urlparse(url)
+    result = urllib.parse.urlparse(url)
```
hackdna: `urlparse`

refinery/tool_manager/test_views.py
```diff
@@ -600,7 +602,7 @@ def test_tool_creation_with_same_display_name_yields_helpful_error(self):
         self.assertEqual(
             "A Tool already exists with a display_name of: '{}'".format(
                 display_name
-            ),
+            ).encode(),
             self.post_response.content
```
hackdna, line 606: `self.post_response.content.decode()` instead?

refinery/ui/cypress/support/commands.js
hackdna: It is still not clear why did this addition become necessary. What was the error message? Also, that path does not exist in Vagrant:
```shell
vagrant@refinery:~$ ls ~/virtualenv/python3.5.6/bin/activate
ls: cannot access '/home/vagrant/virtualenv/python3.5.6/bin/activate': No such file or directory
```

refinery/data_set_manager/utils.py
hackdna: The message is kept as part of the exception that is re-raised:
```python
>>> try:
...     raise RuntimeError('blah')
... except RuntimeError:
...     raise
...
  File "<stdin>", line 2, in <module>
RuntimeError: blah
```

refinery/core/models.py
ilan-gold: this is necessary - you cannot index a `dict_value` object

hackdna: and if you need a list just remove `[0]`:
```python
>>> SiteStatistics.objects.all().aggregate(Sum('datasets_shared')).values()
[230]
```

refinery/ui/cypress/support/commands.js
ilan-gold: The build fails in `travis` without it, if my memory serves.  I'll checkout a separate branch to confirm and remove the line to see  what happens.  If it is necessary, I'll move it so that it executes separate from the other commands.

refinery/analysis_manager/tasks.py
ilan-gold: Can you be more specific? This works in a python shell.

hackdna: No, it doesn't:
```python
>>> import urllib
>>> urllib.parse.urlparse('foo')
  File "<stdin>", line 1, in <module>
AttributeError: module 'urllib' has no attribute 'parse'
>>> from urllib.parse import urlparse
>>> urlparse('foo')
ParseResult(scheme='', netloc='', path='foo', params='', query='', fragment='')
```

refinery/tool_manager/test_views.py
ilan-gold: I thought it better to leave the `content` untouched.

refinery/core/models.py
ilan-gold: That returns a dictionary - I think we need the value of the dictionary, specifically the first one to return the proper element.

PR #3475, opened 2019-10-23
Fix Dict/Unicode bug

ilan-gold:
Related to the fix for #3435 we need to actually check types first -as I understand it, if the dict was stored and then is immediately reused, the django model returns a dict while a subsequent call later will return a string when it actually hits the database.  Correct me if I'm wrong.  In general, it's probably worth migrating to https://docs.djangoproject.com/en/1.9/ref/contrib/postgres/fields/#jsonfield once we have it.  Feel free to comment

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -1067,9 +1067,10 @@ class Analysis(OwnableResource):
                                      self.summary)
 
     def get_expanded_workflow_graph(self):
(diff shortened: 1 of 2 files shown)


hackdna:
Yes, serialization happens when you call `save()`: https://docs.djangoproject.com/en/1.8/ref/models/instances/#what-happens-when-you-save
If you make sure the workflow is always read from the DB then checking the value type won't be necessary.

Issue #2965, opened 2018-08-29
CPU utilization continuously high


hackdna:
* Specific code commit: v1.6.5
* Version of the web browser and OS: N/A
* Environment where the error occurred: AWS

### Steps to reproduce
Unknown

### Observed behavior
CPU utilization of beta.stemcellcommons.org stayed above 30% continuously over the past two weeks.

![screen shot 2018-08-29 at 10 44 38](https://user-images.githubusercontent.com/452575/44796222-8cb78f80-ab7a-11e8-8366-e8e27e9e6411.png)
![screen shot 2018-08-29 at 10 45 06](https://user-images.githubusercontent.com/452575/44796240-93de9d80-ab7a-11e8-808b-ddbd883d08b0.png)

### Expected behavior
CPU utilization should be close to zero when the site is idle.

Ref: #2789

hackdna:
Still an issue in v1.6.6.

hackdna:
RDS CPU utilization is also high.

scc-prod running v1.6.6 (~10%):
![screen shot 2018-11-14 at 10 34 07](https://user-images.githubusercontent.com/452575/48493167-5ee8fb00-e7f9-11e8-8c41-dae611a00629.png)
vs rds-refinery-of963ie running an earlier version (~1.75%):
![screen shot 2018-11-14 at 10 36 32](https://user-images.githubusercontent.com/452575/48493215-6f997100-e7f9-11e8-931c-16bb6d915356.png)

hackdna:
The baseline CPU utilization has now climbed to over 50% on the production SCC instance.

![screen shot 2019-01-16 at 10 38 21](https://user-images.githubusercontent.com/452575/51259901-14136d00-197b-11e9-8ce3-18b3fa1cdcd0.png)
![screen shot 2019-01-16 at 10 38 45](https://user-images.githubusercontent.com/452575/51259902-14136d00-197b-11e9-832e-d313f3e1467a.png)

ngehlenborg:
CPU utilization is hovering around 5% as of today. Can we close this?

ngehlenborg:
Let’s close and create a new issue to set up a CloudWatch monitor.



On Oct 28, 2019, at 11:54 AM, Ilya Sytchev <notifications@github.com> wrote:

﻿

OK but before closing it would be great to set up some CloudWatch alerts to keep an eye on this.

ilan-gold:
I'll set this up in Terraform for deployments going forward and for the current production instance as well

PR #3483, opened 2019-10-30
Set Up CloudWatch Alarms for CPU Utilization

ilan-gold:


ilan-gold:
@hackdna this has been tested

deployment/terraform/modules/ec2/main.tf
```diff
@@ -347,3 +347,21 @@ resource "aws_elb" "https" {
     unhealthy_threshold = 4
   }
 }
+
+resource "aws_cloudwatch_metric_alarm" "cpu_utilization" {
```
hackdna: `resource "aws_cloudwatch_metric_alarm" "app_server_cpu_utilization"`
We may have other CPU utilization alarms.

deployment/terraform/modules/ec2/main.tf
```diff
@@ -347,3 +347,21 @@ resource "aws_elb" "https" {
...
+  alarm_name                = "cpu-utilization-limit-exceeded"
```
hackdna: Alarm name must be unique within AWS account, and also should be descriptive.
`alarm_name = "${var.resource_name_prefix}-app-server-cpu-utilization"`

deployment/terraform/modules/ec2/main.tf
```diff
@@ -347,3 +347,21 @@ resource "aws_elb" "https" {
...
+  comparison_operator       = "GreaterThanOrEqualToThreshold"
+  evaluation_periods        = "5"
+  metric_name               = "CPUUtilization"
+  namespace                 = "AWS/EC2"
+  period                    = "300"
```
hackdna: If I'm not mistaken this is going to raise an alarm if CPU utilization has been >=25% for over 25 minutes. Is that what we want?

deployment/terraform/modules/ec2/main.tf
```diff
@@ -347,3 +347,21 @@ resource "aws_elb" "https" {
...
+  statistic                 = "Average"
+  threshold                 = "25"
+  alarm_description         = "This metric monitors ec2 cpu utilization"
```
hackdna: `"Monitors CPU utilization of ${aws_instance.app_server.tags.Name}"`

deployment/terraform/modules/ec2/main.tf
```diff
@@ -347,3 +347,21 @@ resource "aws_elb" "https" {
     unhealthy_threshold = 4
   }
 }
+
+resource "aws_cloudwatch_metric_alarm" "cpu_utilization" {
+  alarm_name                = "cpu-utilization-limit-exceeded"
+  comparison_operator       = "GreaterThanOrEqualToThreshold"
```
hackdna: Terraform style guide calls for aligning all equal signs within a block to the longest attribute name which is this one.

deployment/terraform/modules/ec2/main.tf
ilan-gold: This is a judgement call as far as I can tell.  25 minutes seemed like a good amount of time to allow for spikes/prolonged periods of usage.

hackdna: I see. Also, how and to whom this alarm notification will be delivered?

ilan-gold: I was under the impression that a notification would go to the root account user, but that was clearly mistaken.  Not sure how I came to that conclusion.  In any case: https://www.terraform.io/docs/providers/aws/r/sns_topic_subscription.html#email

In short, setting up an email endpoint for an sns topic entirely within Terraform is impossible.  My vote would be to create the topic/subscription outside Terraform within AWS and just pass in the arn as a default option to the alarms for subscription.

hackdna: Yes, email subscriptions require an out-of-band confirmation. Also, some deployments may need a different type of notification or no notification at all. Could you set it up such that it is possible to set up a notification but it is not required by default?

ilan-gold: Yes I think that makes sense

deployment/terraform/modules/rds/main.tf
```diff
@@ -45,3 +45,20 @@ resource "aws_db_instance" "default" {
...
+
+resource "aws_cloudwatch_metric_alarm" "default_rds_cpu_utilization" {
+  alarm_name          = "${var.resource_name_prefix}-default-rds-cpu-utilization"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = "5"
+  metric_name         = "CPUUtilization"
+  namespace           = "AWS/RDS"
+  period              = "300"
+  statistic           = "Average"
+  threshold           = "25"
```
hackdna: Given the data (https://github.com/refinery-platform/refinery-platform/issues/2965#issuecomment-438707280) I'd say a 10% threshold would be more appropriate.

deployment/terraform/modules/rds/main.tf
```diff
@@ -45,3 +45,20 @@ resource "aws_db_instance" "default" {
...
+  alarm_description   = "Monitors CPU utilization of default RDS instance"
```
hackdna: `"Monitors CPU utilization of ${aws_db_instance.default.id}"` or whatever is the human readable name.

ilan-gold: I think the most human readable name is "default."  There are no tags or the like for RDS.  We can change that though.

deployment/terraform/modules/rds/main.tf
ilan-gold: Ok

ilan-gold: does this mean for both ec2 and rds?

deployment/terraform/live/vars.tf
```diff
@@ -195,3 +195,8 @@ variable "data_volume_snapshot_id" {
   description = "A snapshot to base the EBS data volume off of"
   default     = ""
 }
+
+variable "alarm_sns_arn" {
+  description = "The arn of an sns topic to publish cloudwatch alarm metrics"
```
hackdna: "The ARN of an SNS topic to publish CloudWatch alarms"

Issue #2917, opened 2018-07-31
Certain user accounts fail to activate


hackdna:
* Specific code commit: v1.6.5
* Version of the web browser and OS: unknown
* Environment where the error occurred: AWS

### Steps to reproduce
Unknown, error report received by email

### Observed behavior
```
Internal Server Error: /accounts/register/
Traceback (most recent call last):
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/django/utils/decorators.py", line 34, in _wrapper
    return bound_func(*args, **kwargs)
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/django/views/decorators/debug.py", line 76, in sensitive_post_parameters_wrapper
    return view(request, *args, **kwargs)
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/django/utils/decorators.py", line 30, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/registration/views.py", line 53, in dispatch
    return super(RegistrationView, self).dispatch(request, *args, **kwargs)
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/django/views/generic/base.py", line 89, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/django/views/generic/edit.py", line 215, in post
    return self.form_valid(form)
  File "/home/ubuntu/.virtualenvs/refinery-platform/lib/python2.7/site-packages/registration/views.py", line 56, in form_valid
    new_user = self.register(form)
  File "/srv/refinery-platform/refinery/core/views.py", line 1110, in register
    first_name, last_name, affiliation)
  File "/srv/refinery-platform/refinery/core/models.py", line 2104, in custom_create_inactive_user
[...]

PR #3486, opened 2019-10-31
Ilan gold/user registration bug

ilan-gold:
i believe this fixes: https://github.com/refinery-platform/refinery-platform/issues/2917 since we are not encoding (via the cast) anymore.


ilan-gold:
confirmed to be fixed:

Screen Shot 2019-11-04 at 12 01 44 PM
Screen Shot 2019-11-04 at 12 01 49 PM

PR #3487, opened 2019-11-01
Fixing the Provenance Graph

ilan-gold:
The problem essentially boiled down to two things:

1) We were not returning expected information in the serializer for Analyses
2) We should not visualize auxiliary nodes (i.e bai files) produced by the refinery system.  They are of no use to the user and also mess up the topological ordering of the visualization.


refinery/core/serializers.py
```diff
@@ -47,6 +49,16 @@ def get_owner(self, analysis):
     def get_data_set_uuid(self, analysis):
         return analysis.data_set.uuid
 
+    def get_creation_date(self, analysis):
```
jkmarx: The provenance client side should be updated to use time_start instead of adding another field to the serializer with the same functionality.

refinery/core/serializers.py
```diff
@@ -47,6 +49,16 @@ def get_owner(self, analysis):
...
+        return analysis.time_start
+
+    def get_workflow_json(self, analysis):
```
jkmarx: This should be added to the workflow serializer and the client should make the appropriate call.

refinery/core/serializers.py
```diff
@@ -27,12 +26,13 @@ class AnalysisSerializer(serializers.ModelSerializer):
...
     time_end = DateTimeWithTimeZone()
     time_start = DateTimeWithTimeZone()
+    workflow__uuid = serializers.SerializerMethodField()
 
     class Meta:
         model = Analysis
         fields = ('data_set_uuid', 'facet_name', 'name', 'owner', 'status',
                   'summary', 'time_start', 'time_end', 'tool_display_name',
-                  'uuid', 'workflow')
+                  'uuid', 'workflow', 'workflow__uuid')
```
jkmarx: Either use the workflow__uuid or workflow id field. We tend to use uuids, so check if we can just remove the workflow field (returns id).

ilan-gold: I think we can remove - I just pushed the change so we'll see if testing backs up the lack of usages I saw

PR #3488, opened 2019-11-06
Adding some tests and better handling for non-ascii throughout

ilan-gold:
No bug here (yet) but while looking at some other code, I figured #2917 could use some testing and more robustness throughout the codebase (i.e beyond just registering people)


ilan-gold:
Unless that is going to be done before the next release, I think we should get this in before the next release.

PR #3491, opened 2019-11-18
IGV Bug Fix

ilan-gold:
`options.json` used in docker container was failing because of `child.uuid` throwing an error since `child` is a string.  this is tested and resolved


ilan-gold:
need to merge into release branch

Issue #3490, opened 2019-11-15
Celery Chain Does Not Fail


ilan-gold:
* Specific code commit: https://github.com/refinery-platform/refinery-platform/tree/d60004dff050f377c659e0b0cb661ab9ab75e363
* Version of the web browser and OS: Any
* Environment where the error occurred (Vagrant VM and site conf mode or AWS instance): 
AWS
### Steps to reproduce
Please list all the actions and the input data used:
This actually appears to be a broad based problem with celery (you will need to move `fail` and `succeed` into a `tasks.py` file so they are discovered):

```
import os;
import django;
os.environ['DJANGO_SETTINGS_MODULE'] = 'config.settings.dev'
django.setup()
from celery import chain

@task
def fail():
    raise IndexError
@task
def succeed()
    print 'foo'

t = chain( fail.subtask(), succeed.subtask(), time_limit=20)
t.delay()
```

### Observed behavior
```
2019-11-15 11:31:13,332 ERROR    celery.worker.job:282 log() - Task data_set_manager.tasks.generate_auxiliary_file[29c3d044-7579-429b-bb91-7ab688f3788a] raised unexpected: SamtoolsError()
Traceback (most recent call last):
  File "/home/ubuntu/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/ubuntu/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/srv/refinery-platform/refinery/data_set_manager/tasks.py", line 307, in generate_auxiliary_file
    auxiliary_file_path = generate_bam_index(datafile_path)
  File "/srv/refinery-platform/refinery/data_set_manager/tasks.py", line 342, in generate_bam_index
    pysam.index(bytes(temp_bam_file_path), bytes(bam_index_file_path))
  File "/home/ubuntu/.virtualenvs/refinery-platform/local/lib/python2.7/site-packages/pysam/utils.py", line 75, in __call__
    stderr))
SamtoolsError

```

There is a failure in the chain, but the whole chian does not fail and instead hangs.
```
[...]

PR #3492, opened 2019-11-18
remove usage of chain and do FileImportTask as post-process step

ilan-gold:
Evidently `chain` does not fail in the case of a failure of the first task.  See #3490


ilan-gold:
need to merge into release branch

PR #3493, opened 2019-11-18
 remove usage of chain and do FileImportTask as post-process step 

ilan-gold:
Evidently chain does not fail in the case of a failure of the first task. See #3490


ilan-gold:
Ok I will remove it then.

ilan-gold:
@hackdna It seems like chains are also somewhat broken (if you could provide some documentation/information as to why signals are also not good, please do!).  I think if we do go the route of chains, then we absolutely need to have a blanket try-catch on certainly the `generate_bam_index` task and potentially (pending testing) the import task as well to avoid workers being alive but not actually doing anything (a huge problem).

ilan-gold:
As for the samtools error, this is the sort of thing I think we should know about so whatever we do, it should include informing us/the user of the issue.

hackdna:
Here's one reference (of many): https://www.quora.com/Are-Django-signals-an-anti-pattern
In this case `post_process_generate_auxiliary()` simply replicates functionality provided by the chain. Btw, chains are not really broken: they simply don't run subsequent tasks after a failure.

Also, when you remove a model field you need to have a migration for that. Otherwise, the databases of running instances will have unused columns (and potentially indices, etc). In this case it's fine because the column was added recently (after latest release) but something to keep in mind in the future.

Finally, if you want to handle the `SamtoolsError` you can simply re-raise it as a `RuntimeError` for now (similar to other exceptions in `generate_bam_index()`). User notification will require a lot more work because it involves creating a new UI, API, etc.

ilan-gold:
Sorry if we keep going back and forth, but I just want to make sure we are on the same page about chains here.  https://github.com/celery/celery/issues/3810 - This chain problem was fixed (maybe) in 4.0.0 and we are in 3.1.26. It's not just that the chains don't run tasks, though - the entire chain parent-task hangs in a `PENDING` state which means that we could eventually run out of workers.  This is my concern, mainly.  If you feel that error-handling is enough to mitigate this risk, I am ok with going back.

hackdna:
No worries. Subsequent tasks are not hanging, they are simply never run. Chain result is just a result of the last task and 'PENDING' is just the default state for tasks. You could test this using your toy example from #3490 and setting concurrency to 1 for `celery-w1`.

ilan-gold:
Ah I see what you are saying - the result is separate from the task itself.  That makes total sense! I will update the code now then.

hackdna:
Great, thanks. Are you planning to handle the `SamtoolsError` as outlined above?

ilan-gold:
I wasn't going to since it won't have any affect on the outcome

hackdna:
True but I think it'd be better to avoid a traceback in the logs (makes it easier to read logs and troubleshoot potential problems).

PR #3494, opened 2019-11-18
Ilan gold/igv bug fix

ilan-gold:
Resolve bug of `child` being a string and not a node object

### DIFF
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -1129,7 +1129,7 @@ class Analysis(OwnableResource):
     def get_status(self):
         return self.status
 
-    def get_auxiliary_node_uuids(self):
+    def get_auxiliary_nodes(self):
         # Return queryset of related auxiliary_nodes
         analysis_nodes = self.get_nodes()
         aux_nodes = Node.objects.none()
@@ -1675,7 +1675,7 @@ def _analysis_delete(sender, instance, *args, **kwargs):
     with transaction.atomic():
         instance.cancel()
         # Delete children aux_nodes Associated w/ the Analysis
-        instance.get_auxiliary_node_uuids().delete()
+        instance.get_auxiliary_nodes().delete()
         # Delete Nodes Associated w/ the Analysis
         instance.get_nodes().delete()
 
--- a/refinery/core/models.py
+++ b/refinery/core/models.py
@@ -1129,7 +1129,7 @@ class Analysis(OwnableResource):
     def get_status(self):
         return self.status
 
-    def get_auxiliary_nodes(self):
+    def get_auxiliary_node_uuids(self):
         # Return queryset of related auxiliary_nodes
         analysis_nodes = self.get_nodes()
         aux_nodes = Node.objects.none()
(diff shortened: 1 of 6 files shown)


refinery/tool_manager/tests.py
```diff
@@ -720,7 +720,7 @@ def _create_detailed_nodes_dict(self, nodes):
             node.uuid: {
                 'file_url': self.node.file_item.get_datafile_url(),
                 'auxiliary_file_list': [
-                    child.uuid for child in
+                    child_uuid for child_uuid in
                     self.node.get_auxiliary_nodes()
```
hackdna: Why iterate over a list of UUIDs to create a list of UUIDs?

refinery/tool_manager/models.py
```diff
@@ -557,8 +557,8 @@ def _get_detailed_nodes_dict(self, node_uuid_list,
                 ),
                 self.AUXILIARY_FILE_LIST: [
                     get_file_url_from_node_uuid(
-                        child.uuid, require_valid_url=require_valid_urls
-                    ) for child in Node.objects.get(uuid=node["uuid"]).
+                        uuid, require_valid_url=require_valid_urls
+                    ) for uuid in Node.objects.get(uuid=node["uuid"]).
```
hackdna: Why not just: `get_file_url_from_node_uuid(node['uuid'], require_valid_url=require_valid_urls)`?

ilan-gold: this is actually what tripped me up and caused the error - not sure how to make it more readable but we are actually getting a list of `uuid`'s from `get_auxiliary_nodes` on the next line

hackdna: Oh, I missed this too. This whole dict comprehension (`node_info`) is way too large. I think it would be much cleaner to rewrite as a simple for loop to avoid problems like this.

refinery/tool_manager/models.py
```diff
@@ -558,12 +559,10 @@ def _get_detailed_nodes_dict(self, node_uuid_list,
                 self.AUXILIARY_FILE_LIST: [
                     get_file_url_from_node_uuid(
                         uuid, require_valid_url=require_valid_urls
-                    ) for uuid in Node.objects.get(uuid=node["uuid"]).
-                    get_auxiliary_nodes()
+                    ) for uuid in
+                    Node.objects.get(uuid=node["uuid"]).get_auxiliary_nodes()
```
hackdna: Looks better but I'd also rename `get_auxiliary_nodes()` to `get_auxiliary_node_uuids()` to reflect the actual return value type.

refinery/core/models.py
```diff
@@ -1129,7 +1129,7 @@ def delete(self, **kwargs):
     def get_status(self):
         return self.status
 
-    def get_auxiliary_nodes(self):
+    def get_auxiliary_node_uuids(self):
```
hackdna: This method belongs to `Analysis` model (not `Node`) and should not be renamed. Please check the rest of the changes to make sure only the correct function is renamed.

refinery/tool_manager/models.py
```diff
@@ -548,22 +548,23 @@ def _get_detailed_nodes_dict(self, node_uuid_list,
...
                 self.AUXILIARY_FILE_LIST: [
                     get_file_url_from_node_uuid(
-                        child.uuid, require_valid_url=require_valid_urls
-                    ) for child in Node.objects.get(uuid=node["uuid"]).
-                    get_auxiliary_nodes()
+                        uuid, require_valid_url=require_valid_urls
+                    ) for uuid in
+                    Node.objects.get(
+                        uuid=node["uuid"]
+                    ).get_auxiliary_node_uuids()
```
hackdna: Could you put this outside the list comprehension for clarity: `auxiliary_nodes = Node.objects.get(uuid=node["uuid"]).get_auxiliary_nodes()`?

